PR #9, opened 2017-02-05
Added mangaeden(it) extension

blackgin:
I take this opportunity to ask a question.
I think I'm going to write also mangaeden(en) and perveden extensions, since the structure is basically the same.
The question is, since perveden is an hentai website, do you still want the extension to be in this repo or not?

### DIFF
--- a/it-mangaeden/src/eu/kanade/tachiyomi/extension/it/mangaeden/Mangaeden.kt
+++ b/it-mangaeden/src/eu/kanade/tachiyomi/extension/it/mangaeden/Mangaeden.kt
@@ -63,7 +63,7 @@ class Mangaeden : ParsedHttpSource() {
         return GET(url.toString(), headers)
     }
 
-    override fun searchMangaSelector() = "table#mangaList > tbody > tr"
+    override fun searchMangaSelector() = "table#mangaList > tbody > tr:has(td:gt(1))"
 
     override fun searchMangaFromElement(element: Element) = SManga.create().apply {
         element.select("td > a").first().let {
--- a/it-mangaeden/src/eu/kanade/tachiyomi/extension/it/mangaeden/Mangaeden.kt
+++ b/it-mangaeden/src/eu/kanade/tachiyomi/extension/it/mangaeden/Mangaeden.kt
@@ -1,0 +1,192 @@
+package eu.kanade.tachiyomi.extension.it.mangaeden
+
+import eu.kanade.tachiyomi.network.GET
+import eu.kanade.tachiyomi.source.model.*
+import eu.kanade.tachiyomi.source.online.ParsedHttpSource
+import okhttp3.HttpUrl
+import okhttp3.Request
+import org.jsoup.nodes.Document
+import org.jsoup.nodes.Element
+import java.text.ParseException
+import java.text.SimpleDateFormat
+import java.util.*
+
+class Mangaeden : ParsedHttpSource() {
+
+    override val id: Long = 11
+
+    override val name = "Mangaeden"
+
+    override val baseUrl = "http://www.mangaeden.com"
+
+    override val lang = "it"
+
+    override val supportsLatest = true
+
+    override fun latestUpdatesRequest(page: Int): Request = GET("$baseUrl/it/it-directory/?order=3&page=$page", headers)
+
+    override fun latestUpdatesSelector() = searchMangaSelector()
+
+    override fun latestUpdatesFromElement(element: Element): SManga = searchMangaFromElement(element)
+
+    override fun latestUpdatesNextPageSelector() = searchMangaNextPageSelector()
+
+    override fun popularMangaRequest(page: Int): Request = GET("$baseUrl/it/it-directory/?order=1&page=$page", headers)
+
+    override fun popularMangaSelector() = searchMangaSelector()
+
+    override fun popularMangaFromElement(element: Element): SManga = searchMangaFromElement(element)
+
+    override fun popularMangaNextPageSelector() = searchMangaNextPageSelector()
+
+    override fun searchMangaRequest(page: Int, query: String, filters: FilterList): Request {
+        val url = HttpUrl.parse("$baseUrl/it/it-directory/").newBuilder().addQueryParameter("title", query)
+        (if (filters.isEmpty()) getFilterList() else filters).forEach { filter ->
+            when (filter) {
+                is Statuses -> {
+                    val id = filter.values[filter.state].id
+                    if (id != -1) url.addQueryParameter("status", id.toString())
+                }
+                is Types -> {
+                    val id = filter.values[filter.state].id
+                    if (id != -1) url.addQueryParameter("type", id.toString())
+                }
+                is GenreList -> filter.state
+                        .filter { !it.isIgnored() }
+                        .forEach { genre -> url.addQueryParameter(if (genre.isIncluded()) "categoriesInc" else "categoriesExcl", genre.id) }
+                is TextField -> url.addQueryParameter(filter.key, filter.state)
+                is OrderBy -> filter.state?.let {
+                    val sortId = it.index
+                    url.addQueryParameter("order", if (it.ascending) "-$sortId" else "$sortId")
+                }
+            }
+        }
+        url.addQueryParameter("page", page.toString())
+        return GET(url.toString(), headers)
+    }
+
+    override fun searchMangaSelector() = "table#mangaList > tbody > tr"
+
+    override fun searchMangaFromElement(element: Element) = SManga.create().apply {
+        element.select("td > a").first().let {
+            setUrlWithoutDomain(it.attr("href"))
+            title = it.text()
+        }
+    }
+
+    override fun searchMangaNextPageSelector() = "a:has(span.next)"
+
+    override fun mangaDetailsParse(document: Document) = SManga.create().apply {
+        val infos = document.select("div.rightbox")
+
+        author = infos.select("a[href^=/it/it-directory/?author]").first().text()
+        artist = infos.select("a[href^=/it/it-directory/?artist]").first().text()
+        genre = infos.select("a[href^=/it/it-directory/?categoriesInc]").map { it.text() }.joinToString()
+        description = document.select("h2#mangaDescription").text()
+        status = parseStatus(infos.select("h4:containsOwn(Stato)").first().nextSibling().toString())
+        thumbnail_url = "http:${infos.select("div.mangaImage2 > img").first().attr("src")}"
+    }
+
+    private fun parseStatus(status: String) = when {
+        status.contains("In Corso", true) -> SManga.ONGOING
+        status.contains("Completato", true) -> SManga.COMPLETED
+        else -> SManga.UNKNOWN
+    }
+
+    override fun chapterListSelector() = "div#leftContent > table > tbody > tr"
+
+    override fun chapterFromElement(element: Element) = SChapter.create().apply {
+        val a = element.select("a[href^=/it/it-manga/]").first()
+
+        setUrlWithoutDomain(a.attr("href"))
+        name = a.select("b").first().text()
+        date_upload = element.select("td.chapterDate").first().text().let { parseChapterDate(it.trim()) }
+    }
+
+    private fun parseChapterDate(date: String): Long = try {
+        SimpleDateFormat("d MMM yyyy", Locale.ITALIAN).parse(date).time
+    } catch (e: ParseException) {
+        0L
+    }
+
+    override fun pageListParse(document: Document): List<Page> = mutableListOf<Page>().apply {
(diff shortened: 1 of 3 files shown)


it-mangaeden/src/eu/kanade/tachiyomi/extension/it/mangaeden/Mangaeden.kt
```diff
@@ -0,0 +1,190 @@
...
+
+    override fun latestUpdatesRequest(page: Int): Request = GET("$baseUrl/it/it-directory/?order=3&page=$page", headers)
+
+    override fun latestUpdatesSelector() = searchMangaSelector()
+
+    override fun latestUpdatesFromElement(element: Element): SManga = searchMangaFromElement(element)
+
+    override fun latestUpdatesNextPageSelector() = searchMangaNextPageSelector()
+
+    override fun popularMangaRequest(page: Int): Request = GET("$baseUrl/it/it-directory/?order=1&page=$page", headers)
```
paronos: `?order=1` can be removed since it is the default ordering (may stay nevertheless)

inorichi:
Yes, it's fine. Should I merge this one?

it-mangaeden/src/eu/kanade/tachiyomi/extension/it/mangaeden/Mangaeden.kt
```diff
@@ -0,0 +1,190 @@
...
+
+    override fun mangaDetailsParse(document: Document) = SManga.create().apply {
+        val infos = document.select("div.rightbox")
+
+        author = infos.select("a[href^=/it/it-directory/?author]").first().text()
+        artist = infos.select("a[href^=/it/it-directory/?artist]").first().text()
+        genre = infos.select("a[href^=/it/it-directory/?categoriesInc]").map { it.text() }.joinToString()
+        description = document.select("h2#mangaDescription").text()
+        status = parseStatus(infos.select("h4:containsOwn(Stato)").first().nextSibling().toString())
+        thumbnail_url = "http:${infos.select("div.mangaImage2 > img").first().attr("src")}"
```
paronos: Same argument, could become something like `infos.select("div.mangaImage2 > img").first()?.attr("src")?.let { thumbnail_url = "http:" + it }`

it-mangaeden/src/eu/kanade/tachiyomi/extension/it/mangaeden/Mangaeden.kt
```diff
@@ -0,0 +1,190 @@
...
+    }
+
+    override fun searchMangaNextPageSelector() = "a:has(span.next)"
+
+    override fun mangaDetailsParse(document: Document) = SManga.create().apply {
+        val infos = document.select("div.rightbox")
+
+        author = infos.select("a[href^=/it/it-directory/?author]").first().text()
+        artist = infos.select("a[href^=/it/it-directory/?artist]").first().text()
+        genre = infos.select("a[href^=/it/it-directory/?categoriesInc]").map { it.text() }.joinToString()
```
paronos: It might be convenient to separate the genres with a comma? `joinToString(", ")`

it-mangaeden/src/eu/kanade/tachiyomi/extension/it/mangaeden/Mangaeden.kt
```diff
@@ -0,0 +1,190 @@
...
+            setUrlWithoutDomain(it.attr("href"))
+            title = it.text()
+        }
+    }
+
+    override fun searchMangaNextPageSelector() = "a:has(span.next)"
+
+    override fun mangaDetailsParse(document: Document) = SManga.create().apply {
+        val infos = document.select("div.rightbox")
+
```
paronos: It might be useful to use null safety after `first()` (e.g. _first()_**?.**_funX()_).
(This also happens later in the file)

blackgin:
Yeah, I think you can if there are no issues

it-mangaeden/src/eu/kanade/tachiyomi/extension/it/mangaeden/Mangaeden.kt
```diff
@@ -0,0 +1,190 @@
...
+    override fun chapterFromElement(element: Element) = SChapter.create().apply {
+        val a = element.select("a[href^=/it/it-manga/]").first()
+
+        setUrlWithoutDomain(a.attr("href"))
+        name = a.select("b").first().text()
+        date_upload = element.select("td.chapterDate").first().text().let { parseChapterDate(it.trim()) }
+    }
+
+    private fun parseChapterDate(date: String): Long = try {
+        SimpleDateFormat("d MMM yyyy", Locale.ITALIAN).parse(date).time
```
paronos: It should also manage dates like "Oggi" (today)

it-mangaeden/src/eu/kanade/tachiyomi/extension/it/mangaeden/Mangaeden.kt
```diff
@@ -0,0 +1,190 @@
...
+
+    override fun imageUrlParse(document: Document): String = "http:${document.select("a#nextA.next > img").first().attr("src")}"
+
+    private class NamedId(val name: String, val id: Int) {
+        override fun toString(): String = name
+    }
+
+    private class Genre(name: String, val id: String) : Filter.TriState(name)
+    private class TextField(name: String, val key: String) : Filter.Text(name)
+    private class Statuses(statuses: Array<NamedId>) : Filter.Select<NamedId>("Stato", statuses)
```
paronos: Handle it in this way does not convince me, because you lose the ability to make multiple selections as in the web interface (e.g. select  "Japanese Manga", "Korean Manhwa", and "Chinese Manhua" but not "Comic" and "Doujinshi").
A `Filter.Group` of `Filter.CheckBox` could be used (somewhat similar to `GenreList` ).

it-mangaeden/src/eu/kanade/tachiyomi/extension/it/mangaeden/Mangaeden.kt
```diff
@@ -0,0 +1,190 @@
...
+    private class Types(types: Array<NamedId>) : Filter.Select<NamedId>("Tipo", types)
```
paronos: Same as above.

paronos:
Sorry for the last minute review.

Yesterday I started MangaEden/EN, luckily I saw this PR before finish it.

it-mangaeden/src/eu/kanade/tachiyomi/extension/it/mangaeden/Mangaeden.kt
blackgin: Isn't comma the default separator?

paronos: It is, my fault.

blackgin:
Thanks very much for the review! I'll make the changes tomorrow and I'll open the other PR for mangaeden(en) + perveden(it+en)

blackgin:
I was implementing the suggestion and I noticed something. It appears that when I try to make a search with only `CheckBox` or `TriState` filters, the `searchMangaRequest` is not called. What am I doing wrong?

inorichi:
Don't save the genres locally (`private val genres = ...`). `getFilterList` must always return a new list of all the filters. Turn those to methods.

inorichi:
> A non related question, the apks will be uploaded here? (maybe an automatic build system is planned?)

Yes. I also want to create an index with versions and links and create an UI in tachiyomi to install/update them. But it's probably going to take a long time 😞 .

blackgin:
It should work now

it-mangaeden/src/eu/kanade/tachiyomi/extension/it/mangaeden/Mangaeden.kt
```diff
@@ -0,0 +1,201 @@
...
+import okhttp3.Request
+import org.jsoup.nodes.Document
+import org.jsoup.nodes.Element
+import java.text.ParseException
+import java.text.SimpleDateFormat
+import java.util.*
+
+class Mangaeden : ParsedHttpSource() {
+
+    override val name = "Mangaeden"
```
paronos: Since the site title is "Leggere Manga Online - Fumetti Online in Italiano - **Manga Eden**", maybe it is better to call it "Manga Eden" (or "MangaEden").

it-mangaeden/src/eu/kanade/tachiyomi/extension/it/mangaeden/Mangaeden.kt
```diff
@@ -0,0 +1,201 @@
...
+
+    override fun mangaDetailsParse(document: Document) = SManga.create().apply {
+        val infos = document.select("div.rightbox")
+
+        author = infos.select("a[href^=/it/it-directory/?author]").first()?.text()
+        artist = infos.select("a[href^=/it/it-directory/?artist]").first()?.text()
+        genre = infos.select("a[href^=/it/it-directory/?categoriesInc]").map { it.text() }.joinToString()
+        description = document.select("h2#mangaDescription").text()
+        status = parseStatus(infos.select("h4:containsOwn(Stato)").first()?.nextSibling().toString())
+        thumbnail_url = infos.select("div.mangaImage2 > img").first()?.attr("src").let { "http:$it" }
```
paronos: This code does not work correctly when the cover is missing (e.g. http://www.mangaeden.com/it/it-manga/nanatsu-no-tazai/) since `attr()` returns a blank string when the attribute is missing. An alternative is:
```kotlin
        val img = infos.select("div.mangaImage2 > img").first()?.attr("src")
        if (!img.isNullOrBlank()) thumbnail_url = "http:$img"
```

it-mangaeden/src/eu/kanade/tachiyomi/extension/it/mangaeden/Mangaeden.kt
```diff
@@ -0,0 +1,201 @@
...
+    private class StatusList(statuses: List<NamedId>) : Filter.Group<NamedId>("Stato", statuses)
+    private class Types(types: List<NamedId>) : Filter.Group<NamedId>("Tipo", types)
+    private class GenreList(genres: List<Genre>) : Filter.Group<Genre>("Generi", genres)
+
+    override fun getFilterList() = FilterList(
+            TextField("Autore", "author"),
+            TextField("Artista", "artist"),
+            Types(types()),
+            StatusList(statuses()),
+            OrderBy(),
```
paronos: Since `Types`, `Statuses` and `GenreList` are all checkboxes, as a personal preference I'd rather have them all after `OrderBy`. If you prefer, you can leave it as it is, otherwise the order would be
```kotlin
            TextField("Autore", "author"),
            TextField("Artista", "artist"),
            OrderBy(),
            Types(types()),
            StatusList(statuses()),
            GenreList(genres())
```

it-mangaeden/src/eu/kanade/tachiyomi/extension/it/mangaeden/Mangaeden.kt
```diff
@@ -0,0 +1,201 @@
...
+    override fun searchMangaNextPageSelector() = "a:has(span.next)"
+
+    override fun mangaDetailsParse(document: Document) = SManga.create().apply {
+        val infos = document.select("div.rightbox")
+
+        author = infos.select("a[href^=/it/it-directory/?author]").first()?.text()
+        artist = infos.select("a[href^=/it/it-directory/?artist]").first()?.text()
+        genre = infos.select("a[href^=/it/it-directory/?categoriesInc]").map { it.text() }.joinToString()
+        description = document.select("h2#mangaDescription").text()
+        status = parseStatus(infos.select("h4:containsOwn(Stato)").first()?.nextSibling().toString())
```
paronos: _nextSibling()_**?.**_toString()_

paronos:
> Yes. I also want to create an index with versions and links and create an UI in tachiyomi to install/update them. But it's probably going to take a long time :disappointed: .

Maybe in the mean time they could be listed in the wiki ([here](https://github.com/inorichi/tachiyomi-extensions/wiki) or in [tachiyomi](https://github.com/inorichi/tachiyomi/wiki)?). It is probably better to make a single list with the three "[YAML Sources](https://github.com/inorichi/tachiyomi/wiki/YAML-Sources)".

Another thing I noticed is that Mangachan has been updated tachiyomi but not here. There may be other cases in the future.

inorichi:
> Maybe in the mean time they could be listed in the wiki (here or in tachiyomi?). It is probably better to make a single list with the three "YAML Sources".

I'll mark that wiki entry as deprecated and link to this repository, but maybe once I release the next version.

> Another thing I noticed is that Mangachan has been updated tachiyomi but not here. There may be other cases in the future.

It'll happen until they are finally removed from tachiyomi. I'll have to keep them updated until then. But since there are no releases yet it's not a big problem.

PR #11, opened 2017-02-07
Added extensions for mangaeden(en) and perveden(it+en)

blackgin:
So here we are with the second PR with the remaining *eden extensions.

I have a question regarding perveden.
Since there are a lot (like **hundreds**) of "_genres_", I decided not to include them among the filters.
What do you think? Do we still want to have them? Do we want a subset of them, maybe?

Also, reviews are super welcome, as always.


en-mangaeden/src/eu/kanade/tachiyomi/extension/en/mangaeden/Mangaeden.kt
```diff
@@ -0,0 +1,203 @@
...
+            } else if ("Yesterday" in date) {
+                Calendar.getInstance().apply {
+                    add(Calendar.DATE, -1)
+                    set(Calendar.HOUR_OF_DAY, 0)
+                    set(Calendar.MINUTE, 0)
+                    set(Calendar.SECOND, 0)
+                    set(Calendar.MILLISECOND, 0)
+                }.timeInMillis
+            } else try {
+                SimpleDateFormat("d MMM yyyy", Locale.ITALIAN).parse(date).time
```
paronos: Dates are like _"Jul 7, 2015"_

en-perveden/src/eu/kanade/tachiyomi/extension/en/perveden/Perveden.kt
```diff
@@ -0,0 +1,161 @@
...
+            } else if ("Yesterday" in date) {
+                Calendar.getInstance().apply {
+                    add(Calendar.DATE, -1)
+                    set(Calendar.HOUR_OF_DAY, 0)
+                    set(Calendar.MINUTE, 0)
+                    set(Calendar.SECOND, 0)
+                    set(Calendar.MILLISECOND, 0)
+                }.timeInMillis
+            } else try {
+                SimpleDateFormat("d MMM yyyy", Locale.ITALIAN).parse(date).time
```
paronos: Dates are like _"Jul 7, 2015"_.

it-perveden/src/eu/kanade/tachiyomi/extension/it/perveden/Perveden.kt
```diff
@@ -0,0 +1,166 @@
...
+                        .filter { it.state }
+                        .map { it.id.toString() }
+                        .forEach { url.addQueryParameter("status", it) }
+                is Types -> filter.state
+                        .filter { it.state }
+                        .map { it.id.toString() }
+                        .forEach { url.addQueryParameter("type", it) }
+                is GenreList -> filter.state
+                        .filterNot { it.isIgnored() }
+                        .forEach { genre -> url.addQueryParameter(if (genre.isIncluded()) "categoriesInc" else "categoriesExcl", genre.id) }
```
paronos: Can be removed if genres are not added.

it-perveden/src/eu/kanade/tachiyomi/extension/it/perveden/Perveden.kt
```diff
@@ -0,0 +1,166 @@
...
+    override fun pageListParse(document: Document): List<Page> = mutableListOf<Page>().apply {
+        document.select("option[value^=/it/it-manga/]").forEach {
+            add(Page(size, "$baseUrl${it.attr("value")}"))
+        }
+    }
+
+    override fun imageUrlParse(document: Document): String = document.select("a#nextA.next > img").first()?.attr("src").let { "http$it" }
+
+    private class NamedId(name: String, val id: Int) : Filter.CheckBox(name)
+    private class Genre(name: String, val id: String) : Filter.TriState(name)
```
paronos: Can be removed if genres are not added.

it-perveden/src/eu/kanade/tachiyomi/extension/it/perveden/Perveden.kt
```diff
@@ -0,0 +1,166 @@
...
+    private class TextField(name: String, val key: String) : Filter.Text(name)
+    private class OrderBy : Filter.Sort("Ordina per", arrayOf("Titolo manga", "Visite", "Capitoli", "Ultimo capitolo"),
+            Filter.Sort.Selection(1, false))
+
+    private class StatusList(statuses: List<NamedId>) : Filter.Group<NamedId>("Stato", statuses)
+    private class Types(types: List<NamedId>) : Filter.Group<NamedId>("Tipo", types)
+    private class GenreList(genres: List<Genre>) : Filter.Group<Genre>("Generi", genres)
```
paronos: Can be removed if genres are not added.

en-mangaeden/src/eu/kanade/tachiyomi/extension/en/mangaeden/Mangaeden.kt
```diff
@@ -0,0 +1,203 @@
...
+                        .filter { it.state }
+                        .map { it.id.toString() }
+                        .forEach { url.addQueryParameter("status", it) }
+                is Types -> filter.state
+                        .filter { it.state }
+                        .map { it.id.toString() }
+                        .forEach { url.addQueryParameter("type", it) }
+                is GenreList -> filter.state
+                        .filter { !it.isIgnored() }
+                        .forEach { genre -> url.addQueryParameter(if (genre.isIncluded()) "categoriesInc" else "categoriesExcl", genre.id) }
```
paronos: Can be removed if genres are not added.

en-mangaeden/src/eu/kanade/tachiyomi/extension/en/mangaeden/Mangaeden.kt
```diff
@@ -0,0 +1,203 @@
...
+    override fun pageListParse(document: Document): List<Page> = mutableListOf<Page>().apply {
+        document.select("option[value^=/en/en-manga/]").forEach {
+            add(Page(size, "$baseUrl${it.attr("value")}"))
+        }
+    }
+
+    override fun imageUrlParse(document: Document): String = document.select("a#nextA.next > img").first()?.attr("src").let { "http:$it" }
+
+    private class NamedId(name: String, val id: Int) : Filter.CheckBox(name)
+    private class Genre(name: String, val id: String) : Filter.TriState(name)
```
paronos: Can be removed if genres are not added.

en-mangaeden/src/eu/kanade/tachiyomi/extension/en/mangaeden/Mangaeden.kt
```diff
@@ -0,0 +1,203 @@
...
+    private class TextField(name: String, val key: String) : Filter.Text(name)
+    private class OrderBy : Filter.Sort("Order by", arrayOf("Manga title", "Views", "Chapters", "Latest chapter"),
+            Filter.Sort.Selection(1, false))
+
+    private class StatusList(statuses: List<NamedId>) : Filter.Group<NamedId>("Stato", statuses)
+    private class Types(types: List<NamedId>) : Filter.Group<NamedId>("Tipo", types)
+    private class GenreList(genres: List<Genre>) : Filter.Group<Genre>("Genres", genres)
```
paronos: Can be removed if genres are not added.

en-mangaeden/src/eu/kanade/tachiyomi/extension/en/mangaeden/Mangaeden.kt
blackgin: Genres have been removed only for perveden

paronos:
Since I do not know how to make a selection of the most important genres, I thought these solutions:
- Add different GenreList by genre first letter (e.g. "Genres A-D", "Genres E-H", etc.). The UI will be still full of entries but more usable. Maybe it can only work for the Italian version (229 genres) but not for the English version (1156 genres).
- Use two text filed ("Include genres:" and "Exclude genres:") for the genre selection. The user will have to write the different genres (separated by commas) and the source will map them to the search ids by using an HashMap. If the written genre is not present in the hashmap the source can either return an error message or ignore it.

paronos:
I tried to add gender as text fields (see this file: [Perveden.kt](https://github.com/inorichi/tachiyomi-extensions/files/762009/Perveden.kt.txt)), I think it's better than nothing.

blackgin:
Hi I'm back! 
Added your implementation and fixed the page image url, where I forgot the colon. It is bugged also on master

paronos:
I tried to do the same thing for the English version (see file: [Perveden.kt](https://github.com/inorichi/tachiyomi-extensions/files/769607/Perveden.kt.txt)). In this case there are some duplicate keys (genres), but I don’t think it worthwhile to do something better.

blackgin:
Wow these sites are really a mess. Going to add this change too, thanks!

blackgin:
So, if we're fine with the current implementation and if there aren't any issues left, I think this can be merged.

By the way, speaking about the new text filters, I think it would be great if there was support for autocompletion, so that in cases like these, the filter would be more usable.

paronos:
It should be fairly simple to add autocompletion by using `AutoCompleteTextView` and `MultiAutoCompleteTextView` (or `AppCompatAutoCompleteTextView` and `AppCompatMultiAutoCompleteTextView`).
However, this would require to change `Filter` in the extensions library, in particular something like
```kotlin
abstract class AutoCompleteText(name: String, val adapter : ListAdapter, state: String = "") : Filter<String>(name, state)
abstract class MultiAutoCompleteText(name: String, val adapter : ListAdapter, val tokenizer : MultiAutoCompleteTextView.Tokenizer, state: String = "") : Filter<String>(name, state)
```
another option could be to add `adapter` and `tokenizer` directly to `Filer.Text`:
```kotlin
abstract class Text(name: String, val adapter : ListAdapter?, val tokenizer : MultiAutoCompleteTextView.Tokenizer?, state: String = "") : Filter<String>(name, state)
```

The only problem I see in this solution is that the autocompletion would be particularly useful for searching by title (i.e. MangaFox has a similar function that could be integrated), but the title search field is the only one not handled by a filter.

inorichi:
I think only a few sites would benefit from autocompletion, and it might be ugly to integrate. I prefer not to have that.

PR #15, opened 2017-02-19
Create ISSUE_TEMPLATE.md

j2ghz:
It could also say put website name in Title, I'm not sure which is better


inorichi:
I would say name of the site in title and clickable url in the description

PR #18, opened 2017-02-26
Add ReadComicOnline

j2ghz:
copied kissmanga, fixed few things that were different

### DIFF
--- a/src/en/readcomiconline/build.gradle
+++ b/src/en/readcomiconline/build.gradle
@@ -1,0 +1,13 @@
+apply plugin: 'com.android.application'
+apply plugin: 'kotlin-android'
+
+ext {
+    appName = 'Tachiyomi: ReadComicOnline'
+    pkgNameSuffix = "en.readcomiconline"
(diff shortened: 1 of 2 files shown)


src/en/readcomiconline/build.gradle
```diff
@@ -0,0 +1,13 @@
+apply plugin: 'com.android.application'
+apply plugin: 'kotlin-android'
+
+ext {
+    appName = 'Tachiyomi: ReadComicOnline'
+    pkgNameSuffix = "en.readcomiconline"
+    extClass = '.readcomiconline'
```
inorichi: Use a capital here.

j2ghz:
```java
pkgNameSuffix = "en.Readcomiconline"
extClass = '.Readcomiconline'
```
like this? or en as well?

inorichi: only `extClass`, package names are lower case.

PR #20, opened 2017-02-26
Add travis and badges

j2ghz:


inorichi:
I'm not sure about badges in this repository. In my opinion they don't add any value.

j2ghz:
You are right with the download. I could also ditch the table. The build badge is useful, I check it whenever my CI fails to build.

inorichi:
Ok, keep only the build badge.

PR #46, opened 2017-05-25
Add Hentai2Read (hentai2read.com) source.

ddmgy:


src/en/hentai2read/src/eu/kanade/tachiyomi/extension/en/hentai2read/Hentai2Read.kt
```diff
@@ -0,0 +1,578 @@
...
+        val infoElement = document.select("ul.list-simple-mini").first()
+
+        val manga = SManga.create()
+        manga.author = infoElement.select("li:contains(Author) > a").first()?.text()
+        manga.artist = infoElement.select("li:contains(Artist) > a").first()?.text()
+        var tags = mutableListOf<String>()
+        infoElement.select("li:contains(Category) > a").forEach {
+            tags.add(it.text())
+        }
+        infoElement.select("li:contains(Content) > a").forEach {
```
null-dev: You can probably combine these two select calls into one with this selector: `li:contains(Category) > a, li:contains(Content) > a`

src/en/hentai2read/src/eu/kanade/tachiyomi/extension/en/hentai2read/Hentai2Read.kt
```diff
@@ -0,0 +1,578 @@
...
+        status.contains("Completed") -> SManga.COMPLETED
+        else -> SManga.UNKNOWN
+    }
+
+    override fun chapterListSelector() = "ul.nav-chapters li a.link-effect"
+
+    override fun chapterFromElement(element: Element): SChapter {
+        val chapter = SChapter.create()
+        chapter.setUrlWithoutDomain(element.attr("href"))
+        chapter.name = element.text().split(" by ").first().trim()
```
null-dev: You can use `element.ownText()` on `element` to get only the text of that element and exclude the text of any child elements.

src/en/hentai2read/src/eu/kanade/tachiyomi/extension/en/hentai2read/Hentai2Read.kt
```diff
@@ -0,0 +1,578 @@
...
+            parseChapterDate(it)
+        } ?: 0L
+        return chapter
+    }
+
+    private fun parseChapterDate(date: String): Long {
+        val dateWords = date.split(" ")
+        if (dateWords.size == 3) {
+            val timeAgo = Integer.parseInt(dateWords[0])
+            return Calendar.getInstance().apply {
```
null-dev: Shorten this block to something along the lines of the code below:
```
Calendar.getInstance().apply {
    when (dateWords[1]) {
	    "minute", "minutes" -> Calendar.MINUTE
	    "hour", "hours" -> Calendar.HOUR
	    "day", "days" -> Calendar.DAY_OF_YEAR
	    "week", "weeks" -> Calendar.WEEK_OF_YEAR
	    "month", "months" -> Calendar.MONTH
	    "year", "years" -> Calendar.YEAR
	    else -> null
    }?.let {
	    add(it, -timeAgo)
    }
}.timeInMillis
```

src/en/hentai2read/src/eu/kanade/tachiyomi/extension/en/hentai2read/Hentai2Read.kt
```diff
@@ -0,0 +1,578 @@
...
+        return 0L
+    }
+
+    override fun pageListRequest(chapter: SChapter) = POST(baseUrl + chapter.url, headers)
+
+    override fun pageListParse(response: Response): List<Page> {
+        val pages = mutableListOf<Page>()
+        val imageBaseUrl = "https://static.hentaicdn.com/hentai"
+        //language=RegExp
+        val p = Pattern.compile("""'images' : \[\"(.*?)[,]?\"\]""")
```
null-dev: Same advice as in the previous PR. Compile the pattern once into a variable outside of this function (maybe use a lazy delegate?) and just use that variable when matching with the pattern.

src/en/hentai2read/src/eu/kanade/tachiyomi/extension/en/hentai2read/Hentai2Read.kt
```diff
@@ -0,0 +1,578 @@
...
+            }.timeInMillis
+        }
+        return 0L
+    }
+
+    override fun pageListRequest(chapter: SChapter) = POST(baseUrl + chapter.url, headers)
+
+    override fun pageListParse(response: Response): List<Page> {
+        val pages = mutableListOf<Page>()
+        val imageBaseUrl = "https://static.hentaicdn.com/hentai"
```
null-dev: This is a constant so it should be moved outside this function (maybe into the companion object even?).

ddmgy:
I've changed the things you mentioned in your comments.

I have a question, however. When I read a chapter with this extension, the percentage that shows while a page is downloading counts down from 0 to -1000000 or so (a large negative number), rather than up from 0 to 100. If I scroll through to the next few images before they finish downloading, the negative counter keeps increasing. It does not seem to actually affect performance or cause any issues while downloading images, but I only ever see this behavior with this extension, and it happens on every chapter.

I can't think of what could be causing this, but I'm not familiar with the Tachiyomi codebase, so I don't even know where to begin looking.

null-dev:
Did this start happening after you applied my suggestions or was it like this before?

ddmgy:
It's been happening since I first wrote the extension.

null-dev:
Ok, that makes sense.

Basically, the hentai2read server does not send the 'Content-Length' header when sending images to Tachiyomi.

Without the 'Content-Length' header, Tachiyomi has no idea what the total size of each image is and thus, cannot properly calculate the download percentage because it doesn't know how much bytes it has to download!

I don't think there is a way to fix this other than contacting hentai2read but it isn't a major bug anyways so don't worry too much about it.

ddmgy:
Ah, okay. I hadn't even thought of the headers, but it makes total sense. Thank you!

ddmgy:
Hentai2Read updated their search options last night, so I've changed the extension to use them.

I added the Doujin tags that are available in Hentai2Read's search (which I avoiding doing before as it was much harder before the update to find the IDs for each tag), in order to be as complete as possible. If this is not wanted (it does add 800+ tags to the search filter), it can be removed with no trouble.

ddmgy:
Okay, so I've got something of an issue. I noticed that since Hentai2Read changed their advanced search (as I wrote in my last message), they add a Base64-encoded string (for example, http://hentai2read/advanced-search/<base64_string>/name-az/<page_number>/) to the advanced search page. Currently, this extension will correctly show the first page of results when searching, but it will then repeat these results when trying to load page 2 and onward.

I've worked out how the Base64-encoded string is generated. The POST parameters are put in a Json object, by default:

`{
    "nme_opr": "1",
    "nme": "",
    "ats_opr": "1",
    "ats": "",
    "chr_opr": "1",
    "chr": "",
    "tag": {
        "inc": [],
        "exc": [],
        "mde": "0"
    },
    "rls_yer_opr": "0",
    "rls_yer": "",
    "sts": "0"
}`

Which is then transformed to a string (not really sure if this is a known format; I've never seen it before):

`a:10:{s:7:"nme_opr";s:1:"1";s:3:"nme";s:0:"";s:7:"ats_opr";s:1:"1";s:3:"ats";s:0:"";s:7:"chr_opr";s:1:"1";s:3:"chr";s:0:"";s:3:"tag";a:3:{s:3:"inc";a:0:{}s:3:"exc";a:0:{}s:3:"mde";s:1:"0";}s:11:"rls_yer_opr";s:1:"0";s:7:"rls_yer";s:0:"";s:3:"sts";s:1:"0";}`

And then encoded in Base64:

`YToxMDp7czo3OiJubWVfb3ByIjtzOjE6IjEiO3M6Mzoibm1lIjtzOjA6IiI7czo3OiJhdHNfb3ByIjtzOjE6IjEiO3M6MzoiYXRzIjtzOjA6IiI7czo3OiJjaHJfb3ByIjtzOjE6IjEiO3M6MzoiY2hyIjtzOjA6IiI7czozOiJ0YWciO2E6Mzp7czozOiJpbmMiO2E6MDp7fXM6MzoiZXhjIjthOjA6e31zOjM6Im1kZSI7czoxOiIwIjt9czoxMToicmxzX3llcl9vcHIiO3M6MToiMCI7czo3OiJybHNfeWVyIjtzOjA6IiI7czozOiJzdHMiO3M6MToiMCI7fQ==`

[Here is a gist](https://gist.github.com/ddmgy/44febc71fbd1207f1745f49ccaca3bc3) of how I accomplish this in the extension. It uses Gson and Kotson, both of which are provided by Tachiyomi proper (and used by a couple of other extensions), so it does not add any further dependencies. It works as expected, it's fast, and it does not seem to affect performance too much.

My problems are: it's probably the ugliest thing I've ever written, and I have to generate it every time searchMangaRequest is called. I can't think of a way to generate it once for every time the user searches the catalogue.

I see two potential paths I can take. One, I can use what I've written, hopefully pared down a little and only computed once for a given FilterList so as not to affect performance; though, as I said, it does not seem to be doing that much to the performance.

Two, I can remove all of this, remove all of the filters, and rely on a simple text search. That requires no Base64 strings, using the URL form "http://hentai2read.com/hentai-list/search/$query/name-az/$page". This will significantly reduce the size of this extension, and simplify searching in the catalogue. So far, there are only two people using this extension (myself and the friend who asked me to write the extension), and the person in #54 may use it at some point.

It seems an awful lot of trouble and bloat for potentially only 3 users who may not ever need the advanced search features.

Anyway, sorry for the very long message. Any suggestions you may have would be much appreciated.

cd0rw0:
since its you already getting it to work, without much impact to performance and you still have additional things to do which expected to improve on current app. i will suggest you release wt advance search.

ddmgy:
Alright, I've updated [my earlier gist](https://gist.github.com/ddmgy/44febc71fbd1207f1745f49ccaca3bc3) with some new changes. The Base64 string is now computed only on the first call to searchMangaRequest for a given FilterList (when page == 1), and then reused on each subsequent call. I'm not fond of the function names (encode, and h2rEncode), but I cannot think of any better.

null-dev:
@ddmgy Why not get the encoded base64 string from the first request and store it for future requests?

If you have to store the string in the companion object anyways, I think you could probably scrape the page for the encoded base64 string (it's in the pagination links) instead of encoding it manually.

ddmgy:
@null-dev Well, I just hadn't thought of the base64 string being in the links in the search page. I was certainly overthinking how to do this.

I'll look into it tonight. Thank you for the suggestion.

ddmgy:
Well, that was embarrassingly easy. The extension now grabs the base64 string from the link to next page of search results.

Thanks again for pointing me in the right direction, @null-dev.

PR #47, opened 2017-05-25
Add ShoujoSense (reader.shoujosense.com) extension.

ddmgy:


src/en/shoujosense/src/en/kanade/tachiyomi/extension/en/shoujosense/ShoujoSense.kt
```diff
@@ -0,0 +1,149 @@
...
+
+    override val supportsLatest = true
+
+    override val client: OkHttpClient = network.cloudflareClient
+
+    override fun popularMangaSelector() = "div.list div.group"
+
+    override fun latestUpdatesSelector() = popularMangaSelector()
+
+    override fun popularMangaRequest(page: Int): Request {
```
null-dev: Just a note that these can also be shortened into something like:
```
override fun popularMangaRequest(page: Int)
    = GET("$baseUrl/directory/$page", headers)
```
Doesn't really affect performance though and it's mainly just about taste.

src/en/shoujosense/src/en/kanade/tachiyomi/extension/en/shoujosense/ShoujoSense.kt
```diff
@@ -0,0 +1,149 @@
...
+        return GET("$baseUrl/directory/$page", headers)
```
null-dev: Just a note that these can also be shortened into something like:
```
override fun popularMangaRequest(page: Int)
    = GET("$baseUrl/directory/$page", headers)
```
Doesn't really affect performance though and it's mainly just about taste.

src/en/shoujosense/src/en/kanade/tachiyomi/extension/en/shoujosense/ShoujoSense.kt
```diff
@@ -0,0 +1,149 @@
...
+    }
+
+    override fun latestUpdatesRequest(page: Int): Request {
+        return GET("$baseUrl/latest/$page", headers)
+    }
+
+    override fun popularMangaFromElement(element: Element): SManga {
+        val manga = SManga.create()
+        element.select("div.title a:eq(0)").first().let {
+            manga.setUrlWithoutDomain(it.attr("href"))
```
null-dev: If the `select` method used here will always only return one element, you don't need to write `first()`. Just call `attr()` or `text()` on the array of elements directly and it will return the same result.

There are several instances in this file where this same concept can be applied.

src/en/shoujosense/src/en/kanade/tachiyomi/extension/en/shoujosense/ShoujoSense.kt
```diff
@@ -0,0 +1,149 @@
...
+    }
+
+    override fun chapterListSelector() = "div.list div.element"
+
+    override fun chapterFromElement(element: Element): SChapter {
+        val urlElement = element.select("a").first()
+
+        val chapter = SChapter.create()
+        chapter.setUrlWithoutDomain(urlElement.attr("href"))
+        print(chapter.url)
```
null-dev: This should be removed. I'm assuming this is only used during debugging.

src/en/shoujosense/src/en/kanade/tachiyomi/extension/en/shoujosense/ShoujoSense.kt
```diff
@@ -0,0 +1,149 @@
...
+        chapter.name = urlElement.text()
+        chapter.date_upload = element.select("div.meta_r").first()?.text()?.split(", ")?.get(1)?.let {
```
null-dev: Instead of `split(", ")?.get(1)`, you can probably use `substringAfterLast(", ")` here.

src/en/shoujosense/src/en/kanade/tachiyomi/extension/en/shoujosense/ShoujoSense.kt
```diff
@@ -0,0 +1,149 @@
...
+    private fun parseChapterDate(date: String): Long {
+        return if ("Today" in date) {
+            Calendar.getInstance().timeInMillis
+        } else if ("Yesterday" in date) {
+            Calendar.getInstance().apply {
+                add(Calendar.DAY_OF_YEAR, -1)
+            }.timeInMillis
+        } else {
+            try {
+                SimpleDateFormat("yyyy.MM.dd").parse(date).time
```
null-dev: Avoid initializing `SimpleDateFormat` for every call to `parseChapterDate`. It's thread safe and immutable so store your `SimpleDateFormat` in a variable outside the function and reference it every time you want to parse a date.

src/en/shoujosense/src/en/kanade/tachiyomi/extension/en/shoujosense/ShoujoSense.kt
```diff
@@ -0,0 +1,149 @@
...
+        }
+    }
+
+    override fun pageListRequest(chapter: SChapter) = POST(baseUrl + chapter.url, headers)
+
+    override fun pageListParse(response: Response): List<Page> {
+        val body = response.body().string()
+        val pages = mutableListOf<Page>()
+
+        val p = Pattern.compile(""""url":"(.*?)"""")
```
null-dev: I think you can do the same thing I mentioned with `SimpleDateFormat` with this `Pattern.compile` statement as well. Keep a single copy of the compiled pattern.

src/en/shoujosense/src/en/kanade/tachiyomi/extension/en/shoujosense/ShoujoSense.kt
```diff
@@ -0,0 +1,149 @@
...
+        return pages
+    }
+
+    override fun pageListParse(document: Document): List<Page> {
+        throw Exception("Not used")
+    }
+
+    override fun imageUrlRequest(page: Page) = GET(page.url)
+
+    override fun imageUrlParse(document: Document) = ""
```
null-dev: You can have this return `null` instead of an empty string if you want as well.

ddmgy:
Thank you for your suggestions, @null-dev. This is my first foray into Kotlin, so any tips are much appreciated :)

I've gone through and (hopefully) fixed and/or changed most of the things on which you commented.

src/en/shoujosense/src/en/kanade/tachiyomi/extension/en/shoujosense/ShoujoSense.kt
ddmgy: According to the method definition in Tachiyomi for [ParsedHttpSource.imageUrlParse](https://github.com/inorichi/tachiyomi/blob/ca2236958a3c06663c042d904c446763cfd18d43/app/src/main/java/eu/kanade/tachiyomi/source/online/ParsedHttpSource.kt#L199), the type must be a String. So, I don't think I can return null here. The type would have to be String?, right?

Again, I'm new to Kotlin, so if I am misunderstanding what you mean, I apologize.

null-dev: Oops, nice catch! I somehow managed to misread `imageUrlParse` as `imageUrlNextPageSelector` (which doesn't even exist :sweat:). Then yeah this line is just fine.

PR #63, opened 2017-06-17
Fix various bugs in MangaPark source

### LINKED ISSUE #61: Parsing error MangaPark
So I tried to refresh the list of chapters of the Manga "DICE the cube..." and it gives an error:

```
unparseable date: "yesterday 08:24 am" (at offset 0) 
```
Plugin was downloaded from f-stop.  

_Sent from my Samsung SM-T320 using [FastHub](https://play.google.com/store/apps/details?id=com.fastaccess.github)_

null-dev:
This PR contains two bug fixes for the MangaPark source:
- Fix broken date parsing for some chapters
- Fix the next page selector selecting the wrong elements when searching
  (This resulted in "No results found" messages when moving between pages)

Resolves #61, #82


null-dev:
I think this is ready.

PR #75, opened 2017-07-04
Added DynastyScans

ThePreviousOne:
I think it works though I am still testing

### DIFF
--- a/src/en/dynasty/src/eu/kanade/tachiyomi/extension/en/dynasty/DynastyScans.kt
+++ b/src/en/dynasty/src/eu/kanade/tachiyomi/extension/en/dynasty/DynastyScans.kt
@@ -1,0 +1,232 @@
+package eu.kanade.tachiyomi.source.online.english.dynasty
+
+import android.net.Uri
+import android.util.Log
+import eu.kanade.tachiyomi.network.GET
+import eu.kanade.tachiyomi.source.model.*
+import eu.kanade.tachiyomi.source.online.ParsedHttpSource
+import eu.kanade.tachiyomi.util.asJsoup
+import okhttp3.*
+import org.json.JSONArray
+import org.jsoup.nodes.Document
+import org.jsoup.nodes.Element
+import org.jsoup.nodes.Node
+import org.jsoup.nodes.TextNode
+import java.text.SimpleDateFormat
+import java.util.*
+import java.util.regex.Pattern
+
+abstract class DynastyScans() : ParsedHttpSource() {
+
+    override val baseUrl = "http://dynasty-scans.com"
+
+    abstract fun popularMangaInitialUrl(): String
+
+    override val lang = "en"
+
+    override val supportsLatest = false
+
+    var parent: List<Node> = ArrayList()
+
+    var list = InternalList(ArrayList(), "")
+
+    var imgList = InternalList(ArrayList(), "")
+
+    var _valid: Validate = Validate(false, -1)
+
+    override fun popularMangaRequest(page: Int): Request {
+        return GET(popularMangaInitialUrl(), headers)
+    }
+
+    override fun popularMangaSelector() = "ul.thumbnails > li.span2"
+
+    override fun popularMangaFromElement(element: Element): SManga {
+        val manga = SManga.create()
+        manga.setUrlWithoutDomain(element.select("a").attr("href"))
+        manga.title = element.select("div.caption").text()
+        return manga
+    }
+
+    override fun popularMangaParse(response: Response): MangasPage {
+        val mangas = response.asJsoup().select(popularMangaSelector()).map { element ->
+            popularMangaFromElement(element)
+        }
+        return MangasPage(mangas, false)
+    }
+
+    override fun searchMangaSelector() = "a.name"
+
+    override fun searchMangaFromElement(element: Element): SManga {
+        val manga = SManga.create()
+            manga.setUrlWithoutDomain(element.attr("href"))
+            manga.title = element.text()
+        return manga
+    }
+
+    override fun searchMangaNextPageSelector() = "div.pagination > ul > li.active + li > a"
+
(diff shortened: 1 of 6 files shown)


ThePreviousOne:
Im confident with this now

src/en/dynasty/src/eu/kanade/tachiyomi/extension/en/dynasty/DynastyScans.kt
```diff
@@ -0,0 +1,232 @@
...
+    }
+
+    protected fun parseThumbnail(manga: SManga) {
+        if (_valid.isManga) manga.thumbnail_url = baseUrl + imgList[_valid.pos].substringBefore('?')
+    }
+
+    protected fun parseHeader(document: Document, manga: SManga) : Boolean {
+        val elements = document.select("div.tags > h2.tag-title").first().getElementsByTag("a")
+
+        if (!elements.isEmpty()) {
```
igalic: this can be simplified for readability to
```kotlin
if (elements.isEmpty()) {
    return false
}
```

ThePreviousOne: no it can't

ThePreviousOne: If you want to try it go ahead, but I'm not

Oxdeception: There is no need for an else block.  Placing the contents of the current if block after the one provided by igalic should work.

Otend:
What's the current status on this?

igalic:
i don't know!
i still think @inorichi should just merge it :)

sirusi:
hey so I was just trying to add this as a source on my phone but I couldn't get it to work.
I built an .apk using Android Studio and OP's Fork, it installed normally but source did not show up.
I also tried other forks like the Mangastream pull request and it worked fine. I did the same process on both.
Does this need any special thing to work? Just wanted to get Dynasty working. sorry if wrong place!

Otend:
Same issue as sirusi here.  Built and installed successfully, but no dice in terms of appearing on the current dev release.

inorichi:
The package name of the classes should be `eu.kanade.tachiyomi.extension.en.dynasty` instead of `eu.kanade.tachiyomi.source.online.english.dynasty`

sirusi:
@inorichi tried that, but still did not work. Any other ideas?

drac0300:
I have the same issue, after installing the apk I built, nothing changed in tachiyomi... is there any chance this is due to tachiyomi and the extension apk being signed with different keys ?

sirusi:
ok found the issue. In addition to what inorichi said above:
In build.gradle, `+    extClass = '.DynastyAnthologies; .DynastyIssues; .DynastySeries: .DynastyDoujins'`
should be `+    extClass = '.DynastyAnthologies; .DynastyIssues; .DynastySeries; .DynastyDoujins'`.

The problem is a semicolon.

drac0300:
It does compile and the sources are accessible after that. Nethertheless, when trying to read a manga, you get an «unparsable date» error, and chapters don’t seem to be accessible.

src/en/dynasty/src/eu/kanade/tachiyomi/extension/en/dynasty/DynastyScans.kt
drommk: Returning as early as possible is in most cases the best option. It increases readability and reduces code complexity, except in binary cases. I'm with @igalic 

Doesn't sound like a showstopper to me though

sirusi:
@drac0300 I've been using the extension after my latest message but I didn't encounter that error yet. Have downloaded and read around 5 or 6 manga. I'm using Android 6.0 (motorola).

drac0300:
that’s probably normal : it turns out it was a locale problem, if your locale is an english one, you will never see it.

EndxAwesome14:
Sorry to ask a dumb question, how do you add Dynasty scans to Tachiyomi? Im new to all of this kotlin stuff

drac0300:
Basically you need to clone ThePreviousOne's forked repository, and then build the plugin into an apk and sign that apk.
Android studio can automate these last steps but I don't really know how to use it, I generally prefer to do things by hand… here's a tutorial though
https://developer.android.com/studio/run/index.html

EndxAwesome14:
When you mean by "cloning" you mean just pasting the code into Android Studio and build a apk?

drac0300:
I meant a git clone. That's how you download a git repository

EndxAwesome14:
Thank you for your help! I'm sorry for bothering you, I'm such a noob when it comes to this ;_;

EndxAwesome14:
Wait one more thing, when I generated a signed apk and installed on my device, it didn't so up in tachiyomi, why is that?

drac0300:
@EndxAwesome14 have you fixed the error reported above by Sirusi ? If you have, then I don't know what the problem is, it works fine here

EndxAwesome14:
@drac0300 Yes, I added the semicolon to the ...DynastySeries:, and still it didn't so up in my cataloges

readher:
Aside from adding the semicolon, you need to change the path at the top of one file to correctly resemble folder structure it's in. After changing those two things I built an .apk and successfully installed the plugin on two devices for Tachiyomi version 0.6.5. After updating to 0.6.6, Dynasty is still present in Catalogues and works as intended. I guess I could send you my .apk somehow if it'd help?

EndxAwesome14:
Yes, just give me a link for the .apk, thank you

readher:
[Here it is](https://drive.google.com/file/d/1XZlEg4VE_b6xMm0EI4A76mmErS84EPUx/view?usp=sharing)

EndxAwesome14:
Thanks

PR #90, opened 2017-08-12
Fix hentai2read search (fixes #89)

### LINKED ISSUE #89: Hentai2Read - Search and filters are not working
Tachiyomi version: 0.6.1-1197
Source version: 1.0.1
Android version: 4.4.2

Entering any text or changing any filters on this source returns the same results as directly visiting https://hentai2read.com/hentai-list/advanced-search/

ddmgy:
Fixes an issue with searching/filtering in the Hentai2Read source.

Hentai2Read used to serve most data over   HTTP (including search results), but they have now moved to only serving over HTTPS. Now, I can safely used https in the baseUrl without having to rely on occasional redirects.

Fixes issue #89.


src/en/hentai2read/build.gradle
```diff
@@ -5,7 +5,7 @@ ext {
     appName = 'Tachiyomi: Hentai2Read'
     pkgNameSuffix = 'en.hentai2read'
     extClass = '.Hentai2Read'
-    extVersionCode = 1
```
inorichi: Is there a reason for skipping a version?

ddmgy: I initially bumped it when I first opened the pull request on 11 August, then again when I changed a few more things on 30 August. It can be reverted to 2 if you like, it just made sense to me to change it at the time.

inorichi: If the other version hasn't been merged, don't update it again.

ddmgy: Right you are, change has been reverted.

PR #156, opened 2018-01-12
Add Madokami source

bobobo1618:


src/en/madokami/build.gradle
```diff
@@ -0,0 +1,13 @@
+apply plugin: 'com.android.application'
+apply plugin: 'kotlin-android'
+
+ext {
+    appName = 'Tachiyomi: Madokami'
+    pkgNameSuffix = 'en.madokami'
+    extClass = '.Madokami'
+    extVersionCode = 1
+    extVersionSuffix = 1
+    libVersion = '1.4'
```
AmineI: Here, the libVersion being set to 1.4 (??) while the last one is 1.2 is making the Travis build fail as you can see in Travis's log. I believe it's the only issue preventing the build.

arkon: It's supposed to go with https://github.com/inorichi/tachiyomi-extensions-lib/pull/4

AmineI: Oh right, didn't notice that pull request on the lib repo. Indeed, it was intended not to pass without that PR being accepted. Ignore my comment then, I'll see if I can edit/remove it, just wanted to help :) .

as280093:
If the PR get fixed here are the res so merge while fixing

[res.zip](https://github.com/inorichi/tachiyomi-extensions/files/2937622/res.zip)
ic_launcher

PR #159, opened 2018-01-15
Fix MangaPark image URLs and some code cleanup

### LINKED ISSUE #158: MangaPark issue: Broken covers in Latest
When browsing 'Latest', most of the covers are completely broken. This doesn't happen in 'Browse'.
screenshot_20180116-023953

null-dev:
Resolves #158

### DIFF
--- a/src/en/mangapark/build.gradle
+++ b/src/en/mangapark/build.gradle
@@ -5,7 +5,7 @@ ext {
     appName = 'Tachiyomi: MangaPark'
     pkgNameSuffix = "en.mangapark"
     extClass = '.MangaPark'
-    extVersionCode = 1
+    extVersionCode = 2
     extVersionSuffix = 2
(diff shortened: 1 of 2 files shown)


inorichi:
Can you upgrade the version?

inorichi:
F-Droid won't show updates if version code isn't upgraded. Sorry for not being clear enough last time :(

null-dev:
@inorichi Yeah, I understood it clear enough. I have upgraded the version.

inorichi:
I meant, version code too. You only upgraded version name, which will show a different version label but AFAIK F-Droid only uses version code to display updates

null-dev:
@inorichi So `versionSuffix` = `versionName`?

inorichi:
Only the last part. `libVersion` has the first two.

Issue #161, opened 2018-01-18
[Site Request] Mangadex


wopian:
https://mangadex.com/
https://www.reddit.com/r/manga/comments/7r4dmn/introducing_mangadex/

Created by FansubDB/Anidex creators and an alternative to batoto

asdkant:
+1

PR #163, opened 2018-01-19
add mangadex

CarlosEsco:
Added mangadex #161. Their site is still developing so pagination, language selection, genres etc are missing for now.  Probably will break as they keep adding features but i'll keep an eye out for issues


Kovaelin:
Thank you!

PR #103, opened 2017-09-16
Add support for My Manga Reader CMS sources (many, many sources)

null-dev:
Sources added by this extension:
- AR: مانجا اون لاين
- AR: Manga FYI
- EN: Read Comics Online
- EN: Fallen Angels Scans
- EN: Mangawww Reader
- EN: MangaForLife
- EN: Manga Spoil
- EN: DManga
- EN: Chibi Manga Reader
- EN: ZXComic
- EN: Mangacox
- EN: KoManga
- EN: Manganimecan
- EN: Hentai2Manga
- EN: White Cloud Pavilion
- EN: 4 Manga
- EN: XYXX.INFO
- ES: My-mangas.com
- ES: SOS Scanlation
- FR: Manga-LEL
- FR: Manga Etonnia
- FR: Scan FR
- FR: ScanFR.com
- FR: Manga FYI
- FR: Mugiwara
- FR: scans-manga
- ID: Manga Desu
- ID: MangaOnline
- ID: Komikid
- ID: MangaID
- ID: Manga Seru
- ID: Manga FYI
- JA: Mangazuki RAWS
- JA: MangaRAW
- PL: Candy Scans
- PL: ToraScans
- PT: Comic Space
- PT: Mangás Yuri
- RU: NAKAMA
- TR: MangAoi
- TR: MangaHanta
- TR: ManhuaTR
- VI: Fallen Angels Scans
- OTHER: HentaiShark

I found out that a bunch of popular websites use the same CMS software!

Instead of writing a source for every manga website that uses [My Manga Reader CMS](https://getcyberworks.com/product/manga-reader-cms/), I decided to write a generator that would generate them for me! I've enabled generation for the sources above, there may still be more sources that support generation.

*If you find a source that uses My Manga Reader CMS (there are many here: <https://www.google.ca/search?q="Powered+by+my+Manga+Reader+CMS">), please help me out by posting the generation code for it!* It's easy:

1. Find the two letter language code for the language website. Use the ISO 639-1 column on the list here: https://www.loc.gov/standards/iso639-2/php/code_list.php
2. Post a comment with the generation code, in the following format:
    ```
    gen "LANGUAGE CODE" "NAME OF WEBSITE" "WEBSITE URL"
    ```
    **Example for Read Comics Online:**
    ```
    gen "en" "Read Comics Online" "http://readcomics.website"
    ```

Resolves #101 
Resolves #98
Resolves #34
Resolves #57

PR #169, opened 2018-01-23
Add french source (Manga-LEL)

Hugome:
Add of one French source (http://manga-lel.com/)

Baguette ! :fr:

### DIFF
--- a/src/fr/mangalel/src/eu/kanade/tachiyomi/extension/fr/mangalel/Mangalel.kt
+++ b/src/fr/mangalel/src/eu/kanade/tachiyomi/extension/fr/mangalel/Mangalel.kt
@@ -1,0 +1,251 @@
+package eu.kanade.tachiyomi.source.online.french
+
+import eu.kanade.tachiyomi.network.GET
+import eu.kanade.tachiyomi.network.POST
+import eu.kanade.tachiyomi.source.model.*
+import eu.kanade.tachiyomi.source.online.ParsedHttpSource
+import okhttp3.FormBody
+import okhttp3.Headers
+import okhttp3.HttpUrl
+import okhttp3.Request
+import org.jsoup.nodes.Document
+import org.jsoup.nodes.Element
+import java.security.SecureRandom
+import java.security.cert.X509Certificate
+import java.text.ParseException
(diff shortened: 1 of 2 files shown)


M2ck:
Finally a french source, thanks! 
Possible to add Scantrad (https://scantrad.fr/)?

null-dev:
There is already an active PR for this source: #103 . *oops*

src/fr/mangalel/src/eu/kanade/tachiyomi/extension/fr/mangalel/Mangalel.kt
```diff
@@ -0,0 +1,251 @@
...
+import java.security.cert.X509Certificate
+import java.text.ParseException
+import java.text.SimpleDateFormat
+import java.util.*
+import javax.net.ssl.SSLContext
+import javax.net.ssl.X509TrustManager
+
+class Mangalel : ParsedHttpSource() {
+
+    override val id: Long = 11
```
inorichi: Remove this line. Id should be automatically asigned

src/fr/mangalel/src/eu/kanade/tachiyomi/extension/fr/mangalel/Mangalel.kt
```diff
@@ -0,0 +1,251 @@
...
+    override val lang = "fr"
+
+    override val supportsLatest = true
+
+    private val catalogHeaders = Headers.Builder().apply {
+        add("User-Agent", "Mozilla/5.0 (Windows NT 6.3; WOW64)")
+        add("Host", "www.manga-lel.com")
+    }.build()
+
+    private val trustManager = object : X509TrustManager {
```
inorichi: Do you really need this or just copied it from mangahere?

inorichi:
Oh wait, I was reviewing this without reading @null-dev's comment.

The problem with the big CMS package is discoverability. How about we use `Tachiyomi: My Manga Reader CMS (Many sources)` as app name?

paronos:
I think that this PR can be closed.

Hugome:
Oh yes my bad i forget.

PR #170, opened 2018-01-25
Update changes for mangadex

CarlosEsco:
fix genre tags
add genre filter when searching
added paging for browse


cd0rw0:
Hi 

@CarlosEsco if im not mistaken current extension (v1.2.1) hide 18+ manga. 
is it possible to fix or its a limitation from site.

TQ

CarlosEsco:
@cd0rw0 not until external sources can have properties, 1.2.1 it defaulted to hide cause thats what mangadex did.  Once external sources is merged into stable this could be adjusted so you have option like website does

CarlosEsco:
@inorichi this should be good to merge

PR #181, opened 2018-01-29
MangaPark fixes

CarlosEsco:
1. Fix low quality covers when browsing
2.  Adjust chapter names to include the title
3. Pull from all versions instead of expanded.  This prevents manga from not showing chapters cause there is no default expanded.  Also shows all chapters including duplicates compared to previously only showing what was in the expanded. I originally wrote to choose the last non expanded but upon looking into mangapark more some versions have more chapters then others and it wasnt consistent.  Easiest solution is to show them all similar to  how batoto had duplicates when more then one scanlator was working on it.

Cover change:
Old

Would update after clicking a manga

New

@null-dev as an FYI dont set the url in the mangaFromElement method, everything that shows on screen calls the mangadetailsparse and they usually have higher quaility images.  Its take a few seconds longer to show covers initially but works out better

### DIFF
--- a/src/en/mangapark/build.gradle
+++ b/src/en/mangapark/build.gradle
@@ -5,8 +5,8 @@ ext {
     appName = 'Tachiyomi: MangaPark'
     pkgNameSuffix = "en.mangapark"
     extClass = '.MangaPark'
-    extVersionCode = 2
-    extVersionSuffix = 2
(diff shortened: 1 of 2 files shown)


null-dev:
@CarlosEsco Thanks! I'll keep that in mind for sources that use different quality thumbnails.

PR #187, opened 2018-02-06
Mangadex fixes and improvements

CarlosEsco:
-fixed details issue #191
-added language support
-fixed minor paging issue
-fixed some webtoons not working
-adjusted browse by letter to be a dropdown
-added drop down for r18+
-optimized grabbing image urls to reduce server hits

@inorichi  this good now pending your review.   country code with region depends on this pull request  or else it will just display Es-419.   https://github.com/inorichi/tachiyomi/pull/1240


src/all/mangadex/src/eu/kanade/tachiyomi/extension/all/mangadex/MangadexLanguages.kt
```diff
@@ -0,0 +1,37 @@
...
+class MangaDexRussian : Mangadex("ru", "ru", 7)
+class MangaDexGerman : Mangadex("de", "de", 8)
+class MangaDexFrench : Mangadex("fr", "fr", 10)
+class MangaDexVietnamese : Mangadex("vi", "vn", 12)
+class MangaDexSpanishSpain : Mangadex("es", "es", 15)
+class MangaDexPortuguese : Mangadex("pt", "br", 16)
+class MangaDexSwedish : Mangadex("sv", "se", 18)
+class MangaDexTurkish : Mangadex("tr", "tr", 26)
+class MangaDexIndonesian : Mangadex("id", "id", 27)
+class MangaDexSpanishLTAM : Mangadex("Español Latin America", "mx", 29)
```
paronos: Maybe "Español Latin America" can be replaced with es_419

paronos:
I was thinking that maybe _R18+_ can be added to the Genres as TriState("R18+"), with STATE_IGNORE corresponding to "Show all" (default), STATE_INCLUDE to "Only R18+" and STATE_EXCLUDE to "Hide R18+".

src/all/mangadex/src/eu/kanade/tachiyomi/extension/all/mangadex/MangadexLanguages.kt
CarlosEsco: Doesn't work has to be an ISO language code or else it just prints whats in the field.  I also tried spa but then you have to 2 groups for spanish and no way to distinguish which is which

CarlosEsco:
@paronos wouldnt work out to well as this tri state would be only useable in search and wouldnt work when  just browsing, searching by letter, or browsing by latest.  External sources that allow properties will be available over the next month or two and then i will adjust r18.

src/all/mangadex/src/eu/kanade/tachiyomi/extension/all/mangadex/MangadexLanguages.kt
paronos: I just checked, the visualized name is not correctly resolved since in the ExtensionViewUtils.kt file it is calculated with:
```
val locale = Locale(lang)
locale.getDisplayName(locale).capitalize()
```
while the locale for a specific country must be instantiated using the constructor `Locale("en", "US")` or with `Locale.forLanguageTag("en-US")` (api 21+).

CarlosEsco: Constructor way would require all extensions to be modified to have the country associated correct? The other way is API 21+ and this app supports back to 16.  Any other solutions?

paronos:
My idea was that R18+ TriState would be used to change the cookie on the search request. Naturally it would influence only the search results like the other genres (the cookie works on the site both for normal searches and when browsing by letter) and not influence _latest_ and _popular_ queries.
The eventual source setting would be used in the _latest_ and _popular_ queries and to set the default selection for the R18+ TriState Filter (even with a source set to _Show all_, it may be useful to change the choice for a single search). In any case I will not insist.

src/all/mangadex/src/eu/kanade/tachiyomi/extension/all/mangadex/MangadexLanguages.kt
paronos: maybe replacing it with something simple like
```
val sp = lang.split('-')
val locale = when (sp.size) {
    2 -> Locale(sp[0], sp[1])
    3 -> Locale(sp[0], sp[1], sp[2])
    else -> Locale(lang)
}
locale.getDisplayName(locale).capitalize()
```

CarlosEsco:
I see where your coming from I'll implement the tri.  Will open a pull request on main repo for the locale too. Though I wish mangadex would just combine the too categories since they are same language 😬

src/all/mangadex/src/eu/kanade/tachiyomi/extension/all/mangadex/MangadexLanguages.kt
CarlosEsco: I'll have to look into more but for some reason es-419 isn't going into that ext..

CarlosEsco:
@paronos  https://github.com/inorichi/tachiyomi/pull/1240
Only the external sources were using that extension.  Which is why i couldnt get it work since this extension lang is all.  After some rewritting and combining into common location catalogs now display correctly.

image

src/all/mangadex/src/eu/kanade/tachiyomi/extension/all/mangadex/MangadexLanguages.kt
```diff
@@ -0,0 +1,37 @@
...
+class MangaDexVietnamese : Mangadex("vi", "vn", 12)
+class MangaDexSpanishSpain : Mangadex("es", "es", 15)
+class MangaDexPortuguese : Mangadex("pt", "br", 16)
+class MangaDexSwedish : Mangadex("sv", "se", 18)
+class MangaDexTurkish : Mangadex("tr", "tr", 26)
+class MangaDexIndonesian : Mangadex("id", "id", 27)
+class MangaDexSpanishLTAM : Mangadex("es-419", "mx", 29)
+class MangaDexCatalan : Mangadex("ca", "ct", 33)
+
+fun getAllEHentaiLanguages() = listOf(
```
Flat: `getAllMangadexLanguages()` perhaps? Is this function called somewhere? Didn't find it with a quick search.

CarlosEsco: no i can remove it.  Was a hold over when i copied from ehentai extension

paronos:
Today there is a notice on MangaDex
> In addition, due to popular demand, the R18+ toggle will become a user setting very soon, with R18+ material hidden by default.
I hope that this will not break this extension again.

Also I had not noticed that ExtensionViewUtils was not used everywhere, using es-419 has required more work than I had thought.

CarlosEsco:
yea it probably will but that is life.  Hopefully they keep it as a cookie option.

CarlosEsco:
@paronos  well that broke already lol.  Im seeing no way to show r18 now either without login

inorichi:
What should we do with people that already migrated to Mangadex? By merging this they'll have to do another migration.

AFAIK we have these options:

1) Merge this, but english is not updated anymore and users have to perform another migration.
2) Keep the english package too so that we don't annoy them, but we have to maintain two packages.
3) Override the generated id of the english source to use the same the previous one has, but users will have to uninstall the english package.

Flat:
I think option 3 is the best way to go here, and involves the least amount of effort on both sides

CarlosEsco:
Let me look into wonder if possible to just have the mangadex english class in the correct package so it generates correct

CarlosEsco:
@inorichi @Flat last commit worked.  Heres how i tested.

fresh install of tachi dev version.  Added english tachi in app.  Added manga from dex to library.  Uninstalled dex and noted the manga was not showing in library anymore.  Installed this new pull.  when to library and manga was back there.

CarlosEsco:
I switched this from a tri state to a dropdown as it was a little ambiguous.
r18 checked does that mean its only r18 or i want my results to include r18
r18 blank does that mean no r18 or both.

Just waiting on cookie change from mangadex and i will push the below and this will be good.
image

wopian:
Installed `all` before uninstalling the old `en` and missing all of the MangaDex manga from my library. Had to reinstall `all` for the manga to show up.

inorichi:
An application restart should suffice. When you uninstall `en` it unregisters its source, which has the same id as the one in the `all` package.

So, either uninstall `en` first, or restart the app.

PR #192, opened 2018-02-10
Add UnionMangas source

### LINKED ISSUE #50: http://unionmangas.net
Other good brazilian source http://unionmangas.net
Please add this too

alessandrojean:
Add of one Brazilian source, Union Mangás (http://unionmangas.cc).

The site doesn't have an filter system, so I leave the filters in blank, but they are testing a new version of the site, so I will keep looking and changing what are needed.

Unfortunately, I had to increase the client timeout because sometimes their server is very slow.

This should close #50.


CarlosEsco:
Their website seems completely broken?  Required captcha to enter, and all the covers on their main page don't load

wopian:
@CarlosEsco just took a look at the site and it all worked perfectly fine for me. No captcha, covers loaded on main page and chapters loaded 🤔

alessandrojean:
@CarlosEsco here the site it's working perfectly fine too. I am not sure, but there's a chance that their server only allow access from some countries. 🤔

CarlosEsco:
Weird if it works for others thats fine i guess. Must not like the U.S
image
image

alessandrojean:
@CarlosEsco most of the manga sites here closes the access just to here, I don't know why. 😅

PR #198, opened 2018-02-13
R18 during search option for Mangadex fixed chapter date

CarlosEsco:
@inorichi this is good to go.


CarlosEsco:
update to add #199

Issue #52, opened 2017-05-29
[English] xkcd


j2ghz:
https://xkcd.com/

CarlosEsco:
How would this work?  daily image per chapter?

PR #210, opened 2018-02-21
added xkcd

CarlosEsco:
#52
@j2ghz  couple things
1.  Do you know a better url i can use for the cover? 
2.  I can't currently thing of a way to see the alt text for each "chapter".   As extensions are only able to pass in urls to images  for a page.  

  |  |

### DIFF
--- a/src/en/xkcd/build.gradle
+++ b/src/en/xkcd/build.gradle
@@ -1,0 +1,19 @@
+apply plugin: 'com.android.application'
+apply plugin: 'kotlin-android'
+
+ext {
+    appName = 'Tachiyomi: xkcd'
+    pkgNameSuffix = "en.xkcd"
+    extClass = '.Xkcd'
+    extVersionCode = 1
+    extVersionSuffix = 0
+    libVersion = '1.2'
+}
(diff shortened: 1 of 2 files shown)


paronos:
I think you could use a placeholder image with the title and the alt text as second image of the "chapter".
E.g. with fakeimg.pl for the last strip:
https://fakeimg.pl/740x150/ffffff/?text=Self-Driving%20Issues%0AIf%20most%20people%20turn%20into%20muderers%20all%20of%20a%20sudden,%20we%27ll%20need%20to%0Apush%20out%20a%20firmware%20update%20or%20something.
%0A to add a new line after the title and to split the alt text around the 65 character mark. The height may be set according to the number of text lines.

CarlosEsco:
@paronos this the best thing ive ever seen.  I will mess around with it

CarlosEsco:
That worked though i realized xkcd has their unicode wrong on the json.  So some comics alt text wont be 100 % correct.  I have some global replaces for a few of their common ones.  
Also used the fakeimg to generate the cover

 |

PR #218, opened 2018-02-26
Add Dmzj manga source

yshui:


yshui:
Looks like there is a couple of bugs I need to fix..

Issue #190, opened 2018-02-08
[English] Jaiminisbox 


zbhoy:
https://jaiminisbox.com

Kovaelin:
I'd also like to see jaiminisbox, if possible. Thank you!

bbtora:
same. let's hope this gets some traction.

PR #251, opened 2018-03-11
Jaimini's Box

CarlosEsco:
#190


Vesnyx:
So, Jamini's Box is based on a fork of FoolSlide, which (for the most part) all have the same basic format for finding new chapters and images. Should we extend the model to permit easy addition of other sites, or should we keep it individual for each site. 

The reason I ask is because there are a very large amount of scanlator sites that use the FoolSlide reader and it may be easier to make one catch-all that can be easily added to.

CarlosEsco:
Do you know some other sites so I can mess around with it to see.

Vesnyx:
Yeah here's a few:
[Mangaichi Scanlation Division](http://mangaichiscans.mokkori.fr/fs/)
[KireiCake](https://reader.kireicake.com/)
[Champion Scans](http://reader.championscans.com/)

Issue #99, opened 2017-08-26
[English] mangarock.com


ChaotikZ:
https://mangarock.com/

arkon:
I'll work on this.

arkon:
I've gotten most of it to work (i.e. browsing, searching), but the hardest part is getting the actual images, which have to be decoded. Would it make sense to add a decoder to the main app?

Check out [this fiddle](https://jsfiddle.net/6h2sLcs4/15/) for an example.

CarlosEsco:
Keep the decoder in the extension for now.

arkon:
Makes sense, since it's tied to the source.

Another issue: even after decoding it, I end up with a data URI (e.g. `data:image/webp;base64,...`. I don't think the app supports that?

Th3-822:
The base64 encoding it's done after decrypting the image, so it's nothing to be worried.
The image is sended to the b64 encoder after regenerating it's headers (first 15 bytes), the crypt it's a simple XOR crypt with 101 as the key

wopian:
@arkon Skia should support webp

arkon:
@wopian Thanks, you're right. Both Rapid and Skia can show [a regular absolute URL](https://www.gstatic.com/webp/gallery3/1_webp_ll.webp) fine, but a data URI doesn't seem to. I'm not really sure if it's an issue with the decoder or possibly [Glide's DataUrlLoader](https://github.com/bumptech/glide/blob/17c1c3d5781e953379904a60bae08524632e5053/library/src/main/java/com/bumptech/glide/load/model/DataUrlLoader.java)... I don't see anything helpful in the logs.

Example data URI:
```
data:image/webp;base64,UklGRi4AAABXRUJQVlA4TCEAAAAvAUAAEB8wAiMwAgSSNtse/cXjxyCCmrYNWPwmHRH9jwMA
```

wopian:
You could potentially create a bitmap (converted to jpeg/png?) from the base64 data and give the image decoders that instead

arkon:
I'll give that a shot. Currently having some issues because Kotlin doesn't natively support unsigned bytes.

Progress here: https://github.com/arkon/tachiyomi/commit/69752c64459bb7f66f036b3ccac8865542301cae

arkon:
Update: I'm getting `Bitmap`s of the pages now, just need to figure out how to display that.

arkon:
Any opinions/suggestions?

[...]

PR #275, opened 2018-04-07
Add Manga Rock source (#99)

arkon:
Since the pages are webp files, it requires the user to either use the Rapid or Skia image decoders.


CarlosEsco:
Is this good to merge?

null-dev:
Usually we handle such cases by simply ignoring the filters when search is active.

PR #276, opened 2018-04-07
Update Mangachan to Tachiyoumi state

Pavkazzz:
Only update from https://github.com/inorichi/tachiyomi


Pavkazzz:
Oooops, done this already

PR #295, opened 2018-04-15
Dmzj: set UA

yshui:
Dmzj is banning okhttp UA now...


CarlosEsco:
Increase build.gradle version

PR #333, opened 2018-05-28
Update Union Mangas with their new domain

masgritz:
I have updated the baseUrl of the source to match their current domain, fixing an HTTP Error 520 that appeared when trying to access the catalogue.


CarlosEsco:
need to update the build.gradle version.  Also add an icon.

masgritz:
Sorry, still new to Github and Android development in general. It should be correct now. 
Let me know if anything needs to be changed.

PR #365, opened 2018-06-19
mangadex: fix exception when chapter number is not a valid number; remove initial hyphen in chapter name

paronos:


CarlosEsco:
looks good.

PR #393, opened 2018-07-09
[FoolSlide/WpManga] Set time to midnight for 'today' and 'yesterday' date handling

FlaminSarge:
This is so that date_upload doesn't change on every refresh.

The use of Calendar.getInstance() in these causes the date_upload value for a chapter to change every time the source manga is refreshed, but only if that chapter's upload date was 'today' or 'yesterday'.
With the introduction of https://github.com/inorichi/tachiyomi/pull/1535, this would cause any manga with 'today' or 'yesterday' chapters to display as updated even if they were not updated.

One caveat to this is that if the same chapter is uploaded twice in a single day (with the same metadata), ChapterSourceSync.shouldUpdateDbChapter() won't be able to tell that the chapter has changed. However, this is consistent with the source's actual website, where the chapter would display "Today" for both uploads such that users wouldn't be able to tell if a chapter was reuploaded.

### DIFF
--- a/src/all/foolslide/build.gradle
+++ b/src/all/foolslide/build.gradle
@@ -5,8 +5,8 @@ ext {
     appName = 'Tachiyomi: FoolSlide'
     pkgNameSuffix = "all.foolslide"
     extClass = '.FoolSlideFactory'
-    extVersionCode = 7
-    extVersionSuffix = 7
+    extVersionCode = 8
+    extVersionSuffix = 8
     libVersion = '1.2'
 }
 dependencies {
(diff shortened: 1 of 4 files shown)


FlaminSarge:
I'm sure that this issue occurs for several other sources, but I've yet to hunt down all of them; any source that doesn't provide smaller than daily granularity for chapter upload dates should be using 'midnight' instead of 'now' when using Calendar.getInstance().

Let me know if there's a better way to handle this.

CarlosEsco:
You need to update build.gradle

FlaminSarge:
Ahhhh gotcha. Extversion. Will do.

wopian:
You need to increment `extVersionCode` and `extVersionSuffix` by 1 in the `build.gradle`, otherwise clients will never see the update.

FlaminSarge:
Bumped the numbers.

PR #407, opened 2018-07-17
rawLh and rawQQ

### LINKED ISSUE #403: Source request: RawQQ (Japanese)
Source RawLH has been shutdown by Kadokawa: https://www.reddit.com/r/manga/comments/8z3zgy/rawlh_was_forced_to_shut_down_by_kadokawa/

There's another website that carries raws, http://rawqq.com
Interface is similar to RawLH (basically the same UI) so porting the RawLH plugin to make it work with RawQQ shouldn't be a problem to someone with the know-how.

### LINKED ISSUE #405: RawLH.com URL change
BUGS:
RawLH version 1.2.0
Due to the recent shutdown of RawLH by Kadokawa RawLH's new URL is now http://lhscan.net

CarlosEsco:
fixes #405 
fixes #403 

Thanks to @ramattos for getting 99.9% of the rawQQ together.


uenoshita:
Are we sure that rawqq is the same rawlh site and not just a clone by someone else? Someone from /a/ said that the quality was lower compared to rawlh site which is why I proposed http://lhscan.net/ which is the same owner of the previous site.

ramattos:
@uenoshita that's a non-issue, we have two extensions now, RawLH (which now uses lhscan.net) and RawQQ (which uses rawqq.com). Feel free to use any of the two.

CarlosEsco:
I thought it might be a clone but the latest added wasn't an exact match.

PR #408, opened 2018-07-19
Replace any whitespace characters with spaces in query before searching

mattbague:
I noticed that when copying/pasting text from websites into the catalog search, the MangaDex extension would sometimes come back with empty search results. The cause appears to be due to other whitespace characters (e.g. tab and newline) not being stripped before sending the API request. I've attached some GIFs to hopefully better show what I mean.
#### Current Behavior
Current Behavior

#### With Fix
With Fix


CarlosEsco:
Will merge this after beta changes go live

PR #459, opened 2018-08-24
Fix Webtoons (closes #458)

### LINKED ISSUE #458: Webtoon lists are getting stuck
```
BUGS:
Provide tachiyomi version as well as version of source the report is about
```

When using the Webtoon at Catalogues, i can only see a few entries (both in Latest and Browse), and the list won't update, just gets stuck, like there are no more entries. While i can search for specific webtoons in the search menu, this behaviour makes browsing for new ones pretty hard/not user friendly.

Webtoon extension: English 1.2.3
Tachiyomi: 0.7.4

arkon:
**WIP; DO NOT MERGE**

This is for Webtoons not loading more than 10 items. The site was updated to only show 10 pages on the pages that we were scraping, so this uses different pages.

For some reason, the next page selector doesn't seem to work as expected so it keeps getting the first page though. Not sure if anyone else might have any insight.

### DIFF
--- a/src/en/webtoons/src/eu/kanade/tachiyomi/extension/en/webtoons/Webtoons.kt
+++ b/src/en/webtoons/src/eu/kanade/tachiyomi/extension/en/webtoons/Webtoons.kt
@@ -22,25 +22,9 @@ class Webtoons : ParsedHttpSource() {
 
     override val supportsLatest = true
 
-    val day: String
-        get() {
-            return when (Calendar.getInstance().get(Calendar.DAY_OF_WEEK)) {
-                Calendar.SUNDAY -> "div._list_SUNDAY"
-                Calendar.MONDAY -> "div._list_MONDAY"
-                Calendar.TUESDAY -> "div._list_TUESDAY"
-                Calendar.WEDNESDAY -> "div._list_WEDNESDAY"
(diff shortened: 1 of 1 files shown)


TonyToscana:
-The functionality of "Search" has already been discussed and will stay as it is now (was decided by a poll).

-"Latest" functionaity will not be changed either, as it's doing the work correctly, maybe we can change it so it shows the whole last week worth of webtoons instead of just a day, but it should not mimic the functionality of "Explore".

-You can change "Explore" as you want. Maybe make a poll about the possibilities (this was made when "Search" was changed).

arkon:
- You mean only return 1 page's worth of results?

- I'll need to see if there's a better way to do it then, otherwise we'll just have 10 results.

- Same as above.

TonyToscana:
1. Yes, only a page, but a "special" page. The only one where webtoon and discover comics are seen together. (Pic 1)

2. Not exactly 10 results, but all the comics that have been released on the current day. (Pic 2, day on green, and the webtoons marked with the UP label).
firefox_2018-08-24_13-40-53
Pic 1
firefox_2018-08-24_13-39-44
Pic 2

arkon:
I actually forgot about this. Based on the feedback above though, it's supposedly working as intended, although I do think the intended behaviour is pretty weird.

PR #473, opened 2018-09-01
Added Japscan (FR) source

G4bleb:
- There is no research page on this source, so searching just returns all mangas from the source
- There are also no thumbnails for mangas

https://www.japscan.cc/


AmineI:
I see it was already merged, so I don't really know if I should make another issue for this, but after installation of a self-compiled apk, after trusting the extension and it disappears from the installed extensions list. (even before the PR merge)
Now, after installation from the extensions tab of Tachiyomi (last dev version), it doesn't appear as an installed extension after that either.

CarlosEsco:
@G4bleb can you fix this. It looks like package name is wrong in the kotlin class. Which is causing this issur

G4bleb:
Done !
#476 
Thank you @CarlosEsco for spotting the issue, I couldn't figure out what was wrong

AmineI:
@G4bleb I noticed any manga from Japscan show an "unknown" description. I'll fixed that in #478 : you selected the first div with the "synopsis" class, while the description is in fact in the div with a "synopsis" **id** ;)

PR #515, opened 2018-09-24
fix mangasee latest

CarlosEsco:


src/en/mangasee/src/eu/kanade/tachiyomi/extension/en/mangasee/Mangasee.kt
```diff
@@ -168,7 +168,7 @@ class Mangasee : ParsedHttpSource() {
     override fun latestUpdatesSelector(): String = "a.latestSeries"
 
     override fun latestUpdatesRequest(page: Int): Request {
-        val url = "http://mangaseeonline.net/home/latest.request.php"
+        val url = "http://mangaseeonline.us/home/latest.request.php"
```
arkon: Should use `$baseUrl` to prevent this issue in the future.

CarlosEsco:
Yeah I thought of that after the fact. :smiley:

PR #526, opened 2018-09-30
Dynasty-Scans: search results return page 1, only and infinitely

ianharmon:
Tachiyomi version: 0.7.4
Dynasty-Scans version: 1.2.7

For all 5 views of the Dynasty-Scans plugin, search results only return the first page of 20. You can continue to scroll and scroll and it will keep duplicating those results.

loop

I checked the source and the search requests were missing the $page parameter. I added it to all 5 views and tested with the following queries:

Dynasty-Anthologies: `touhou`
Dynasty-Chapters: `touhou`
Dynasty-Doujins: `doujin`
Dynasty-Issues: `vol`
Dynasty-Series: `touhou`

### DIFF
--- a/src/en/dynasty/build.gradle
+++ b/src/en/dynasty/build.gradle
@@ -5,8 +5,8 @@ ext {
     appName = 'Tachiyomi: Dynasty'
     pkgNameSuffix = 'en.dynasty'
     extClass = '.DynastyFactory'
-    extVersionCode = 7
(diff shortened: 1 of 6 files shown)


CarlosEsco:
You need to update the build Gradle file

ianharmon:
Like that?

CarlosEsco:
Looks good

PR #533, opened 2018-10-02
Update OkHttp

arkon:


CarlosEsco:
Once these get updated on main app i will merge these

inorichi:
This is actually irrelevant since okhttp isn't compiled on the extensions an their API haven't changed.

arkon:
Ah, true.

PR #567, opened 2018-10-19
Fixes TuMangaOnline, #561

### LINKED ISSUE #561: [TMO] http error 504/503
First than nothing thank you for adding TuMangaOnline extension. I tried reading a manga but I got a message saying "http error 503", went back and selected the same chapter and pages loaded normally, then in the next chapter I got a "http error 504", and I'm getting the same in all the mangas I choose to read (I must clarify thus I got only when selecting the chapters, everything else is working fine).

mll-Kerberos:
Fixes #561.

Overriding headersBuilder and adding headers to the GET request.


CarlosEsco:
Need to update the build.gradle

PR #568, opened 2018-10-19
Fixes TuMangaOnline, #561

### LINKED ISSUE #561: [TMO] http error 504/503

mll-Kerberos:
Fixes #561.

Overriding headersBuilder and adding headers to the GET request.

Reopened because I forgot to change the extension version.

### DIFF
--- a/src/es/tumangaonline/build.gradle
+++ b/src/es/tumangaonline/build.gradle
@@ -5,9 +5,9 @@ ext {
     appName = 'Tachiyomi: TuMangaOnline'
     pkgNameSuffix = 'es.tumangaonline'
     extClass = '.TuMangaOnline'
-    extVersionCode = 1
+    extVersionCode = 2
     extVersionSuffix = 1
-    libVersion = '1.2.1'
+    libVersion = '1.2'
 }
 
 apply from: "$rootDir/common.gradle"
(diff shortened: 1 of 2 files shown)


mll-Kerberos:
Changed wrong version number, fixed in last commit.

edgarMejia:
Did you test it on different devices?
Yesterday I made some changes in the headers and it worked for me but when I tried on another device the error appeared again.

mll-Kerberos:
I tried on a Nexus emulator and with my physical Huawei, both working fine

src/es/tumangaonline/build.gradle
```diff
@@ -5,7 +5,7 @@ ext {
     appName = 'Tachiyomi: TuMangaOnline'
     pkgNameSuffix = 'es.tumangaonline'
     extClass = '.TuMangaOnline'
-    extVersionCode = 1
+    extVersionCode = 2
     extVersionSuffix = 1
```
arkon: Update this too.

mll-Kerberos: Which one, the suffix? What should it be?

mll-Kerberos: Should I put extVersionCode back to 1 and Suffix up to 2?

arkon: Increment both.

mll-Kerberos: Done!

Issue #546, opened 2018-10-11
Bug mangakakalot some broken manga


thu2468:
So for mangakakalot.com some of their manga redirect to manganelo.com
This breaks some of the manga in the catelog to not work.
For example this manga:
Tensei shitara slime datta ken
http://mangakakalot.com/manga/tensei_shitara_slime_datta_ken
Will redirect to
http://manganelo.com/manga/tensei_shitara_slime_datta_ken

joseph619:
Mangakakalot and Manganelo I think are owned by the same people. So it's not a bug.

thu2468:
Yes ik they are the same people but the manga inside of the app for some manga will not load b.c it redirect to manganelo. The manga will not list the chapters. If u looks at the examole link of the manga i said in the post. Search it in the catelog it wkll not show chapters. So it is a bug...

thu2468:
Here a pictute of the catelogue mangakakalot error of those type error
![screenshot_20181015-022917_tachiyomi](https://user-images.githubusercontent.com/1341488/46933465-2ac7c200-d022-11e8-9765-b167ac896c5a.jpg)

PR #570, opened 2018-10-20
Add Manganelo

thu2468:
New extension based on mangakakalot. To fix mangakakalot redirecting some manga to manganelo


thu2468:
#546 to counter that

thu2468:
yeah i tested it it works. Idk how to make the icons

PR #575, opened 2018-10-20
Remove extVersionSuffix

arkon:
There isn't much value in having `extVersionSuffix` for the version name since just incrementing the `extVersionCode` can ensure unique names already. Instead, having both just seems to confuse developers.


CarlosEsco:
Will cause some extensions to "update" but worth it

Issue #566, opened 2018-10-19
myreadingmanga doesn't load chapters


aurynluna:
myreadingmanga is having issues again. I can't load any manga chapters, so it's impossible to read any manga there right now. I tried with several chapters and none of them load no matter how much time I wait. 

I'm using the most recent version for both the extension and the application.

missimperfect:
I'm facing the same case as the original poster. In addition since it keeps on loading so I decided to download the chapter/s itself, but whenever I tried to download it it will show 'Chapter xx - Not used' and there will always be 2 pages left unable to complete the download. Ex: Chapter 1 has 15 pages, and the downloader will only download until 13 pages and stop downloading. Different problem but I thought maybe it has something to do with it. If moderator wants I can open another issue pages for this problem. Thank you.

mll-Kerberos:
The problem with MyReadingManga has been fixed, you should be able to update the extension now, thanks for reporting the problem.

PR #578, opened 2018-10-21
Fixes MyReadingManga downloads and selector

### LINKED ISSUE #577: MyReadingManga - unable to download + some pages doesn't load
Tachiyomi ver: 0.8,0
MRM extension ver: 1.2.11

As at the previous closed issue that I've mentioned that I was unable to download the chapters from MRM since the pages doesn't load. And after I've downloaded the new extension, and have clear my cache, cookies, database I was still unable to download despite the chapters have fully loaded and the downloader was downloading all the pages it still shows 'error'.

screenshot_2018-10-21-12-14-27
screenshot_2018-10-21-12-14-37

This might happen to most of the manga.

And I found that there's 1 manga in my library where the 1st chapter doesn't load fully after the first 2 pages **(Title: Himono, Tokodoki, Shanairen'ai chapter 1)**

screenshot_2018-10-21-12-17-12

and another one where the 1st chapter is "Page list is Empty". **(Title: Kanawanu Koi no Musubikata chapter 1)**

screenshot_2018-10-21-12-17-21

Thank you and sorry for troubling you guys again.

mll-Kerberos:
Fixes #577 
The Page parser was setting 1 as index on every page, making it impossible to download, also the selector was not totally correct, since some mangas randomly have a different html structure.

### DIFF
--- a/src/all/myreadingmanga/build.gradle
+++ b/src/all/myreadingmanga/build.gradle
@@ -5,7 +5,7 @@ ext {
     appName = 'Tachiyomi: MyReadingManga'
     pkgNameSuffix = 'all.myreadingmanga'
     extClass = '.MyReadingMangaFactory'
-    extVersionCode = 12
+    extVersionCode = 13
     libVersion = '1.2'
 }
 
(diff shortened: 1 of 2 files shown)


einstein95:
I personally would have done `.entry-content .separator noscript img` as the selector, removing the need for getting the lazyloaded images

mll-Kerberos:
@einstein95 It was getting the noscript img originally, but for some reason, those images never loaded (#566), that's why I had to change it to fix that issue. But the problem now is that some pages have img inside <a>, and others don't, so that's the best I came up with, since you also have the 'Related Manga' images on the same parent.

einstein95:
@mll-Kerberos Oh wow, fun site to deal with then.

CarlosEsco:
Yeah glad someone else is looking at it 😂

PR #585, opened 2018-10-24
Fix manhuaren thumbnail url

geniusgordon:


arkon:
Would've been nice to see this use `firstOrNull`

geniusgordon:
@arkon do you mean something like this
is there a better way so I don't need to call `getString` twice
```
val key = arrayOf(
    "mangaCoverimageUrl",
    "mangaPicimageUrl",
    "shareIcon"
).firstOrNull {
    obj.getString(it) != ""
}
thumbnail_url = obj.getString(key)
```

arkon:
You can do something along the lines of this to avoid eager evaluation (just going off the top of my head, so it might be wrong):

```kotlin
thumbnail_url = listOf(
      "mangaCoverimageUrl",
      "mangaPicimageUrl",
      "shareIcon")
    .asSequence()
    .map { obj.optString(it) }
    .firstOrNull { it != "" }
```

geniusgordon:
@arkon thanks for the advice, my original code looks so redendant 😂

PR #636, opened 2018-11-24
[DNM] Add Arc-Relight source

ObserverOfTime:
This PR adds a source for [Arc-Relight](https://arc-relight.site/).

I've tested and debugged it, but there's one issue that I just can't figure out how to solve:
When `CategoryList` is the only applied filter (i.e. when `Person` is blank and `Status` is "Any"),
it doesn't get passed to `searchMangaRequest` and the list of `filters` is empty.
But, when applying it along with another filter, everything works fine.

I don't know if I've done something wrong, hopefully you can help me out.


CarlosEsco:
Filter bug is a main app issue.

ObserverOfTime:
I see, thanks. :+1:

PR #639, opened 2018-11-26
Fixes MyReadingManga selector and jpeg thumbnails

happywillow0:
Hello and apologies in advanced as this is my first time using github, contributing to this project, and teaching myself enough code within the last few hours to patch some fixes. This pull should fix:

- "page list in empty" error (again). Instead of using the attribute [data-lazyloaded='1'] which I could not find (maybe due to an update on lazyloader), I switched it out to just [data-original-width] which is part of each page (leaving the exclusions)

- I noticed a few thumbnails were not loading. They were jpegs which was supposed to be fixed in a previous patch but one of the elements was a duplicate so I fixed it from element.attr("data-src") to element.attr("src") in the get image section. 

Hope this was helpful and can be accepted.

### DIFF
--- a/src/all/myreadingmanga/build.gradle
+++ b/src/all/myreadingmanga/build.gradle
@@ -5,7 +5,7 @@ ext {
     appName = 'Tachiyomi: MyReadingManga'
     pkgNameSuffix = 'all.myreadingmanga'
     extClass = '.MyReadingMangaFactory'
-    extVersionCode = 15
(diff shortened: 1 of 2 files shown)


happywillow0:
It seems like they been updating things so this breaks old mangas such as via search. I haven't tested it out but maybe using [src*=blogspot] since all the images are hosted on blogspot?

PR #644, opened 2018-11-28
Change mangakalot domain to manganelo.com

einstein95:
Fixes manga like Nanatsu no Taizai which has a JS/noscript redirect on mangakalot to manganelo.

Bug reported on Discord.

### DIFF
--- a/src/fr/japscan/build.gradle
+++ b/src/fr/japscan/build.gradle
@@ -5,7 +5,7 @@ ext {
     appName = 'Tachiyomi: Japscan'
     pkgNameSuffix = 'fr.japscan'
     extClass = '.Japscan'
-    extVersionCode = 3
+    extVersionCode = 4
     libVersion = '1.2'
 }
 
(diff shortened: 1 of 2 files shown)


CarlosEsco:
Changed wrong thing build.gradle

einstein95:
Fixed and squished

einstein95:
Not everything but manga which doesn't redirect work fine on manganelo

arkon:
So we should actually just have a Manganelo extension and kill this one?

einstein95:
Merely changing "Mangakalot" to "Manganelo" in the name should be good enough and should mean that the extension updates and any existing manga in the library move to the new site (I'm not sure this actually happens).

arkon:
There's already a Manganelo extension though.

einstein95:
In that case, yeah, probably best to merge the two extendions

CarlosEsco:
Is everything mangakalot mangelo though?

PR #648, opened 2018-11-28
Updated MRM page list selection (11/28/18)

happywillow0:
It almost seems like the owner of MRM may be watching us... or I may just be paranoid. 😛 
I updated the selection again to work with the current site code.


CarlosEsco:
I felt the same way when I was doing changes for this one before you started

PR #687, opened 2018-12-29
[japscan] Updated code to work with the new site design

Peuuuur-Noel:
Extension updated to work with new site design.
- Fixes for new site design
- Added cloudflare support
- Search by keywords (accented characters, parse json results)
- Manga thumbnails in listview

### DIFF
--- a/src/fr/japscan/src/eu/kanade/tachiyomi/extension/fr/japscan/Japscan.kt
+++ b/src/fr/japscan/src/eu/kanade/tachiyomi/extension/fr/japscan/Japscan.kt
@@ -5,17 +5,23 @@ package eu.kanade.tachiyomi.extension.fr.japscan
  * @brief Defines class Japscan for french source Japscan
  * @date 2018-09-02
  * @version 1.0
- *
- * There is no research page on this source, so searching just returns all mangas from the source
- * There are also no thumbnails for mangas
  */
 
+import com.google.gson.JsonObject
+import com.google.gson.JsonParser
 import eu.kanade.tachiyomi.network.GET
+import eu.kanade.tachiyomi.network.POST
 import eu.kanade.tachiyomi.source.model.*
 import eu.kanade.tachiyomi.source.online.ParsedHttpSource
+import okhttp3.FormBody
 import okhttp3.Request
+import okhttp3.Response
+import okhttp3.OkHttpClient
 import org.jsoup.nodes.Document
 import org.jsoup.nodes.Element
+import java.text.ParseException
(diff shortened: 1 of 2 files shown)


src/fr/japscan/src/eu/kanade/tachiyomi/extension/fr/japscan/Japscan.kt
```diff
@@ -96,29 +135,36 @@ class Japscan : ParsedHttpSource() {
...
+    override fun chapterListSelector() = "div#chapters_list > div.collapse > div.chapters_list"
 
     override fun chapterFromElement(element: Element): SChapter {
         val urlElement = element.select("a").first()
 
         val chapter = SChapter.create()
         chapter.setUrlWithoutDomain(urlElement.attr("href"))
-        chapter.name = urlElement.text()
-        chapter.date_upload = 0
+        chapter.name = urlElement.text().replace(" VUS", "")
```
AmineI: May I ask where do they put "VUS" badges on the website ? What does that mean ? Very few chapters have these badges but I cannot figure out what that means on the website.

Peuuuur-Noel: VUS tag is for chapter posted in english translation (Version US), waiting for the localized.
You can find it on this manga : https://www.japscan.to/manga/the-promised-neverland/
Or in Thursday tab in the last release on home page.
image

Now that you mention it, I think we could skip them and wait for the french version. But they update the chapter by replacing images and changing title, not creating a new one. I wasn't sure if the app was triggered when chapter was localized, so I've let them.

AmineI: Ah, got it, thanks ! May as well add it to the exclusions, as I have a PR open already to exclude RAWs and "spoiler pictures" at #972

AmineI: While we're there, wanted to let you know about the .ownText() method, which helps in this case : it gets only the text in the element and not in its childs (in our example, "VUS" in a <span> child of our urlElement) - Implemented that in my PR, it avoids the use of a .replace(), and takes care of any other badges we may not know about (like RAW)

Peuuuur-Noel: I discovered this method in your PR. I was writing a text to ask you to update them in your PR.

PR #691, opened 2018-12-30
Add QuestionableContent Source

JamaicanMoose:


einstein95:
On doing the call to `https://www.questionablecontent.net/archive.php`, I'm getting a `lateinit property url has not been initialized` error

Issue #694, opened 2019-01-02
MerakiScans error


tblader:
MerakiScans giving HTTP error 404
They recently switched to a new site. 

https://merakiscans.com
MerakiScans 1.2.1

Tachiyomi version 0.8.0

einstein95:
@Vesnyx Your fix didn't work as the URLs that Tachiyomi tries now 404.

Vesnyx:
Yes, I know. I've almost finished completing the fix. I'm still finishing off the image parser.

PR #701, opened 2019-01-04
[Final] MerakiScans #694

Vesnyx:
Everything is fixed now, unless there's some weird dependency I'm forgetting. I can successfully get it working on my device. 

@CarlosEsco This should be good to merge. I've tested everything, but if you want to double check that would be great.


CarlosEsco:
@Vesnyx doesnt work.  

 /home/travis/build/inorichi/[secure]-extensions/src/en/merakiscans/src/eu/kanade/[secure]/extension/en/merakiscans/MerakiScans.kt: (16, 1): Class 'MerakiScans' is not abstract and does not implement abstract base class member protected abstract fun latestUpdatesRequest(page: Int): Request defined in eu.kanade.[secure].source.online.ParsedHttpSource

PR #741, opened 2019-01-25
Added NANI? Scans

Dinokin:


SynderBlack:
Worth noting that NANI already exists in the foolslide extension, although it is broken.

Dinokin:
NANI is not using foolslide anymore.

PR #766, opened 2019-02-05
Added Genres (BL)

as280093:
<li class="dropdown-option " data-type="**22**">
        <a href="#" class="option-title" data-label="BL">**BL**</a>

BL Genre was missing from genres list


CarlosEsco:
Update build.gradle

PR #886, opened 2019-03-10
mangadex: Add search by id

jonas-haeusler:
Allows to search manga on Mangadex using their respective IDs and prefixing the search with "id:".


as280093:
gradle missing

PR #890, opened 2019-03-11
Update MMRCMS

AmineI:
Update MMRCMS :
Fixed generation of js-minified websites

Removed :
- Readmanhua (Uses WPManga now)

- mangablue, henka no kaze, manga online, bacamangaku, XYXX (all offline),


src/all/mmrcms/src/eu/kanade/tachiyomi/extension/all/mmrcms/Generator.kt
```diff
@@ -135,7 +135,10 @@ class Generator {
     }
 
     private fun getItemUrl(document: Document): String {
-        return document.toString().substringAfter("showURL = \"").substringBefore("/SELECTION\";")
+        return document.toString().substringAfter("showURL = \"").substringAfter("showURL=\"").substringBefore("/SELECTION\";")
```
AmineI: I described the need of the second check in the comments : minified js, without the spaces, .substringAfter wouldn't find the string with spaces. Adding a check this isn't the most elegant way to do that, especially since mangasyuri is the only website to minify its js in the whole list.

PR #947, opened 2019-03-20
NHentai: update user agent, introduce artificial delay for fetching images (fixes #938)

### LINKED ISSUE #938: nHentai error 429
Tachiyomi version: 0.8.1
nHentai version: 1.2.6

Error is as stated in the title. Nothing loads when opening that source and that error prompt appears.

arkon:
This probably isn't the best way of handling it, but it's the most rudimentary form of "throttling" I could think of within the constraint of an extension.

I'm open to suggestions.


arkon:
Maybe coroutines would be a nicer way of writing it? Not really sure if a static amount of delay per request is useful either.

PR #949, opened 2019-03-20
Make MangaDex search work by using a shared account

az4521:
i made an account named "Tachiyomi" on the useragent "Tachiyomi". rememberme tokens are useragent-based. this should get around the restrictions on searching for now.

### DIFF
--- a/src/all/mangadex/build.gradle
+++ b/src/all/mangadex/build.gradle
@@ -5,7 +5,7 @@ ext {
     appName = 'Tachiyomi: MangaDex'
     pkgNameSuffix = 'all.mangadex'
     extClass = '.MangadexFactory'
-    extVersionCode = 53
+    extVersionCode = 54
     libVersion = '1.2'
 }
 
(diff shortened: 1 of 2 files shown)


DitFranXX:
Please update `build.gradle`.

az4521:
alright i've done it. all good?

Flat:
Hardcoding some random user cookie is not a fix. If they aren't completely incompetent in defending against the DDoS attacks they would ban this account fairly quickly after large usage. Otherwise the ones doing the attacks would just make an account and use the same bypasses we're basically doing.

az4521:
Fair enough. There isn't really another fix, you'd need a user login or captcha to do anything else. I'll close it now.

CarlosEsco:
They already are making accounts and ddosing

AmineI:
Ah, just noticed this PR, that's what I was temporarily doing on my end with my own token.
I'm actually looking into extension shared prefs, what if the user had to enter that token manually in the extension settings ? Log in on computer, gets own token, copy personal token in tachiyomi>extensions>mangadex. That's workaroundish to say the least but more secure than a hardcoded shared thing .

wopian:
@AmineI it'd be easier and far better UX to just have an account section in the extension preferences like Batoto before it shut down.

AmineI:
> @AmineI it'd be easier and far better UX to just have an account section in the extension preferences like Batoto before it shut down.

https://github.com/inorichi/tachiyomi-extensions/pull/887
If I understood correctly, that wouldn't be possible until 1.x . Which is why we would be searching for workarounds in the first place I think.

wopian:
I know :wink:

PR #954, opened 2019-03-20
set nhentai thumbnails without extra requests

az4521:
i'm 99% sure this is what's causing the extra page requests with the nhentai extension. the extension doesn't scrape thumbnails from the search results page, and so tachi requests the full manga details in order to get the thumbnail.


az4521:
i've been sniped. closing.

PR #959, opened 2019-03-21
Fix TruyenTranhLH & HocVienTruyenTranh error

namnguyendeveloper:
- Fixed TruyenTranhLH 403 error when request image store at their CDN
- Fixed HocVienTruyenTranh url


CarlosEsco:
update the build.gradle files

PR #963, opened 2019-03-23
Update MMRCMS (Add scan-vf.co, along with support for websites storing mangas in root)

### LINKED ISSUE #958: [French] scan-vf.co
(CATALOGUE) SOURCE REQUESTS:
ScanVF: https://www.scan-vf.co

Thanks

AmineI:
Add scan-vf.co to the sources.

They don't use a prefix (ex : https://www.scan-vf.co/one_piece instead of https://www.scan-vf.co/**manga/**one_piece ), 

Due to this, some changes were required in both the generator and the Source class in order to support parsing from the root of the website.

Closes #958


src/all/mmrcms/src/eu/kanade/tachiyomi/extension/all/mmrcms/MyMangaReaderCMSSource.kt
```diff
@@ -25,7 +25,7 @@ class MyMangaReaderCMSSource(override val lang: String,
                              private val categoryMappings: List<Pair<String, String>>,
                              private val tagMappings: List<Pair<String, String>>?) : HttpSource() {
     private val jsonParser = JsonParser()
-    private val itemUrlPath = Uri.parse(itemUrl).pathSegments.first()
+    private val itemUrlPath = Uri.parse(itemUrl).pathSegments.firstOrNull()
```
AmineI: When there's no path segments (ie itemUrl is the root of the website), itemUrlPath is null, whereas it previously would just error out.

src/all/mmrcms/src/eu/kanade/tachiyomi/extension/all/mmrcms/MyMangaReaderCMSSource.kt
```diff
@@ -226,7 +226,8 @@ class MyMangaReaderCMSSource(override val lang: String,
 
         // Ensure chapter actually links to a manga
         // Some websites use the chapters box to link to post announcements
-        if (!Uri.parse(url).pathSegments.firstOrNull().equals(itemUrlPath, true)) {
+        // The check is skipped if mangas are stored in the root of the website (ex '/one-piece' without a segment like '/manga/one-piece')
+        if (itemUrlPath!=null&& !Uri.parse(url).pathSegments.firstOrNull().equals(itemUrlPath, true)) {
```
AmineI: If mangas are stored in the root of the website, we can't check for a path prefix to differentiate announcements and mangas, so the check is skipped in this case.

PR #974, opened 2019-03-27
create mangadex.org intent filter to open mangadex urls with tachiyomi

dexmanga:
Typing "id:xxx" in the searchbox due to the mangadex ddos is a drag. Here's a shortcut:

Second pr for supporting "tachiyomi://search" links: inorichi/tachiyomi#1936

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,11 +1,11 @@
 buildscript {
-    ext.kotlin_version = '1.3.20'
+    ext.kotlin_version = '1.3.21'
     repositories {
         jcenter()
         google()
     }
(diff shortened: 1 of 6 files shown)


inorichi:
Another thing I forgot. Since an extension provides multiple sources, you have to select the language. We'd need another screen to select the source that should handle the link... 1.x doesn't have this issue since extension-sources are 1-1.

thu2468:
That would be nice

inorichi:
I've updated a bit your PR. This can now be merged.

PR #977, opened 2019-03-28
[mangarock] Add null check for authors, convert hardcoded JSON string…

einstein95:
…s to maps


CarlosEsco:
you gonna need to merge in master into it

PR #986, opened 2019-03-30
Mangadex: add an offline search

paronos:
Since the search function has not worked for a while this PR adds a local search db (as a string array in the `Data` class). The idea is that the old manga can be searched directly using the local db, while the most recent ones can be scrolled manually using the "Latest" option.
The manga id/titles have been downloaded from the mangadex site yesterday/today. Some metadata is also present: publication status, genres, original language, vm18, **completed English translation** (search option not available on the mangadex site).
The .apk size is ~896 KB.

Lastly the "Completed" status detection for Oneshoots is now based on the "Oneshoot" tag presence (the old method gave false "Completed" status).


CarlosEsco:
Hmm don't know if I will merge this. Cause anytime a new manga gets added the db needs updated. @inorichi thoughts

arkon:
I'm not really a fan of this either, although kudos for the creative workaround.

The most ideal workaround for now would be to somehow allow the user enter a session token (similar to #949), but I don't know if the extension preferences can accept strings like that, or how easy it is for users to obtain that token.

inorichi:
I think we're going to use login source here. IIRC you'll still be able to update your library and read chapters even if you aren't logged in. Only the browse catalogs will be affected.

arkon:
Yeah, [it only shows login as an option for LoginSources](https://github.com/inorichi/tachiyomi/blob/136e90638a8f006996c9398e8768629b08be97ff/app/src/main/java/eu/kanade/tachiyomi/ui/catalogue/SourceHolder.kt#L47-L58). I'm wondering if it'd be a good idea to update it so that it shows the other options too?

paronos:
This was obviously intended as a temporary solution until the login source implementation (since I thought it would not be ready until the Tachiyomi 1.x release). If the ETA is near then it is useless to think of a workaround.
Anyway another attempt I tried was to use a generic web search engine (https://jivesearch.com/?o=json&q=site:https://mangadex.org/title/%20dragon%20ball) to perform the search, but it is less precise than the local db.
Would it be a more acceptable workaround? Maybe the two options could also be combined (web search and local db) for better accuracy.

paronos:
On a whim I just added a db update function (from a remote json) to the extension.

If someone wants to try the extension a test apk is at https://github.com/paronos/tachiyomi-extensions/raw/mangadexupdate/tachiyomi-all.mangadex-v1.2.54-debug.apk

inorichi:
I'd close this in favour of #974

arkon:
It's also possible to log into MangaDex via the webview in Tachiyomi v0.8.4, so search works as expected.

PR #1024, opened 2019-04-14
Update Korean extensions

### LINKED ISSUE #1009: Newtoki search not working.
The search doesn't work for this source, I am using the latest Dev version. It didn't work before as well. I searched this > 버림 받은 황비 nothing showed up but on site when I search this series it shows up.

DitFranXX:
JMana: Temporary pagination fix.
MSM: Improve image decoder(which don't modifying url anymore), Use `s3` and `img_list1` as fallback, Fix description issue, and Update filter and domain
NewToki: Move to Factory, Add Webtoon Source(Not tested, fix #1009), and Update domain.

### DIFF
--- a/src/ko/jmana/src/eu/kanade/tachiyomi/extension/ko/jmana/JMana.kt
+++ b/src/ko/jmana/src/eu/kanade/tachiyomi/extension/ko/jmana/JMana.kt
@@ -49,12 +49,8 @@ class JMana : ParsedHttpSource() {
             popularMangaFromElement(element)
         }
 
-        val hasNextPage = try {
-            val page = document.select(popularMangaNextPageSelector())
-            !page[page.size - 2].getElementsByTag("a").attr("href").isNullOrEmpty()
-        } catch (_: Exception) {
-            false
-        }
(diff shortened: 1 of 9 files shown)


src/ko/jmana/src/eu/kanade/tachiyomi/extension/ko/jmana/JMana.kt
```diff
@@ -130,10 +126,20 @@ class JMana : ParsedHttpSource() {
     }
 
     override fun latestUpdatesSelector() = popularMangaSelector()
-    override fun latestUpdatesParse(response: Response) = popularMangaParse(response)
     override fun latestUpdatesFromElement(element: Element) = popularMangaFromElement(element)
     override fun latestUpdatesRequest(page: Int) = GET("$baseUrl/frame")
```
arkon: Not strictly related, but I wonder if [this page](https://mangahide.com/comic_recent_frame) would be better?

MPThLee: `comic_recent_frame` returns duplicate mangas into chapters

Issue #390, opened 2018-07-08
[English] comicextra.com


borouj-rashid-UOIT:
https://www.comicextra.com/

arkon:
Duplicate of #32

CarlosEsco:
Keeping this one cause other one references yml still lol

PR #1040, opened 2019-04-20
ComicExtra Extension Source

Yong-L:
Addresses #390


src/en/comicextra/build.gradle
```diff
@@ -0,0 +1,18 @@
...
+
+ext {
+    appName = 'Tachiyomi: ComicExtra'
+    pkgNameSuffix = 'en.comicextra'
+    extClass = '.ComicExtra'
+    extVersionCode = 2
+    libVersion = '1.2'
+}
+
+dependencies {
```
arkon: You can remove these if you're not using them.

src/en/comicextra/build.gradle
```diff
@@ -0,0 +1,18 @@
+apply plugin: 'com.android.application'
+apply plugin: 'kotlin-android'
+
+ext {
+    appName = 'Tachiyomi: ComicExtra'
+    pkgNameSuffix = 'en.comicextra'
+    extClass = '.ComicExtra'
+    extVersionCode = 2
```
arkon: Start at 1.

Yong-L: Done. Should I also change the libversion to 1.0 as well?

arkon: Nope, using the latest version of the library is preferred.

PR #1047, opened 2019-04-21
Update Gradle, Kotlin, and enable parallel builds

arkon:
Requires the latest stable version of Android Studio (3.4).

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,11 +1,11 @@
 buildscript {
-    ext.kotlin_version = '1.3.21'
+    ext.kotlin_version = '1.3.30'
     repositories {
         jcenter()
         google()
     }
     dependencies {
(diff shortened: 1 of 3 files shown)


inorichi:
I wouldn't upgrade kotlin too much as the bytecode might be incompatible with the main app. Or at least test everything is still ok, but there's not much to gain from a kotlin update here.

inorichi:
Although I'm more worried about the new D8 rather than kotlin updates.

null-dev:
Upgrading Kaitlin may be a bad idea: https://stackoverflow.com/questions/55708793/switch-to-kotlin-1-3-30-breaks-hashcode-only-on-android-api-21

arkon:
Welp, let's hold off then.

inorichi:
Parallel builds are probably a good idea, tho. Even more here considering the amount of modules.

Issue #943, opened 2019-03-19
[Source Request] [Spanish] Ninemanga


Credits125:
URL:

[Ninemanga {Es} ](http://es.ninemanga.com)

CarlosEsco:
#584

Issue #584, opened 2018-10-23
[Multiple] NineManga.com


MerradOklom:
Source: https://ninemanga.com
Language: Multilingual (ES, EN, RU, DE, IT, BR)

edgarMejia:
I'm working on this source.

PR #1049, opened 2019-04-21
Add NineManga (#584) (#943)

edgarMejia:


src/all/ninemanga/src/eu/kanade/tachiyomi/extension/all/ninemanga/NineManga.kt
```diff
@@ -0,0 +1,170 @@
...
+                it.text()
+            }
+            author = it.select("a[itemprop=author]").text()
+            artist = ""
+            description = it.select("p[itemprop=description]").text().orEmpty()
+            status = parseStatus(it.select("a.red").first().text().orEmpty())
+        }
+    }
+
+    private fun parseStatus(status: String) = when { // TODO: agregar estados para cada idioma
```
arkon: You could probably override this in the different classes to avoid having all the languages here.

src/all/ninemanga/src/eu/kanade/tachiyomi/extension/all/ninemanga/NineMangaFactory.kt
```diff
@@ -0,0 +1,35 @@
...
+        BrNineManga(),
+        EnNineManga(),
+        RuNineManga(),
+        DeNineManga(),
+        ItNineManga(),
+        FrNineManga()
+    )
+}
+
+class EsNineManga : NineManga("EsNineManga", "http://es.ninemanga.com", "es")
```
arkon: Bit of a nitpick, but can you name them with the language code as the suffix instead? (e.g. `NineMangaES`

src/all/ninemanga/src/eu/kanade/tachiyomi/extension/all/ninemanga/NineManga.kt
edgarMejia: Ok  🤔

Issue #471, opened 2018-08-30
Webtoon - Fan translation


MartassCZ:
Hello, Sorry if i post it in the wrong place, but would be possible to add function, that i could read fan translation from webtoons?

CarlosEsco:
I'm not familiar with webtoon what is fan translation?

MartassCZ:
webtoon is site where you can read webtoon with original translate, or you can read it with fan translation in your language, so if you are from Germany, and someone from fans translate it to German, you can read it in German.

PR #1060, opened 2019-04-24
Add lang support to webtoons (#471)

Rajh:
Added FR support to webtoons.

This is my first time in kotlin and in github


src/all/webtoons/src/eu/kanade/tachiyomi/extension/all/webtoons/WebtoonsDefault.kt
```diff
@@ -0,0 +1,36 @@
...
+import eu.kanade.tachiyomi.network.GET
+import eu.kanade.tachiyomi.source.model.Page
+import eu.kanade.tachiyomi.source.model.SChapter
+import eu.kanade.tachiyomi.source.model.SManga
+import org.jsoup.nodes.Document
+import org.jsoup.nodes.Element
+import java.text.SimpleDateFormat
+import java.util.*
+
+open class WebtoonsDefault(override val lang: String) : Webtoons(lang) {
```
arkon: Why the distinction between the abstract class and a default class?

Rajh: Well I kept the manga search features from the english webtoons so I made it abstract.
Since the translated version is based around json and is not working like the english parser, I decided to split into 2 different classes. The chapter and page parsing have nothing in common between the WebtoonsDefault and the Translated version.

arkon:
@Rajh Awesome, thanks! Would you be willing to add the other languages too?

PR #1061, opened 2019-04-26
Update MSM Extension.

DitFranXX:
MSM
----
* Update filter.
* Fix tags
* Fix image fetch due to site changes (v4 -> v11)
* Add BaseUrl Override for site domain changes.
  - Korean sources are change domain every 1-2 weeks due to internet censorship.
  - For this update, I change some codes on `preference-stub`

Preference-stub
----
* Re-arrange some stubs to correct class.
  - Some of `Preference` stubs are located in `ListPreference`. I just move to `Preference` class.
* Add EditTextPreference
  - For this, I add some codes about dialog in `DialogPreference`.

Thus this update cause a bug in another extension that uses preference.
ps. There's no compile error.

### DIFF
--- a/lib/preference-stub/src/main/java/android/support/v7/preference/DialogPreference.java
+++ b/lib/preference-stub/src/main/java/android/support/v7/preference/DialogPreference.java
@@ -4,4 +4,20 @@ package android.support.v7.preference;
  * Created by Carlos on 5/9/2018.
  */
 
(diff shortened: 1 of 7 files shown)


arkon:
@DitFranXX Is this ready to be merged?

DitFranXX:
@arkon Yes.

PR #1075, opened 2019-05-02
manhuaren: make manga urls consistent

yshui:
Include only the manga id in SManga.url. The remaining required parameters
are attached by overriding *Request() functions.

This makes manga urls consistent, so mangas already in the library will
be correctly displayed as such, and user won't be able to add the same
manga to the library multiple times (although mangas added to the
library before this update will still have this problem).


arkon:
Update the version in `build.gradle` too.

yshui:
Fixed!

PR #1079, opened 2019-05-05
added SimpleHTTP extension: for http file listings

unnurq:
Just a very simple http listing parser. Used with "python -m http.server", but should be usable with other simple directory http listings.
Poor man's solution to https://github.com/inorichi/tachiyomi/issues/88

Address, port, and base path can be set in preferences view of the extension

**Limitation:**
Directory has to look like this.
[BASEURL]:[BASEPORT]/[BASEPATH]/manga_title/ch1/01.png
(...)
[BASEURL]:[BASEPORT]/[BASEPATH]/manga_title/ch1/20.png

### DIFF
--- a/src/all/simplehttp/build.gradle
+++ b/src/all/simplehttp/build.gradle
@@ -1,0 +1,19 @@
+apply plugin: 'com.android.application'
+apply plugin: 'kotlin-android'
+
+ext {
+    appName = 'Tachiyomi: SimpleHTTP'
+    pkgNameSuffix = 'all.simplehttp'
+    extClass = '.SimpleHTTP'
+    extVersionCode = 1
+    libVersion = '1.2'
+}
+
+dependencies {
+    compileOnly project(':preference-stub')
+    compileOnly 'com.google.code.gson:gson:2.8.2'
+    compileOnly 'com.github.salomonbrys.kotson:kotson:2.5.0'
+    compileOnly 'com.github.inorichi.injekt:injekt-core:65b0440'
+}
+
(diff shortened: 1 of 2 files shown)


arkon:
I think this is neat, but not entirely sure if it belongs in the extensions catalog. 

@inorichi Any thoughts?

Efreak:
This sounds nice to me. Even better would be ability to read common types of formatted

arkon:
There's some similar extensions now (e.g. Komga) that are a bit more robust.

Efreak:
I won't deny that komga is more robust, however (tell me if I'm wrong), that extension requires you to actually be using a specific 3rd party software, and won't just work with any old directory of mangas. It's a different use case entirely.

The description above sounds like simply run any web server, then connect to it via tachiyomi. That way you don't have to be using any specific server software--it'll work with any server that'll output links with the correct structure; if you don't like organizing your library that way (I don't), you can just use a script to generate a directory structure of symlinks and serve that instead. (I've done this with other software, though usually for other purposes - Calibre has been an example).

My personal use consists of using a cheap NAS (WD MyCloud) as a server--it doesn't take much to modify the built-in Apache config to serve a directory structure on another port, but running another service entirely would be hard to set up and also likely interfere with the existing software (just run apt-get upgrade once to break it).

CarlosEsco:
This is a very specific situation with a very small use case though.  Komanga fits most needs

arkon:
Maybe. My biggest gripe is that it's still dependant on a pretty specific file and page structure.

If the OP decides to rebase the PR, I can reconsider merging it.

Issue #924, opened 2019-03-17
[Source Request] [Spanish] TMOHentai


Credits125:
URL:

[TMOHentai](https://tmohentai.com)

edgarMejia:
I'll work on it

PR #1095, opened 2019-05-09
Add TMOHentai #924

edgarMejia:


as280093:
Res folder here

https://github.com/as280093/TIcons/tree/master/Requested%20Extensions%20Icons/tmohentai.com/res

ic_launcher

edgarMejia:
Thanks!

arkon:
Quick question, how similar is this to the other TMO extension(s)? Would it make sense to combine them?

edgarMejia:
@arkon this in particular (TMOHentai) is only similar in the search request (a little similar), I don't think it's convenient to combine it with the others (I don't think it's a good idea to put a hentai extension with the main), but "TuMangaOnline" and "LectorManga" are identical, we should combine them.

PR #1100, opened 2019-05-09
Move rate limit interceptor logic to module

arkon:
Figured this will need to be reusable across extensions.

Might also be worth moving into a separate project and publish to Maven?

### DIFF
--- a/lib/ratelimit/build.gradle
+++ b/lib/ratelimit/build.gradle
@@ -1,0 +1,29 @@
+apply plugin: 'com.android.library'
+apply plugin: 'kotlin-android'
+
+android {
+    compileSdkVersion 27
+    buildToolsVersion '28.0.3'
+
+    defaultConfig {
+        minSdkVersion 16
+        targetSdkVersion 27
+        versionCode 1
+        versionName '1.0.0'
+    }
+
+    buildTypes {
+        release {
+            minifyEnabled false
+        }
+    }
+}
+
+repositories {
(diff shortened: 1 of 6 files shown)


inorichi:
Why a factory? A simple class would work too.

Since you want to have a common module, I'd have `RateLimiter(permits: Int, period: Int = 1, unit: TimeUnit = TimeUnit.SECONDS)` because some sites might want a limit per minute for example.

Then, instead of hardcoding `1000`, you convert `unit` to millis and multiply by `period`.

I don't think it's worth it publishing to maven.

arkon:
> Why a factory? A simple class would work too.

I wanted to avoid having to instantiate a class to get an instance of an interceptor. Or maybe it's just the Java part of me leaking out.

> [...] because some sites might want a limit per minute for example.

Good idea, I'll add that.

> I don't think it's worth it publishing to maven.

What do you think about moving it to be part of the extension library itself?

inorichi:
> I wanted to avoid having to instantiate a class to get an instance of an interceptor. Or maybe it's just the Java part of me leaking out.

It might be a factory in 1.x (under the `Dependencies` class), but because it'll be the app who provides the factory. In this case there's no difference as each extension will ship a copy.

> What do you think about moving it to be part of the extension library itself?

I think it'll be better to just maintain it here. The extensions lib was kind of a mistake and should have been part of this repository from the very beginning.

Issue #1101, opened 2019-05-11
[Bug] ComicExtra


Bombonfun:
Comicextra.com

Some comics on cextra.com have more than one page per chapter. Tachi only shows the first page of chapters, not the entire comic list. An example is Fantastic Four(1961)

Bombonfun:
Okay, so comicextra got an update 1.2.3, and the chapters got fixed. I'm closing it.

PR #1107, opened 2019-05-14
ComicExtra doesn't show chapters from multiple pages

Yong-L:
Addresses #1101


3
```diff
@@ -0,0 +1,16 @@
+apply plugin: 'com.android.application'
```
arkon: Remove this file.

Yong-L: Just removed

arkon: I mean the whole file, not just the line.

Yong-L: Remove the gradle file?

arkon: Yes, it's a bad merge conflict resolution.

Yong-L: Oh I see. Didn't see that there was created. Just removed it

PR #1114, opened 2019-05-16
add bainian-manga(zh-cn)

feilongfl:
add chinese manga source
https://m.bnmanhua.com


arkon:
@as280093 Can you make some icons?

PR #1128, opened 2019-05-23
Abstract URL handling intent logic to library

arkon:


ZaneHannanAU:
Nice.

arkon:
@ZaneHannanAU You'll need to update your PRs, but hopefully they're a bit cleaner after this change.

PR #1139, opened 2019-05-27
Update Foolslide (Hatigarm)

SaruDesu:
Hatigarm's Active domain is "http://hatigarmscans.net"


bbtora:
@SaruDesu Catalogue still giving 404

Issue #1143, opened 2019-05-28
Manga Plus by Shueisha - Thumbnail Bug.


MercerBoi:
Extension Version: 1.2.2.

Issue: None of the thumbnails are loading for any of the manga for this extension source, neither in the catalogue, nor in the library.

![Screenshot_2019-05-28-22-45-49-015_eu kanade tachiyomi](https://user-images.githubusercontent.com/28984045/58497809-602cb480-819a-11e9-8d7b-e7a92d621f0c.png)

metroid112:
I think it's related to #866 since Shueisha also hosts Spanish versions of the manga AFAIK which sometimes don't get properly loaded.

MercerBoi:
I just checked out your issue, and yeah I've experienced that as well. Perhaps a webview solution can be implemented to bind the "No Spanish Content" setting from the website, with the app...

alessandrojean:
This happens with the mangas in the library because all the images requires a key to access and a maximum time to view it. So in this case the time probably expired and the app can't display it anymore. I tried to find some way to access the images without this maximum time, but I haven't succeeded yet. When I figure out a way to do that, I can submit a update to fix it.

About the pictures not showing in the catalogue screen, I'll take a look and try to fix it.

Issue #866, opened 2019-02-28
MangaPlus duplicate search


metroid112:
Tachiyomi r1463
Manga Plus 1.2.2
Xiaomi Mi A2 Android Pie

Searching on the Manga Plus app through the catalogue will bring duplicate entries. The 2nd seems to be in Spanish from what I've checked but doesn't bring the cover (EDIT: nor title nor author nor description) and displays this error if you try to open it.

![](https://i.imgur.com/I2Avkvz.jpg)

![](https://i.imgur.com/5RECPRE.png)

![](https://i.imgur.com/lfDW4Qe.png)

xChickens:
https://mangaplus.shueisha.co.jp/manga_list/all
I assume that they can't do much about it if it shows up on the official website.

Issue #845, opened 2019-02-25
[Source Request] [Spanish] Manga Plus by SHUEISHA


MakuSensei:
Manga plus by SHUEISHA has just announced the release of the Spanish version today.

PR #1146, opened 2019-05-29
Some fixes for Manga Plus

alessandrojean:
- Potentially fix thumbnail images not showing (#1143);
- Remove duplicate content, only showing the english titles (#866);
- Fix the "Open in browser" with a workaround based on the MangaRock one;
- Show the error message from the API when it occurs.

I pretend to implement the Spanish version (#845), but I am not sure what the steps to do that. Is it okay to create a new extension in the `all` folder that includes the both and create an update to the existing one where all the methods throws an Exception telling the user to install the new Extension and migrate the existing titles in their library? If this is okay, I pretend to do that with Manga Plus; and in a similar case unite the MangasProject and MangaLivre with another english source (7Sama) that is from the same site in a new extension in a future PR. If this steps are not okay, what are the correct ones?

### DIFF
--- a/src/en/mangaplus/build.gradle
+++ b/src/en/mangaplus/build.gradle
@@ -5,7 +5,7 @@ ext {
     appName = 'Tachiyomi: Manga Plus by Shueisha'
     pkgNameSuffix = 'en.mangaplus'
     extClass = '.MangaPlus'
-    extVersionCode = 2
+    extVersionCode = 3
     libVersion = '1.2'
 }
 
--- a/src/en/mangaplus/src/eu/kanade/tachiyomi/extension/en/mangaplus/MangaPlus.kt
+++ b/src/en/mangaplus/src/eu/kanade/tachiyomi/extension/en/mangaplus/MangaPlus.kt
@@ -1,17 +1,17 @@
 package eu.kanade.tachiyomi.extension.en.mangaplus
 
+import android.util.Log
 import com.github.salomonbrys.kotson.*
 import com.google.gson.JsonObject
 import com.google.gson.JsonParser
 import com.squareup.duktape.Duktape
 import eu.kanade.tachiyomi.network.GET
-import eu.kanade.tachiyomi.network.asObservableSuccess
 import eu.kanade.tachiyomi.source.model.*
 import eu.kanade.tachiyomi.source.online.HttpSource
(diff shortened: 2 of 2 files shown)


MerradOklom:
Hi, does this pr also fix the Open in Browser? It goes to the api page instead of the selected manga info.

alessandrojean:
@MerradOklom Actually no, but I can study how to change that and submit a new PR later if I succeed.

alessandrojean:
Based on the MangaRock solution, I have made the `mangaDetailsRequest` always return the site url instead of the API one. As the manga `url` is set as `#/titles/{id}` since the first version of the extension, I believe this should not make the app stop handling the mangas in the library, because the `url` was not changed. I tested with some titles in my library before changing this and they are working fine. So, I believe the "Open in browser" feature is fixed too, @MerradOklom.

MerradOklom:
Thank you for all your time and effort.

arkon:
> I pretend to implement the Spanish version (#845), but I am not sure what the steps to do that. Is it okay to create a new extension in the all folder that includes the both and create an update to the existing one where all the methods throws an Exception telling the user to install the new Extension and migrate the existing titles in their library?

That could work. I believe Webtoons was recently updated to support other languages too, so you could look at that as well. That can be in another PR though.

MercerBoi:
@alessandrojean, hey, thanks for the work, but unfortunately the thumbnail issue doesn't seem to be resolved (for me, at least). I discovered that I have to select each of the manga individually to load their covers, and only then it appears on the catalogue, but even that has exceptions, as you can see from the screenshot given below.

Screenshot_2019-05-30-10-42-35-187_eu kanade tachiyomi

alessandrojean:
@MercerBoi This happens because the app registered in the database the old thumbnail url, even for the mangas you don't have in your library. So, to fix that, you can go to Preferences > Advanced and do the "Clear Database" and "Refresh library metadata". I have done this here and all the thumbnails came back.

Unfortunately I realized now that in the Popular Mangas I forgot to do the trick to make the thumbnails stay forever, so eventually this can happen again with the titles that are not in your library. But don't worry, I will fix that soon.

MercerBoi:
Thank you, it works! Keep up the good work, man. It's much appreciated.

PR #1154, opened 2019-05-30
Add language support to MangaPlus

alessandrojean:
- Add Spanish version (#845);
- Remove `duration` query parameter in popular mangas;
- Update Protobuf schema.

This is based on the Webtoons language support PR (#1060), but I am not sure if the app will show to the user that the extension have a update as the package name has changed. If there's something more to change, let me know.


src/all/mangaplus/src/eu/kanade/tachiyomi/extension/en/mangaplus/MangaPlusEnglish.kt
```diff
@@ -0,0 +1,5 @@
+package eu.kanade.tachiyomi.extension.en.mangaplus
```
arkon: I feel like the only benefit here is that it's the same as before, but I don't think the extension would even autoupdate due to the changed package name in the Gradle file anyway. I suspect there's no point in doing this rather than just doing it directly in the factory, unless the factory itself retained the old package name like the old version.

alessandrojean: I just followed exactly the way done at the Webtoons PR (#1060). In the case, they changed the package name and moved the files from the `en` to the `all` folder, renaming the package names.

alessandrojean:
@arkon I am thinking, since the source name and the language (in the english version) is the same, consequently the id generated will be the same as the existing extension in the new one on the english version. So, maybe if I release a update to the existing one that throws exceptions telling the user to uninstall the current one and install the new, maybe the user doesn't even need to migrate their library since the id of the source will be the same. So, is it better to do this too?

src/all/mangaplus/src/eu/kanade/tachiyomi/extension/en/mangaplus/MangaPlusEnglish.kt
arkon: Yeah, it didn't actually trigger an update since it appeared like a different extension to the system. Bit unfortunate, but not the end of the world.

arkon:
I'm not really sure. I'm fine with how it is now and needing users to have to migrate too (I don't _think_ downloads would be affected?).

alessandrojean:
So, just as a guarantee I will change the existing one too just to tell the users to install the new extension and migrate if they need to.

alessandrojean:
@arkon If everything goes as expected, the old extension now tell the user to install the new one and migrate. After a while, the deprecated extension can be removed from the repository. If there's no problems anymore, I think it can be merged.

src/all/mangaplus/src/eu/kanade/tachiyomi/extension/en/mangaplus/MangaPlusEnglish.kt
```diff
@@ -0,0 +1,5 @@
...
+
+import eu.kanade.tachiyomi.extension.all.mangaplus.MangaPlus
+
+class MangaPlusEnglish : MangaPlus("en", "eng", 0)
```
arkon: Just create these directly in the factory.

src/all/mangaplus/build.gradle
```diff
@@ -0,0 +1,18 @@
+apply plugin: 'com.android.application'
+apply plugin: 'kotlin-android'
+
+ext {
+    appName = 'Tachiyomi: Manga Plus by Shueisha'
+    pkgNameSuffix = 'all.mangaplus'
+    extClass = '.MangaPlusFactory'
+    extVersionCode = 4
```
arkon: Can start from 1.

src/en/mangaplus/build.gradle
```diff
@@ -2,17 +2,11 @@ apply plugin: 'com.android.application'
 apply plugin: 'kotlin-android'
 
 ext {
-    appName = 'Tachiyomi: Manga Plus by Shueisha'
+    appName = 'Tachiyomi: Manga Plus by Shueisha (Deprecated)'
```
arkon: I'm like 99% sure changing the name changes the ID.

You could potentially name the new one `MANGA Plus by SHUEISHA` to match their official naming in order to distinguish them.

Or maybe even change the icon?

alessandrojean: So, as far as I know, the id is related to the `name` property inside the source, and not the `appName` on the Gradle file.

alessandrojean: But okay, I will rename the new to match the official name.

arkon: Ah, that's cool. :)

Issue #1082, opened 2019-05-06
[Source Request][English] LeviatanScans


tblader:
[https://leviatanscans.com/](https://leviatanscans.com/)
It's a Spanish group but they recently picked up two very popular series in English with a delay on upload to other sites. 

Thank you!

Xeiniac:
https://mangasushi.net/ seems to use the same template too.

arkon:
#1164

PR #1164, opened 2019-06-06
Add Madara (LeviatanScans and others) support

SkiFire13:
~~This adds support for the LeviatanScans source.
It's listed under "en" but it mixes both english and spanish translations and I don't think there's a way to distinguish between them with a CSS selector~~

It now adds support for some Madara sources.
It currently supports:
- LeviatanScans
- Mangasushi
- NinjaScans
- ReadManhua
- ZeroScans


as280093:
use this res file
[ic_launcher.zip](https://github.com/inorichi/tachiyomi-extensions/files/3263117/ic_launcher.zip)

arkon:
> I don't think there's a way to distinguish between them with a CSS selector

You might be able to be base it on whether an entry contains the word "Chapter" or "Capitulo"? I haven't looked through the site very deeply though.

SkiFire13:
This is my first time using Jsoup so I thought it only supported the vanilla css selectors but looks like I was wrong. I'll try to split the english chapters from the spanish ones and refactor the extension under the "all" section.

I also found this issue #1082 where it's stated that both leviatanscans.com and mangasushi.net are using the same templace. I did a bit of reserch and looks like it's named "Madara" and it's used by other websites too. 
For example:
- https://leviatanscans.com
- https://mangasushi.net
- https://ninjascans.com
- https://readmanhua.net
- https://zeroscans.com

I'll see if I can use the same logic for all of them.

SkiFire13:
I made some progress:
- The extension is no longer en.leviatanscans but all.madara
- It now includes LeviatanScans (en and es mangas are separated), Mangasushi, NinjaScans, ReadManhua and ZeroScans
- The search now works on all these sites (even on the ones where it doesn't work in the web version lol)

There are still some minor problems:
- In LeviatanScans the mangas that don't have any chapter won't be show in neither of the versions. I could default it to english version since looks like the scanlator isn't updating any spanish manga
- The icon is still the LeviatanScans one. I searched an icon of the Madara theme but I dind't find any decent icon I could use.
- The popular and search page selector for some sources uses a workaround that doesn't always work, in particular when the last page is full. If that's the case the app will get a 404 error but at least it loads all the mangas. The sources that need to use the workaround are LeviatanScans, Mangasushi and ReadManhua. Actually only Mangasushi it gives the 404 error.

arkon:
Nice! 

@as280093 We'll need new icons.

as280093:
[zeroScans.zip](https://github.com/inorichi/tachiyomi-extensions/files/3268200/zeroScans.zip)
[ninjaScans.zip](https://github.com/inorichi/tachiyomi-extensions/files/3268201/ninjaScans.zip)
[mangaSushi.zip](https://github.com/inorichi/tachiyomi-extensions/files/3268202/mangaSushi.zip)
[readmanhua.zip](https://github.com/inorichi/tachiyomi-extensions/files/3268203/readmanhua.zip)

made via phone just check if everything is ok if not ping me

as280093:
lol didn't read much here the original one
[madara WordPress.zip](https://github.com/inorichi/tachiyomi-extensions/files/3268223/madara.WordPress.zip)

SkiFire13:
I made some changes regarding these issues:

> The popular and search page selector for some sources uses a workaround that doesn't always work, in particular when the last page is full. If that's the case the app will get a 404 error but at least it loads all the mangas. The sources that need to use the workaround are LeviatanScans, Mangasushi and ReadManhua. Actually only Mangasushi it gives the 404 error.

I fixed this problem by making a POST request for these sources. I couldn't make the POST request for NinjaScans and ZeroScans too because it would break the page selector for them... The solution was to make the Madara class abstract and 2 different implementations.

> In LeviatanScans the mangas that don't have any chapter won't be show in neither of the versions. I could default it to english version since looks like the scanlator isn't updating any spanish manga

I think the default to english is now required because the search response contains the words "Latest Chapters" even in the spanish mangas divs.

I still have to implement the filters and after that I think it's finished.

SkiFire13:
Filters look a real pain to implement since every site use a different set of ids and the order of sorting can be controlled only in some specific cases. Considering that these sites are very small (only zeroscans has more than 2 pages) it's not an essential feature and I don't think it's worth it. What do you think? 

Ps: I found out that the CONTRIBUTING.md file has a link and a list broken. Should I fix it and add the commit in this PR or should I make another PR?

arkon:
> Filters look a real pain to implement since every site use a different set of ids and the order of sorting can be controlled only in some specific cases. Considering that these sites are very small (only zeroscans has more than 2 pages) it's not an essential feature and I don't think it's worth it. What do you think?

If you feel that they're not necessary as a user, I'm fine without them. We can always add them at a later date if needed.

> Ps: I found out that the CONTRIBUTING.md file has a link and a list broken. Should I fix it and add the commit in this PR or should I make another PR?

Do that in a separate PR.

arkon:
Is this in a good state to be merged?

arkon:
Also, I feel like some of these are in the existing WPManga extension too. Would these actually fit into there?

SnakeDoc83:
WPManga consists of TrashScanlations and ZeroScans.  Trash Scans has rebranded to HappyTeaScans and their old website that WPManga calls seems to be down (new one uses Madara too).

SkiFire13:
I didn't include HappyTeaScans since I didn't know it existed. I'll try to include it with the current implementation.

somebody1234:
Leviatan 404s now, maybe they switched away from Madara?

arkon:
Yes, it's been moved to the Genkan extension.

PR #1175, opened 2019-06-09
Fix TMO error #1166

edgarMejia:


src/es/tumangaonline/src/eu/kanade/tachiyomi/extension/es/tumangaonline/TuMangaOnline.kt
```diff
@@ -200,57 +201,92 @@ class TuMangaOnline : ParsedHttpSource() {
...
     }
 
-    override fun pageListParse(document: Document): List<Page> = mutableListOf<Page>().apply {
-        document.select("div#viewer-container > div.viewer-image-container > img.viewer-image")?.forEach {
-            add(Page(size, "", it.attr("src")))
+    override fun pageListParse(response: Response): List<Page> {
+        val pages = mutableListOf<Page>()
+        val body = response.body()!!.string()
+        var images = ""
+        val script = """
```
arkon: Does this really need to be run as JS?

edgarMejia: I'm not sure, it's my first time using Duktape, for now it's all I can do.

arkon: I meant more along the lines of can't you just get this from the page source in Kotlin?

edgarMejia: Now they use canvas, I don't know how to get the images without JS, I'm not sure if there's another way to do it with Kotlin.

Like this:

```html
<div class="viewer-image-container text-center">
    <figure>
        <canvas id="vYha5qsLE5" class="viewer-image" oncontextmenu="return false;" height="1821" width="1280"></canvas>
    </figure>
</div>
```

arkon: That's actually script tag on the page that sets all of them, although I guess this is a bit more future-proof?

edgarMejia: Hope so.

Also this solves the error that doesn't allow you to download many chapters.
Now you can download without limits.

PR #1188, opened 2019-06-12
New Hatigarm Scans extension

SnakeDoc83:
Hatigarm is currently in the Foolslide extension but it's been 404ing for 2 weeks or so.  Having checked their website, it doesn't appear to be using the Folslide reader any longer.  That being the case I've got this stand alone extension for Hatigarm going.

The search function works but may need to be updated at some point.  Their website's search appears to only give results with JSON.  I spent a little time looking in to using gson to parse the results but ultimately went a different direction.  Their website has an open directory listing their manga so I'm just searching through the titles in the extension itself.  That said, open directories don't always stay open forever; but it works for now.


Vesnyx:
Well, this is super awkward. I've already pushed the working version since they switched to MMR-CMS now. I'm not sure if the search function still works, but everything else is already taken care of. :(

as280093:
[ic_launcher (1).zip](https://github.com/inorichi/tachiyomi-extensions/files/3280358/ic_launcher.1.zip)
 use this icons as the one used are square rather then rounded square

SnakeDoc83:
Ha ha, yeah, I saw your push right after I made mine.  I didn't realize MMR would cover Hatigarm's current website layout.  Oh well.

arkon:
Sorry, looks like this isn't needed. 😥

PR #1189, opened 2019-06-12
WPManga Additions

Vesnyx:
- Removed Trash Scanlations
- Added Leviatan Scanlations (ES, but has some EN translations)
- Added Manga Sushi


Vesnyx:
Removing this since WPManga appears to have been deprecated in favor of Madara.

arkon:
Eeer honestly I'm not sure if WPManga and Madara are doing the same thing. If that's true, we should probably just keep WPManga.

Vesnyx:
I think Madara fixed a few bugs that were in the original extension and added some new sources, so I'm in favor of switching over to Madara.

Not to mention the official extension name is as such: https://themeforest.net/item/madara-wordpress-theme-for-manga/20849828

PR #1201, opened 2019-06-16
Adds a preference to change user-agent header.

ZaneHannanAU:
This commit adds the option to change the user-agent header; for privacy concerns and cloudflare issues.

---

The cloudflare issues have not been resolved, by the way. After testing at current the same error is thrown whether or not it is using http.agent bare or tachiyomi+http.agent even after passing the cf challenge.


arkon:
What's the purpose of this change if it doesn't fix the CF issues?

MangaDex explicitly asked us for this user agent, so I wouldn't want to merge this

PR #1216, opened 2019-06-22
Unites the MangaLivre source with MangasProject

alessandrojean:
Since the MangaLivre is a mirror with licensed manga, and they use the same layout and API, I moved the MangaLivre source to the MangasProject extension.

As the `name` attribute of the MangaLivre is the same as before, users will only have to uninstall the MangaLivre extension (after they update and receive the message telling them to do that through the exception) and install the MangasProject extension. They won't need to migrate their library :smile:.


src/pt/mangalivre/src/eu/kanade/tachiyomi/extension/pt/mangalivre/MangaLivre.kt
```diff
@@ -30,273 +14,18 @@ class MangaLivre : HttpSource() {
...
+    override fun mangaDetailsParse(response: Response) = throw Exception(NEED_MIGRATION)
+    override fun chapterListParse(response: Response) = throw Exception(NEED_MIGRATION)
+    override fun pageListParse(response: Response) = throw Exception(NEED_MIGRATION)
+    override fun imageUrlParse(response: Response) = throw Exception(NEED_MIGRATION)
 
     companion object {
-        val jsonParser by lazy {
-            JsonParser()
-        }
+        private const val NEED_MIGRATION = "Catálogo incorporado na nova versão da extensão mangásPROJECT."
```
arkon: Make this message shorter, otherwise users will only see the first part in the snackbar.

src/pt/mangasproject/src/eu/kanade/tachiyomi/extension/pt/mangasproject/MangasProject.kt
```diff
@@ -1,242 +1,235 @@
...
-        val hasNextPage = response.request().url().queryParameter("page")!!.toInt() < 10
+        val popularMangas = result["most_read"].array
+            .map { popularMangaItemParse(it.obj) }
 
-        if (popularMangas != null)
-            return MangasPage(popularMangas, hasNextPage)
+        val page = response.request().url().queryParameter("page")!!.toInt()
 
-        return MangasPage(emptyList(), false)
+        return MangasPage(popularMangas, page < 10)
```
arkon: Why 10? Is it the same for any of these sites?

src/pt/mangasproject/src/eu/kanade/tachiyomi/extension/pt/mangasproject/MangasProject.kt
```diff
@@ -1,242 +1,235 @@
...
-        val hasNextPage = response.request().url().queryParameter("page")!!.toInt() < 5
+        val latestMangas = result["releases"].array
+            .map { latestMangaItemParse(it.obj) }
 
-        if (latestMangas != null)
-            return MangasPage(latestMangas, hasNextPage)
+        val page = response.request().url().queryParameter("page")!!.toInt()
 
-        return MangasPage(emptyList(), false)
+        return MangasPage(latestMangas, page < 5)
```
arkon: Same here, 5 seems like an arbitrary number.

src/pt/mangalivre/src/eu/kanade/tachiyomi/extension/pt/mangalivre/MangaLivre.kt
alessandrojean: I tested on the emulator and all the message appears (https://imgur.com/D59Tgxh).

src/pt/mangasproject/src/eu/kanade/tachiyomi/extension/pt/mangasproject/MangasProject.kt
alessandrojean: Because the API only return the first 10 pages of data, after that it returns `false` on the `most_read` property (https://mangalivre.com/home/most_read?page=11).

And yes, it's the same for both sites.

src/pt/mangasproject/src/eu/kanade/tachiyomi/extension/pt/mangasproject/MangasProject.kt
alessandrojean: Same case as the previous one, the API only returns the first 5 pages. After that, it returns a HTTP 500 error (https://mangalivre.com/home/releases?page=6). And it's the same for both sites too.

Issue #1220, opened 2019-06-23
[Source Request] Heaven Manga


ramattos:
[http://ww0.heavenmanga.vip/](http://ww0.heavenmanga.vip/)

zagrib:
Sure https://mangatensei.com

PR #1226, opened 2019-06-24
Heaven Manga extension

ramattos:
Heaven Manga extension ( #1220 )


as280093:
hey use this res they are same as this site
https://github.com/as280093/TIcons/tree/master/Requested%20Extensions%20Icons/holymanga.net/res

ramattos:
I didn't know your icons existed, or that the mirror site existed in the first place.

You could always replace the icons afterwards since it isn't really a functional change.

as280093:
it isn't a functional change but if i can afterwards it cause extension version change and those who are using the extension have to update just for icons and extra pull request too

ramattos:
Well I can add the icons if you're all okay with waiting for a few hours before I send a new pull request since I'm not at my desktop right now.

CarlosEsco:
Yep please add when you can

ramattos:
- Added icons from @as280093 
- Fixed behavior on chapters with no date to parse

And it's all still a single commit. @CarlosEsco you may proceed :)

PR #1235, opened 2019-06-27
Remove MangaCow, MangaFox, MangaGo, MangaPlus (EN), ManwhaHentai

### LINKED ISSUE #988: Broken extensions
This is a list of extensions that are currently broken, with no active plans to fix them:

- Mangafox (#620)
- Mangago (#377)

### LINKED ISSUE #1229: MangaHentai won't load
If I try to load into the Catalogue for the source is says, "No Results" and the Manwha I have in my Library opens and closes everytime I click on a chapter, though it doesn't give any error message, I am on the latest dev build.

arkon:
Closes #1229
Closes #988


CarlosEsco:
Would need to delete the old apks too

PR #1244, opened 2019-07-01
Mangasail extension

SnakeDoc83:
Filling request: https://github.com/inorichi/tachiyomi-extensions/issues/1211


src/en/mangasail/build.gradle
```diff
@@ -0,0 +1,16 @@
...
+ext {
+    appName = 'Tachiyomi: Mangasail'
+    pkgNameSuffix = 'en.mangasail'
+    extClass = '.Mangasail'
+    extVersionCode = 1
+    libVersion = '1.2'
+}
+
+dependencies {
+    implementation 'org.apache.commons:commons-text:1.6'
```
arkon: I'd prefer if you didn't need this. You could use kotson for parsing JSON?

arkon:
You should probably learn how to use Git rather than uploading files through GitHubt too.

CarlosEsco:
Yea the commit history needs cleaned up for sure

src/en/mangasail/build.gradle
SnakeDoc83: The particular JSON string I was using that library for would sometimes have escaped characters that I didn't think I could solve with kotson.  Figured out a way after giving it some thought.  I won't need that library anymore.

SnakeDoc83:
Going to resubmit after cleaning up.

rama982:
Actually you can force push it without closing this pull.

PR #1258, opened 2019-07-04
New Factory (Psycho Play, Leviatan Scans, One Shot Scans)

SnakeDoc83:
For sites "powered by Genkan."  Currently that's Psycho Play, Leviatan Scans, and One Shot Scans (request https://github.com/inorichi/tachiyomi-extensions/issues/1211).  There's 2 or 3 other sites I found through Google also using Genkan, but they weren't worth adding in my opinion.  

This is basically the Psycho Play extension repacked and changed to a factory; figured that would be a better way to go rather than changing the Psycho Play extension (as they could always leave Genkan behind eventually).  

Need new icons though: @as280093, I'd appreciate it.


bbtora:
1. Can you mention the other sites you found? 
2. Would you have to make changes in Madara factory too as Leviatan is not working in it anymore?
3. What happens to the Psycho Play extension?

SnakeDoc83:
These are what I found but didn't add:
https://kaguyadex.com/comics
https://komiscans.com/comics
https://hunlight-scans.info/comics

Regarding Madara, Leviatan should probably be removed from it.

Regarding the Psycho Play extension, I can only speculate that it would be removed from the repo if Genkan were to be approved.  Anyone with Psycho Play installed should still be able to use it until it breaks, I think.

as280093:
official logo of Genkan is not meant for logo creation so i just made a simple logo as their official logo is based on 3d 
logo sample

[ic_launcher.zip](https://github.com/inorichi/tachiyomi-extensions/files/3358073/ic_launcher.3.zip)

src/en/Genkan/src/eu/kanade/tachiyomi/extension/en/genkan/Genkan.kt
```diff
@@ -0,0 +1,211 @@
...
+
+    private val popularMangaUrl = "$baseUrl/comics?page=" // Search is also based off this val
+    override fun popularMangaRequest(page: Int): Request {
+            return GET("$popularMangaUrl$page")
+    }
+
+    override fun latestUpdatesSelector() = popularMangaSelector()
+
+    // Track manga in latest updates page
+    private val latestUpdatesTitles = mutableListOf<String>()
```
arkon: Using a set would give better lookup performance than a list.

src/en/Genkan/src/eu/kanade/tachiyomi/extension/en/genkan/Genkan.kt
```diff
@@ -0,0 +1,211 @@
...
+            if (page == 1) latestUpdatesTitles.clear()
+            return GET("$baseUrl/latest?page=$page")
+    }
+
+    // To prevent dupes
+    override fun latestUpdatesParse(response: Response): MangasPage {
+        val latestManga = mutableListOf<SManga>()
+        val document = response.asJsoup()
+
+        document.select(latestUpdatesSelector()).forEach { element ->
```
arkon: You could probably write this in a more functional way, something like:

```
val latestManga = document.select(latestUpdatesSelector())
  .forEach { latestUpdatesFromElement(it) }
  .filter { it.title !in latestUpdatesTitle }
  .map {
    latestUpdatesTitles.add(it.title)
    it
  }
```

src/en/Genkan/src/eu/kanade/tachiyomi/extension/en/genkan/Genkan.kt
SnakeDoc83: Ok

src/en/Genkan/src/eu/kanade/tachiyomi/extension/en/genkan/Genkan.kt
SnakeDoc83: I can't figure how to do this, I get "Unresolved reference. None of the following candidates is applicable because of receiver type mismatch" for the .filter.  I assume it's because it wants a collection but the foreach is providing a unit.  I'm afraid I'm stuck if you want me to change this.

arkon: Ah well, that's okay.

src/en/Genkan/build.gradle
```diff
@@ -0,0 +1,12 @@
+apply plugin: 'com.android.application'
+apply plugin: 'kotlin-android'
+
+ext {
+    appName = 'Tachiyomi: Genkan (multiple sources)'
+    pkgNameSuffix = 'en.genkan'
```
arkon: One last thing: make this `all`, since there's no guarantee that everything will be in English in the future.

SnakeDoc83: Moved to /src/all and changed the appropriate lines.

kendoclub:
manhwahentai is back up but under this domain https://manhwahentai.site/  , hopefully, it can be re-added to tachiyomi. Also,  not all the manhwa have been fully updated, but it seems like they are working on it.

arkon:
Not sure why you're commenting here, but there's already an open issue about that.

bbtora:
Leviathan Scans should've been removed from Madara in this commit.

Issue #1271, opened 2019-07-07
1stkissmanga.com


princess15mm:
(CATALOGUE) SOURCE REQUESTS:
 I really hope you will add https://1stkissmanga.com/ as an extension. It is a great source for reading webtoons.

Issue #1261, opened 2019-07-05
[Source Request][English] Manhwa.club


joseph619:
Requesting https://manhwa.club/ , it's one of the replacements for Manhwahentai

dragonofdragons:
add this please

Titus30:
💯💯

Issue #1251, opened 2019-07-03
[Source Request][English] Wuxiaworld.site [Madara Theme]


Xeiniac:
https://wuxiaworld.site/
well specifically it would be https://wuxiaworld.site/tag/webcomics/

It uses the same theme(?) as the sources on the madara extension

Issue #1269, opened 2019-07-06
Mangasy.com


princess15mm:
(CATALOGUE) SOURCE REQUESTS:
 Please add https://www.mangasy.com/ as an extension.

PR #1284, opened 2019-07-13
Madara: Added Multiple Sources

### LINKED ISSUE #1279: [Source Request] [English] mangazuki.online
After seeing everyone's request for new sites since manhwahentai went down, I combed through all the alternatives and found https://www.mangazuki.online to be the best. Not only has it scrapped all the up-to-date series from manhwahentai but it seems to be updating from other sources also. I don't think they have any relations to mangazuki.co so this is a new extension. It is also more updated than www.mangazuki.me when it comes to adult manhwa. Thank you.

Vesnyx:
Closes #1279 

I'm not sure if these names are appropriate for the three sites since they all share the same original name. :/ 

mangazuki.club also contains both Korean and Japanese raws so I'm not sure which of the two to classify it as.

### DIFF
--- a/src/all/madara/build.gradle
+++ b/src/all/madara/build.gradle
@@ -5,7 +5,7 @@ ext {
     appName = 'Tachiyomi: Madara'
     pkgNameSuffix = "all.madara"
     extClass = '.MadaraFactory'
-    extVersionCode = 3
+    extVersionCode = 4
     libVersion = '1.2'
 }
 
--- a/src/all/madara/src/eu/kanade/tachiyomi/extension/all/madara/Madara.kt
+++ b/src/all/madara/src/eu/kanade/tachiyomi/extension/all/madara/Madara.kt
@@ -62,7 +62,7 @@ open class Madara(
         return POST("$baseUrl/wp-admin/admin-ajax.php", headers, form.build(), CacheControl.FORCE_NETWORK)
     }
 
-    override fun popularMangaNextPageSelector(): String? = "body:not(:has(.no-posts))"
+    override fun popularMangaNextPageSelector(): String? = searchMangaNextPageSelector()
 
     // Latest Updates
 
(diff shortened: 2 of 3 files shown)


Vesnyx:
Closes #1271, #1269, and #1267

Vesnyx:
Closes #1261

arkon:
> I'm not sure if these names are appropriate

You could just name them like URLs, like `Mangazuki.me`.

> mangazuki.club also contains both Korean and Japanese raws so I'm not sure which of the two to classify it as

Maybe you could override it to do searches based on manga/manhua/manhwa or whatever to try to distinguish between them, and have multiple versions for each language?

Vesnyx:
Closed #1251, #1238, #1231 

I tried to use the search parameters to modify it, but since everything is paginated, I'd have to reuse everything from the search classes for every part of the code, leading to a lot duplicated code. Additionally, the search parameters used for the form data contain a huge number of variables, once again cluttering up the code.

Secondly, the only filtering that happens is for Korean raws. This means that searching for "manga" or "manwha" still returns the Korean raws. Searching for Korean raws will only return Korean comics. If someone else has a better way of implementing it that doesn't force a ton more code or they want to add the full form requests to each, feel free to do so.

Thoasty:
First off, thank you very much for the sources. Most of them work great but I like to report a couple issues. Not sure if I should do it here or create a new issue. All mangazuki.online series give me "lateinit property name has not been initialized" error in the chapters section. 

[https://i.imgur.com/0P4ctkk.jpg](url)

A lesser issue, very few mangazuki.me series give me "no chapter found" error.

[https://i.imgur.com/wvPE80T.jpg](url)

I have a Samsung S10+ and using Madara All 1.2.4 on Tachiyomi 0.8.4

Vesnyx:
Yeah, so I checked this out and it turns out that this cluster of sites is a massive annoyance. 

If you're on mangazuki.me, some URLs will redirect to YoManga https://mangazuki.me/manga/winner-takes-all/

Similar things occur on the other cluster of "Mangazuki" sites. Sometimes they even post the chapter as a blogpost instead, which is why you occasionally get the chapter list error. Frankly, this kinda ticks me off, and I won't be trying to fix it because of how many other domains there are (mangazuki.info, etc.). 

As for the lateinit issue, I don't actually know what's causing it. I took a look at the website's source code and the selectors look correct. It might have something to do with how they've inserted an ad alternating every chapter listing.

Thoasty:
Thanks for looking into it. Yeah I don't mind mangazuki.me since the few series can be found elsewhere. I do hope you continue to keep mangazuki.online in mind since it's my favorite out of the bunch.

pavan3999:
Screenshot_20190814-023619
some manga from mangazuki.online giving this error

PR #1285, opened 2019-07-13
MangaCards (Valhalla Scans and NANI? Scans)

SnakeDoc83:
Filling request: https://github.com/inorichi/tachiyomi-extensions/issues/1223

Just needs icons, @as280093 the original requestor linked an icon if you want to do something with that, but it's pretty small.  A reverse image search says it's a "Nordic symbol of death" if you want to look for something else to work with.


as280093:
[ic_launcher.zip](https://github.com/inorichi/tachiyomi-extensions/files/3388547/ic_launcher.zip)

here are the icons

Vesnyx:
I believe this site uses an underlying CMS that some other sites also use. I can't quite recall what it was called, but there may be some other sites in the requests that do something similar.

SnakeDoc83:
@Vesnyx 
It sort of become a game of memory working on these extensions, doesn't it?  I've definitely caught myself thinking "wait, haven't I seen this before?" more than once.

@as280093 Thanks

Vesnyx:
Found it @SnakeDoc83 
NANI? Scans uses the same thing.
https://github.com/inorichi/tachiyomi-extensions/blob/master/src/en/naniscans/src/eu/kanade/tachiyomi/extension/en/naniscans/NaniScans.kt
https://naniscans.xyz/

SnakeDoc83:
@Vesnyx 
It's impressive you found that.  Looked around a bit on Google for other sites using that CMS but haven't seen any other than these 2; can't even identify what it is.  The source uses "card" as a div.class name so I'm going with that for now--not the most unique convention but I had to call it something.

@as280093 
Since this is a factory extension now, could you provide some generic icons?

as280093:
[manga cards logo sample 1.zip](https://github.com/inorichi/tachiyomi-extensions/files/3390340/ic_launcher.2.zip)
sample


It's just what I can up in short time when I heard the name
I will upload one more sample let's see which one looks good

as280093:
[manga cards sample 2.zip](https://github.com/inorichi/tachiyomi-extensions/files/3390354/ic_launcher.3.zip)
manga cards sample 2

SnakeDoc83:
Thanks, although I like the second one a bit more aesthetically, I think the first reads better at smaller sizes; so I'll go with those.

PR #1295, opened 2019-07-14
Madara tweaks

SnakeDoc83:
Fixed a few sources that had broken chapters, think they were 1stKiss, Yokai, Tritinia, maybe others.

Filtering out chapters in Mangazuki.me and Mangazuki.online that aren't hosted on their own sites.  I think it's better to potentially have a "no chapters found" message pop up rather than clicking on chapters that get you nothing.

Fixed Mangazuki.online lateinit error.

New chapter dates are sometimes represented by a graphic instead of a text date, now parsing that image's alt text for a date.

Cleaned up some no longer used imports.

AOC is a little nutty.  Like Mangazuki they link chapters to sites other than their own (MangaDex and Imgur)--filtered those out.  Also, they have 2 different ways of managing their chapters--one's a simple chapter list and the other is by volume and chapter.  The second form was causing an empty chapter list in Tachiyomi; we're parsing both forms now.  Finally, reversed their chapter order to be in-line with usual source order.

### DIFF
--- a/src/all/madara/build.gradle
+++ b/src/all/madara/build.gradle
@@ -5,7 +5,7 @@ ext {
     appName = 'Tachiyomi: Madara'
     pkgNameSuffix = "all.madara"
     extClass = '.MadaraFactory'
-    extVersionCode = 5
+    extVersionCode = 6
     libVersion = '1.2'
 }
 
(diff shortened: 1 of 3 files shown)


Vesnyx:
You da best. Thanks for finishing and fixing everything up. :)

PR #1305, opened 2019-07-17
Webcomics: Limit to 20 chapters

SnakeDoc83:
Source only allows 20 chapters per manga, after that they're locked to their own app.  Since they're unavailable to us, might as well not create chapters past 20.  Also added manga.status to details.

Closes https://github.com/inorichi/tachiyomi-extensions/issues/1298

### DIFF
--- a/src/en/webcomics/build.gradle
+++ b/src/en/webcomics/build.gradle
@@ -5,7 +5,7 @@ ext {
     appName = 'Tachiyomi: Webcomics'
     pkgNameSuffix = 'en.webcomics'
     extClass = '.Webcomics'
-    extVersionCode = 1
+    extVersionCode = 2
     libVersion = '1.2'
 }
 
--- a/src/en/webcomics/src/eu/kanade/tachiyomi/extension/en/webcomics/Webcomics.kt
+++ b/src/en/webcomics/src/eu/kanade/tachiyomi/extension/en/webcomics/Webcomics.kt
@@ -123,9 +123,12 @@ class Webcomics : ParsedHttpSource() {
         } else {
             val chapters = mutableListOf<SChapter>()
             for (i in 1..20)
-            document.select("${chapterListSelector()}:nth-child($i)").map{
-                chapters.add(chapterFromElement(it))
-            }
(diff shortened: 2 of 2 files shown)


as280093:
hey isn't it possible to add [paid] in chapter name rather then not showing the chapters

CarlosEsco:
[Paid]  is a good idea

as280093:
if possible using emoji or certain caracter that stands out

yambadr:
I disagree. I believe it will be confusing. Some users may think they have to pay for Tachiyomi to unlock those chapters. I agree with the first opinion. If it's not available for reading, it should not be in Tachiyomi.

hXtreme:
What if clicking on a paid chapter re-directed you to their app/website...

Then there shouldn't be any confusion with people wondering if they have to pay tachiyomi, right?

CarlosEsco:
not possible to do with an extension

as280093:
best option just keep its [paid] 
and redirecting will increase extension size and has more complicated coding 
2nd opinion add [paid] on clicking rather then `page list empty` toast change to `Read on official site`

bbtora:
>I believe it will be confusing. Some users may think they have to pay for Tachiyomi to unlock those chapters.

Now that this has been brought up, I would suggest to use [locked] instead of straight up indicating [paid] as it will definitely be seen as a negative with many users.

SnakeDoc83:
I think it's better to not have a bunch of empty chapters.  That said, I've added _one_ chapter that has the title: "[Attention] Additional chapters are restricted by the source to their own app"

as280093:
tachiyomi will serve as notification for new chapters

SnakeDoc83:
By notification I mean that when a user is looking at their chapter list they'll see a chapter that has the title "[Attention] Additional chapters are restricted by the source to their own app" as opposed to what we have now which is a list of chapters that aren't readable.  So the user is being notified, or informed if you prefer, that there are additional chapters that aren't accessible by Tachiyomi because of the source's restrictions.

as280093:
what i mean wasthat whenever new chapter arives rather then opening webcomics, tachiyomi will show in library updates even if page list is empty
thus making one platform for reading notifications

PR #1314, opened 2019-07-19
Update four "all" sources

TacoTheDank:
~~I had done work on the rest of the "all" sources but fucking Android Studio decided to delete all of it when I tried to reword Comicake to ComiCake after committing the Genkan changes. So I'll have to do those all over again (ノಠ益ಠ)ノ彡┻━┻ at least it makes this merge request smaller~~

~~Anyway, the codestyle changes were made automatically with the .editorconfig using Ctrl+Alt+L~~


arkon:
I'd prefer the cleanup work separately.

IIRC, fiddling around with the Kotlin libs actually messed up the bytecode and wasn't working properly in older versions of Android for some reason. It's also preferable to just keep Gradle and other dependencies at the same versions as the main app for now -- updating it will need more thorough testing.

TacoTheDank:
@arkon I've separated the gradle changes from the changes to the root files. Hopefully this is better. Look over the commits again just in case something seems off to you.

PR #1321, opened 2019-07-21
Added guya.moe support.

funkyhippo:
This is a plugin for [guya.moe](https://guya.moe), a website for the Kaguya-sama manga. 

It's pretty basic but it works.


as280093:
[ic_launcher.zip](https://github.com/inorichi/tachiyomi-extensions/files/3414509/ic_launcher.8.zip)
Use this icons

SnakeDoc83:
Are you the guy who said on Discord they were making a Kaguya based website?  If so, nice job, looks pretty nice.

src/en/guya/src/eu/kanade/tachiyomi/extension/en/guya/GuyaFactory.kt
```diff
@@ -0,0 +1,10 @@
+package eu.kanade.tachiyomi.extension.en.guya
+
+import eu.kanade.tachiyomi.source.Source
+import eu.kanade.tachiyomi.source.SourceFactory
+
+class GuyaFactory : SourceFactory {
```
arkon: This factory isn't necessary if you don't intend on it supporting multiple sites.

funkyhippo: Good catch; I didn't even think about that fact since I started off modifying an extension with a factory and just kept it as I wrote the extension. I've removed it!

funkyhippo:
> Are you the guy who said on Discord they were making a Kaguya based website? If so, nice job, looks pretty nice.

Sort of! Not the original guy but I figured I'd help out by making the extension; the site itself actually has a pretty slick reader, built from the ground up check it out.

src/en/guya/src/eu/kanade/tachiyomi/extension/en/guya/Guya.kt
```diff
@@ -0,0 +1,303 @@
...
+    private val SCANLATOR_PREFERENCE = "SCANLATOR_PREFERENCE"
+
+    init {
+        // Update the scanlator list if we need to on initialization
+        updateScanlators()
+    }
+
+    // Request builder for the "browse" page of the manga
+    override fun popularMangaRequest(page: Int): Request {
+        updateScanlators()
```
arkon: Is it actually necessary to update it here / other places if it's already called on init?

src/en/guya/src/eu/kanade/tachiyomi/extension/en/guya/Guya.kt
```diff
@@ -0,0 +1,303 @@
...
+    override val lang = "en"
+
+    private val SCANLATORS = HashMap<String, String>()
+
+    private val preferences: SharedPreferences by lazy {
+        Injekt.get<Application>().getSharedPreferences("source_$id", 0x0000)
+    }
+    private val SCANLATOR_PREFERENCE = "SCANLATOR_PREFERENCE"
+
+    init {
```
arkon: Wondering if you could just do this in a lazy instantiation instead of the `init` block.

appu1232:
> Are you the guy who said on Discord they were making a Kaguya based website? If so, nice job, looks pretty nice.

I'm the original guy. funkyhippo here was kind enough to offer to do it yesterday while I focused more on the finalizing of the website. Exposed many of the APIs to make this easier. Thanks again funky.

src/en/guya/src/eu/kanade/tachiyomi/extension/en/guya/Guya.kt
```diff
@@ -0,0 +1,303 @@
...
+                        override fun onFailure(call: Call, e: IOException) {}
+                    }
+            )
+        }
+    }
+
+    private fun clientBuilder(): OkHttpClient = network.cloudflareClient.newBuilder()
+            .connectTimeout(10, TimeUnit.SECONDS)
+            .readTimeout(30, TimeUnit.SECONDS)
+            .addNetworkInterceptor { chain ->
```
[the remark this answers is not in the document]
funkyhippo: You're right, I had intended it for something else that was scrapped, but didn't remove it. Will update.

src/en/guya/src/eu/kanade/tachiyomi/extension/en/guya/Guya.kt
```diff
@@ -0,0 +1,303 @@
...
+                val newReq = chain
+                        .request()
+                        .newBuilder()
+                        .build()
+                chain.proceed(newReq)
+            }.build()!!
+
+    // ----------------- Things we aren't supporting -----------------
+
+    override fun searchMangaRequest(page: Int, query: String, filters: FilterList): Request {
```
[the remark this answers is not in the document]
funkyhippo: It looks like it's the case; returning an empty observable just leaves a spinner running indefinitely so I'll implement the search functionality instead.

src/en/guya/src/eu/kanade/tachiyomi/extension/en/guya/Guya.kt
funkyhippo: The problem I came across is if Tachiyomi was launched with no network connectivity; the init will fail to update the map, thus we'd be out of sync. It's sort of a hack, but I figured I'd just piggyback onto all the other calls to make sure we're never in an empty state for the scanlator mapping.

src/en/guya/src/eu/kanade/tachiyomi/extension/en/guya/Guya.kt
funkyhippo: I was considering that, but I wasn't sure if a lazy instantiation of something that relies on a network call was a good idea.

arkon: Fair. Maybe a getter with a backing property would be a bit more elegant, e.g.

```kotlin
private var _scanlators: Map<String, String>?
private val SCANLATORS: Map<String, String>
    get() {
        if (_scanlators == null) { ... }
        return _scanlators;
    }
```

funkyhippo: I was trying to make this work nicely with poor network connectivity, but it doesn't seem like it recovers from poor network conditions well enough to my liking. I opted to throw the scanlator mapping logic into an inner class, which I think makes sense considering the operations I need from it (it's a really inefficient BidiMap, essentially).

PR #1334, opened 2019-07-23
remove manhwahand (dead site, part 2)

ZaneHannanAU:
The manhwahand site (`https://manhwahand.com/`) is dead, has redirects to pornographic advertisments, and does not host any manhwa at this point in time.

This removes the APK from the repo hosting.


arkon:
Remove the icon too.

Issue #728, opened 2019-01-14
[English] Bato.to


Xeiniac:
**Source Website:** https://bato.to
**Language:** English

There are a few manga/manhwa/manhua on bato.to that I haven't been able to find anywhere else.
An example would be "Who Made Me A Princess".
Thanks in advance :D

Locked-Key:
Bato.to already closed down.

einstein95:
@Locked-Key This is Bato.to in name only. It is run by the same people as https://mangawindow.net/ but has quite a few manga missing.

Xeiniac:
Yeah, it might not be the real bato.to but it still has few manga that aren't available on mangadex and other sites.

Sakurahime08:
can't seem to find bato.to in the current app? is it already available?  thanks!

einstein95:
@Sakurahime08 No, it is not available. The issue would have been closed when an extension was made

Issue #450, opened 2018-08-10
[English] mangawindow.net


Reraru:
https://mangawindow.net

It has the manga that mangapark doesn't.

as280093:
Here the icons
[Mangawindow.zip](https://github.com/inorichi/tachiyomi-extensions/files/3615156/ic_launcher.zip)

PR #1355, opened 2019-07-29
Mangatensei Extension

### LINKED ISSUE #1078: [Source Request][English] mangatensei.com
Hi, if anyone can make a catalogue extension for https://mangatensei.com. Some paid webtoons are uploaded in the site. Also the site has tons of adverts that you cant really see anything anymore.

rama982:
Closes #1078


as280093:
Here are the icons 
https://github.com/as280093/TIcons/tree/master/Requested%20Extensions%20Icons/mangatensei.com/res

MerradOklom:
Maybe you can add MangaWindow (requested in #450) using this extension as a base, because they're basically clones, only difference is that MW library seems to be bigger. There is also Bato.to (#728), but I think that one can be closed because MT is identical at least in library quantity.

PR #1359, opened 2019-07-30
Madara: Add Manhwa Club search filters

rama982:


rama982:
I will close this for now because i see other sites have same filters method too.

rama982:
Finding ways to make common search filters

Issue #262, opened 2018-03-18
Myreadingmanga search bug


InToTheEnd:
There seems to be a bug in the myreadingmanga extension where when you search for any tag, the results shown don't pass aren't a full showing of what is available on the site

PR #1360, opened 2019-07-31
MyReadingManga Alternative Search Function

happywillow0:
Fixes / Changes:
- Changed search method to one that has pages to be able to pull all search results. 

MyReadingManga currently uses WPSolr search plugin which is a good and fast but returns limited results in Tachiyomi due to the the use of infinite scroll as apposed to pages. 

This proposed change uses the search built into WordPress which returns the results in the same style as the home page which includes pages so the full results can be pulled. 

Thank you for your time and expertise in maintaining the extensions and the app. 
happywillow0


src/all/myreadingmanga/src/eu/kanade/tachiyomi/extension/all/myreadingmanga/MyReadingManga.kt
```diff
@@ -65,30 +66,39 @@ open class MyReadingManga(override val lang: String) : ParsedHttpSource() {
     override fun latestUpdatesNextPageSelector() = popularMangaNextPageSelector()
 
     override fun searchMangaRequest(page: Int, query: String, filters: FilterList): Request {
+
+        val query2 = URLEncoder.encode(query, "UTF-8")
         val uri = Uri.parse("$baseUrl/search/").buildUpon()
-        uri.appendQueryParameter("search", query)
+                uri.appendEncodedPath(query2)
+                uri.appendPath("page")
+                uri.appendPath("$page")
```
arkon: Unnecessary string interpolation.

src/all/myreadingmanga/src/eu/kanade/tachiyomi/extension/all/myreadingmanga/MyReadingManga.kt
```diff
@@ -65,30 +66,39 @@ open class MyReadingManga(override val lang: String) : ParsedHttpSource() {
     override fun latestUpdatesNextPageSelector() = popularMangaNextPageSelector()
 
     override fun searchMangaRequest(page: Int, query: String, filters: FilterList): Request {
+
+        val query2 = URLEncoder.encode(query, "UTF-8")
         val uri = Uri.parse("$baseUrl/search/").buildUpon()
-        uri.appendQueryParameter("search", query)
+                uri.appendEncodedPath(query2)
```
arkon: Is it necessary to use `appendEncodedPath` if it's already encoded above (or the other way around -- do we need to encode it first if we're doing it here)?

src/all/myreadingmanga/src/eu/kanade/tachiyomi/extension/all/myreadingmanga/MyReadingManga.kt
```diff
@@ -65,30 +66,39 @@ open class MyReadingManga(override val lang: String) : ParsedHttpSource() {
...
+            select.text().contains("[$lang", true)
+        }
+        for (element in list) {
+            mangas.add(popularMangaFromElement(element))
+
         }
 
-        return MangasPage(mangas, false)
-    }
+        val hasNextPage = popularMangaNextPageSelector().let { selector ->
```
arkon: Is using `let` here really necessary?

src/all/myreadingmanga/src/eu/kanade/tachiyomi/extension/all/myreadingmanga/MyReadingManga.kt
happywillow0: Searching directly on the website results in `https://myreadingmanga.info/search/?search=search+term`
appendQueryParamter encodes the query as `search%20term` which results in the url
`https://myreadingmanga.info/search/?search=search term` (as Firefox displays it)

The wpsolr search plugin is able to handle a search term with spaces (but returns results with infinite scroll). 
However, The built in search as part of wordpress / theme pulls directly from the url submitted and will directly search for `search%20term` resulting in `Search Results for: search%20term` 

The query is first formatted by URLencoder so a space (" ") is formatted as "+" rather than "%20".
Then appendEncodedPath adds it directly (without extra encoding) to the uri in order to not encode the "+" into something else (Appends the given segment to the path.) 

appendPath on the other hand encodes the given segment and appends it to the path which is why I tried the former.

src/all/myreadingmanga/src/eu/kanade/tachiyomi/extension/all/myreadingmanga/MyReadingManga.kt
happywillow0: I'm sorry as my coding is the best. I referenced some examples on [stackoverflow](https://stackoverflow.com/questions/19167954/use-uri-builder-in-android-or-create-url-with-variables). 

Did you mean that I didn't need the extra "uri" in front so it can be like this: `val uri = Uri.parse("$baseUrl/search/").buildUpon().appendEncodedPath(query2).appendPath("page").appendPath("$page")`

or like this:
`uri.appendPath("page/$page")`

src/all/myreadingmanga/src/eu/kanade/tachiyomi/extension/all/myreadingmanga/MyReadingManga.kt
happywillow0: This part I'm not really sure. To be honest, I don't know what `let` does. 

I just copied the block of code from  `override fun popularMangaParse(response: Response):` in order for the `override fun searchMangaParse` to work the same. I may be able to get rid of the duplicate code by referencing the former directly. Let me take a look.

happywillow0: I just got rid of the duplicate block of code to make it easier. Since it is the same as the homepage, I referenced back to the `popularMangaParse`. 

I did read up on `.let` a little bit but not enough to tell me why the original author put it in.

src/all/myreadingmanga/src/eu/kanade/tachiyomi/extension/all/myreadingmanga/MyReadingManga.kt
happywillow0: I found an answer on [stackoverflow](https://stackoverflow.com/questions/37442198/how-does-string-interpolation-work-in-kotlin) again that may be what you were referencing to. 

I got rid of the extra references so it should preform all the the steps at once.

happywillow0:
I should mention this fixes the outstanding issue #262. 

Let me know if there are any other changes requested.

PR #1361, opened 2019-07-31
Madara: Rework search filters and some improvements

rama982:
* Workaround for chapters not found on Mangazuki's linked titles
* Add genre on manga description
* Disable search for wuxia

Add Filters
* Search by Artist / Author
* Search by Year
* Sort by Status
* Order by

### DIFF
--- a/src/all/madara/build.gradle
+++ b/src/all/madara/build.gradle
@@ -5,7 +5,7 @@ ext {
     appName = 'Tachiyomi: Madara'
     pkgNameSuffix = "all.madara"
     extClass = '.MadaraFactory'
-    extVersionCode = 7
+    extVersionCode = 8
     libVersion = '1.2'
 }
 
--- a/src/all/madara/src/eu/kanade/tachiyomi/extension/all/madara/Madara.kt
+++ b/src/all/madara/src/eu/kanade/tachiyomi/extension/all/madara/Madara.kt
@@ -1,18 +1,18 @@
 package eu.kanade.tachiyomi.extension.all.madara
 
+import eu.kanade.tachiyomi.network.GET
+import eu.kanade.tachiyomi.network.POST
+import eu.kanade.tachiyomi.network.asObservable
 import eu.kanade.tachiyomi.source.model.*
 import eu.kanade.tachiyomi.source.online.ParsedHttpSource
 import okhttp3.*
-import org.jsoup.nodes.Document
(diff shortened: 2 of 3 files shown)


rama982:
Some sources still need fix.

SnakeDoc83:
I only had a brief chance to check what you're changing, but I have to ask, why?  

You're removing the code to do client-side search for Tritinia, but when I check their website it still looks broken to me.  Am I overlooking something, are you getting results when you test it?  

You're also removing code that filters Mangazuki.me and Mangazuki.online.  The reason that's there is because those two sites link to other websites instead of hosting manga themselves (in some instances).  So when you were seeing manga with no chapters, that's because they didn't have it on their website, they were linking to some other domain.  Without those filters, yes you'd get chapters, but clicking on them would get you an error (can't remember if it was 404 or something else) and no pages to read.  Did you test to see if you were getting actual readable chapters where you weren't before?

Why switch from POST to GET?  I think the original author used POST to avoid the last page issue you're handling by doing response checks and issuing " error("Already on the Last Page!")."  The POST method avoids all that, no?

rama982:
@SnakeDoc83 

- Trinitian search is working (tested)
- Mangazuki link to other sites which using madara too so content is readable (tested)
also the source have many of linked manga titles make user think extension broken. 
i'm also fixed that 404 error on it.
- For last yes it is, just thinking to revert back to POST for last updates / popular because i was wrong assuming before.

rama982:
* I'm also disable search for wuxia because even if using filters, search result still showing result for non comics titles

SnakeDoc83:
Well, if they work, fair enough, that's better if you can deliver readable chapters.  It's not always the case though.  On AOC look at "Just Be Friends" as an example.  On Mangazuki.online look at "Brawling GO!"  AOC changes fail because unlike the 2 mangazukis, AOC links to sources not using Madara (links are sometimes to MangaDex and Imgur).  "Brawling GO!" is a problem because its host site is flagged as a security risk.  You could probably add more code to check if either mangazuki is linking to bad sites, but AOC should be reverted.

rama982:
ok i will check another workaround for mangazuki,
for AOC i agreed and already reverted on force push

rama982:
my main concern on this pull is adding search filters.
mangazuki's linked host which has security issue is caused from ssl certificate error, and already guarded.

PR #1362, opened 2019-07-31
Gunnerkrigg Court

InfinityPhase:
Adds [Gunnerkrigg Court](https://www.gunnerkrigg.com/) as a source. Each page is it's own chapter, but perhaps it would be better to group them into the chapters provided? Each page is labeled with the chapter, followed by the page number. I did not find (or look very hard for) a small icon for the extension.


as280093:
Use this icon
[ic_launcher.zip](https://github.com/inorichi/tachiyomi-extensions/files/3453191/ic_launcher.11.zip)

PR #1367, opened 2019-08-02
Manhuadui: fix image server changed problem & support category search

neverfelly:


src/zh/manhuadui/src/eu/kanade/tachiyomi/extension/zh/manhuadui/Manhuadui.kt
```diff
@@ -3,6 +3,7 @@ package eu.kanade.tachiyomi.extension.zh.manhuadui
 //import android.util.Base64.NO_WRAP
 //import java.util.*
 import android.util.Base64
+import android.util.Log
```
arkon: Remove this.

neverfelly: Done.

PR #1373, opened 2019-08-05
MyReadingManga Manga Details parse update

happywillow0:
Fixed:
-Author (or should it be the artist? or both?)
-Description
-Status

Noticed that the description wasn't pulling when available. While I was at it, I added fixed author and added additional status function.

I do apologize as since this is a different pull request / fixes a different issue from my other request, how I should increment the build numbers? Please let me know if it needs to be changed. (Maintainers can edit) 

Thank you,
happywillow0

### DIFF
--- a/src/all/myreadingmanga/src/eu/kanade/tachiyomi/extension/all/myreadingmanga/MyReadingManga.kt
+++ b/src/all/myreadingmanga/src/eu/kanade/tachiyomi/extension/all/myreadingmanga/MyReadingManga.kt
@@ -115,18 +115,23 @@ open class MyReadingManga(override val lang: String) : ParsedHttpSource() {
 
     //cleans up the name removing author and language from the title
     private fun cleanTitle(title: String) = title.substringBeforeLast("[").substringAfterLast("]").substringBeforeLast("(")
+    private fun cleanAuthor(title: String) = title.substringAfter("[").substringBefore("]")
 
 
     override fun mangaDetailsParse(document: Document): SManga {
         val manga = SManga.create()
-        manga.author = document.select(".entry-content p")?.find { it ->
-            it.text().contains("artist", true) || it.text().contains("author", true)
-        }?.text()?.substringAfter(":")
-
+        manga.author = cleanAuthor(document.select("h1").text())
         val glist = document.select(".entry-header p a[href*=genre]").map { it -> it.text() }
         manga.genre = glist.joinToString(", ")
-        manga.description = document.select("h1").text() + "\n" + (document.select(".entry-content blockquote")?.first()?.text() ?: "")
-        manga.status = SManga.UNKNOWN
+        manga.description = document.select("h1").text() + "\n" + document.select(".info-class").text()
+        val mstatus = document.select("a[href*=status]").first().text()
+        when (mstatus) {
+            "Ongoing" -> manga.status = SManga.ONGOING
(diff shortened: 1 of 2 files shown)


happywillow0:
Pause this request, there is one more bug to fix.

happywillow0:
Fixed a null bug when status didn't appear on the page. Tested and confirmed things should be working as intended.

src/all/myreadingmanga/src/eu/kanade/tachiyomi/extension/all/myreadingmanga/MyReadingManga.kt
```diff
@@ -115,18 +115,24 @@ open class MyReadingManga(override val lang: String) : ParsedHttpSource() {
...
         val glist = document.select(".entry-header p a[href*=genre]").map { it -> it.text() }
         manga.genre = glist.joinToString(", ")
-        manga.description = document.select("h1").text() + "\n" + (document.select(".entry-content blockquote")?.first()?.text() ?: "")
-        manga.status = SManga.UNKNOWN
+        manga.description = document.select("h1").text() + "\n" + document.select(".info-class")?.text()
+        val mstatus = document.select("a[href*=status]")?.first()?.text()
+        when (mstatus) {
+            "Ongoing" -> manga.status = SManga.ONGOING
+            "Completed" -> manga.status = SManga.COMPLETED
+            else -> { // Note the block
```
arkon: You can just pull out the assignment so you don't need the block.

Something like:
```
manga.status = when (...) {
 ...
}
```

happywillow0: done, thank you.

src/all/myreadingmanga/src/eu/kanade/tachiyomi/extension/all/myreadingmanga/MyReadingManga.kt
```diff
@@ -115,18 +115,24 @@ open class MyReadingManga(override val lang: String) : ParsedHttpSource() {
...
-            it.text().contains("artist", true) || it.text().contains("author", true)
-        }?.text()?.substringAfter(":")
-
+        manga.author = cleanAuthor(document.select("h1").text())
+        manga.artist = cleanAuthor(document.select("h1").text())
         val glist = document.select(".entry-header p a[href*=genre]").map { it -> it.text() }
         manga.genre = glist.joinToString(", ")
-        manga.description = document.select("h1").text() + "\n" + (document.select(".entry-content blockquote")?.first()?.text() ?: "")
-        manga.status = SManga.UNKNOWN
+        manga.description = document.select("h1").text() + "\n" + document.select(".info-class")?.text()
```
happywillow0: The description update is coded based on the current uploading design of the website. Would it be better if it included both to allow for backwards compatibility with older, previously bookmarked manga? 
`manga.description = document.select("h1").text() + "\n" + document.select(".entry-content blockquote")?.first()?.text() + "\n" + document.select(".info-class")?.text()`

Also while we are at it, what does `?: ""` do?

arkon: I don't we have to care about backwards compatibility of the description.

`?:` is [the Elvis operator](https://kotlinlang.org/docs/reference/null-safety.html#elvis-operator).

Issue #1380, opened 2019-08-09
[SOURCE REQUEST] Zandy no Fansub


Orochi101:
Link: http://zandynofansub.aishiteru.org/reader/

happywillow0:
Your lucky Zandy no Fansub uses FoolSlide for their reader. I just tested it out and it seems to be working. Let me update my branch and submit a pull request to add it to the foolslide list.

Orochi101:
Thank you!

PR #1381, opened 2019-08-09
Source Request Zandy no Fansub

happywillow0:
Requested by #1380 
Used foolslide so I added it to the list. 
Tested it and it looks like it is working.


happywillow0:
Conflicting files, let me see if I can fix my master first.

PR #1382, opened 2019-08-09
 [Source Update] FoolSlide: Zandy no Fansub

happywillow0:
Requested by #1380
Used foolslide so I added it to the list.
Tested it and it looks like it is working.


Orochi101:
Thank you!

PR #1383, opened 2019-08-10
Add extension for LANraragi

Seflyx:
LANraragi project page: https://github.com/Difegue/LANraragi

Happen to host my own local manga and want to read it directly from tachiyomi.
I'm not sure which is the best for non-specific language extension, "all", "other" or instantiate every single language available in the SourceFactory so I'll be using "all" first

TO DO:
- [x] Update API


src/all/lanraragi/src/eu/kanade/tachiyomi/extension/all/lanraragi/LANraragiFactory.kt
```diff
@@ -0,0 +1,12 @@
+package eu.kanade.tachiyomi.extension.all.lanraragi
+
+import eu.kanade.tachiyomi.source.Source
+import eu.kanade.tachiyomi.source.SourceFactory
+
+class LANraragiFactory : SourceFactory {
```
arkon: This factory isn't really necessary.

src/all/lanraragi/src/eu/kanade/tachiyomi/extension/all/lanraragi/Archive.kt
```diff
@@ -0,0 +1,8 @@
+package eu.kanade.tachiyomi.extension.all.lanraragi
+
+class Archive {
```
arkon: A data class might be slightly nicer, like:

```
data class Archive(
    val arcid: String = "",
    val isnew: String = "",
    val tags: String = "",
    val title: String = ""
)
```

src/all/lanraragi/src/eu/kanade/tachiyomi/extension/all/lanraragi/LANraragi.kt
```diff
@@ -0,0 +1,220 @@
...
+
+    override val name = "LANraragi"
+
+    override val baseUrl: String
+        get() = preferences.getString("hostname", "http://127.0.0.1:3000")
+
+    override val supportsLatest = true
+
+    private val apiKey: String
+        get() = preferences.getString("apikey", "")
```
arkon: You could pull out `"apikey"` as a constant.

src/all/lanraragi/src/eu/kanade/tachiyomi/extension/all/lanraragi/LANraragi.kt
```diff
@@ -0,0 +1,220 @@
...
+            uriBuilder.appendQueryParameter("key", apiKey)
+        }
+
+        val chapters = ArrayList<SChapter>()
+        chapters.add(SChapter.create().apply {
+            val uri = uriBuilder.build()
+
+            url = "${uri.encodedPath}?${uri.encodedQuery}"
+            chapter_number = 1F
+            name = "Chapter 1"
```
arkon: It only supports a single chapter?

src/all/lanraragi/src/eu/kanade/tachiyomi/extension/all/lanraragi/LANraragi.kt
```diff
@@ -0,0 +1,220 @@
...
+    override fun chapterListParse(response: Response): List<SChapter> {
+        val id = response.request().url().queryParameter("id").toString()
+
+        val uriBuilder = Uri.parse("$baseUrl/api/extract").buildUpon()
+        uriBuilder.appendQueryParameter("id", id)
+        if(apiKey.isNotEmpty()) {
+            uriBuilder.appendQueryParameter("key", apiKey)
+        }
+
+        val chapters = ArrayList<SChapter>()
```
arkon: Can just do `return listOf(SChapter[...]))`

src/all/lanraragi/src/eu/kanade/tachiyomi/extension/all/lanraragi/LANraragi.kt
```diff
@@ -0,0 +1,220 @@
...
+    }
+
+    override fun imageUrlParse(response: Response) = throw UnsupportedOperationException("imageUrlParse is unused: ${response.request().url().encodedPath()}")
+
+    override fun latestUpdatesParse(response: Response) = genericMangaParse(response)
+
+    override fun latestUpdatesRequest(page: Int) = searchMangaRequest(page, "", FilterList())
+
+    override fun mangaDetailsParse(response: Response): SManga {
+        return SManga.create()
```
arkon: I'm guessing LANraragi doesn't contain manga details or something?

src/all/lanraragi/src/eu/kanade/tachiyomi/extension/all/lanraragi/LANraragi.kt
```diff
@@ -0,0 +1,220 @@
...
+    }
+
+    override fun pageListParse(response: Response): List<Page> {
+        val pageList = Gson().fromJson<PageList>(response.body()!!.string())
+
+        val pages = ArrayList<Page>()
+        var i = 0
+        pageList.pages.forEach { url ->
```
arkon: Could use a `map` instead of `forEach` here.  I believe there's also iterators that give you an index.

src/all/lanraragi/src/eu/kanade/tachiyomi/extension/all/lanraragi/LANraragi.kt
```diff
@@ -0,0 +1,220 @@
...
+            val uri = Uri.parse("$baseUrl/${url.trimStart('.')}")
+
+            pages.add(Page(i++, uri.toString(), uri.toString(), uri))
+        }
+        return pages
+    }
+
+    override fun popularMangaParse(response: Response) = genericMangaParse(response)
+
+    override fun popularMangaRequest(page: Int)= searchMangaRequest(page, "", FilterList())
```
arkon: So popular and latest just do the same thing?

src/all/lanraragi/src/eu/kanade/tachiyomi/extension/all/lanraragi/LANraragi.kt
```diff
@@ -0,0 +1,220 @@
...
+                summary = hostname
+                preferences.edit().putString("hostname", hostname).commit()
+            }
+        }
+
+        val apikeyPref = EditTextPreference(screen.context).apply {
+            key = "API Key"
+            title = "API Key"
+            text = apiKey
+            summary = apiKey.replace(Regex("."), "*")
```
arkon: What's this for?

src/all/lanraragi/src/eu/kanade/tachiyomi/extension/all/lanraragi/LANraragi.kt
```diff
@@ -0,0 +1,220 @@
...
+
+        screen.addPreference(hostnamePref)
+        screen.addPreference(apikeyPref)
+    }
+
+    private fun genericMangaParse(response: Response): MangasPage {
+        val archives = Gson().fromJson<List<Archive>>(response.body()!!.string())
+
+        val mangas = ArrayList<SManga>()
+        val queries = response.request().url().queryParameter("query")?.trim()?.split(Regex(" (?=(?:[^\"]*\"[^\"]*\")*[^\"]*$)"))
```
arkon: Extract the regex into a lazily instantiated class-level variable so it doesn't need to be recreated constantly. Same for other regexes.

src/all/lanraragi/src/eu/kanade/tachiyomi/extension/all/lanraragi/LANraragi.kt
```diff
@@ -0,0 +1,220 @@
...
+
+        archives.forEach archiveForEach@{ a ->
```
arkon: What's the `archiveForEach` tag for?

src/all/lanraragi/src/eu/kanade/tachiyomi/extension/all/lanraragi/LANraragi.kt
```diff
@@ -0,0 +1,220 @@
...
+                    } else {
+                        val child = split[0].toLowerCase()
+                        shortTags.add(child)
+                    }
+                }
+            }
+
+            var tagsMatched = true
+            var titleMatched = false
+            queries?.forEach queryForEach@{ q ->
```
arkon: Same here, what's the `queryForEach` tag for?

src/all/lanraragi/src/eu/kanade/tachiyomi/extension/all/lanraragi/LANraragi.kt
```diff
@@ -0,0 +1,220 @@
...
+        }
+
+        return MangasPage(mangas, false)
+    }
+
+    private val preferences: SharedPreferences by lazy {
+        Injekt.get<Application>().getSharedPreferences("source_$id", 0x0000)
+    }
+
+    private fun getThumbnailsUri(id: String): String {
```
arkon: Nitpick: `getThumbnailUri`

Seflyx:
Happen to really busy last year so I hadn't much time to look into this. Apparently LANraragi updated their API so I have to look over everything again.

src/all/lanraragi/src/eu/kanade/tachiyomi/extension/all/lanraragi/LANraragiFactory.kt
Seflyx: Removed and decided to use "all" only instead

src/all/lanraragi/src/eu/kanade/tachiyomi/extension/all/lanraragi/Archive.kt
Seflyx: I already transform them into data class

src/all/lanraragi/src/eu/kanade/tachiyomi/extension/all/lanraragi/LANraragi.kt
Seflyx: I tried using const, but throw me an error instead so I didn't kinda bother with it since val already reach the purpose?

src/all/lanraragi/src/eu/kanade/tachiyomi/extension/all/lanraragi/LANraragi.kt
Seflyx: yes

src/all/lanraragi/src/eu/kanade/tachiyomi/extension/all/lanraragi/LANraragi.kt
Seflyx: Removed the regex

src/all/lanraragi/src/eu/kanade/tachiyomi/extension/all/lanraragi/LANraragi.kt
Seflyx: Removed the regex

src/all/lanraragi/src/eu/kanade/tachiyomi/extension/all/lanraragi/LANraragi.kt
Seflyx: Removed since their API now support searching

src/all/lanraragi/src/eu/kanade/tachiyomi/extension/all/lanraragi/LANraragi.kt
Seflyx: I plan to use that function in multiple code blocks later

src/all/lanraragi/src/eu/kanade/tachiyomi/extension/all/lanraragi/LANraragi.kt
Seflyx: Removed since API now support searching

Seflyx:
Updated to their newer API

src/all/lanraragi/src/eu/kanade/tachiyomi/extension/all/lanraragi/LANraragi.kt
Seflyx: Replaced them with listOf

src/all/lanraragi/build.gradle
```diff
@@ -0,0 +1,21 @@
+apply plugin: 'com.android.application'
+apply plugin: 'kotlin-android'
+
+ext {
+    appName = 'Tachiyomi: LANraragi'
+    pkgNameSuffix = 'all.lanraragi'
+    extClass = '.LANraragi'
+    extVersionCode = 3
```
arkon: Don't really need to increment since it was never actually released as part of this repo.

src/all/lanraragi/src/eu/kanade/tachiyomi/extension/all/lanraragi/LANraragi.kt
```diff
@@ -0,0 +1,255 @@
...
+
+        if (query.isNotEmpty()) {
+            uri.appendQueryParameter("filter", query)
+        }
+
+        return GET(uri.toString())
+    }
+
+    override fun searchMangaParse(response: Response): MangasPage {
+        val jsonResult = Gson().fromJson<ArchiveSearchResult>(response.body()!!.string())
```
arkon: Have a single instance of `Gson()` declared at the class level. It's an expensive object to create, so we don't want to do that every time a manga is parsed.

src/all/lanraragi/src/eu/kanade/tachiyomi/extension/all/lanraragi/LANraragi.kt
```diff
@@ -0,0 +1,255 @@
...
+                val uriBuild = uri.build()
+
+                url = "${uriBuild.encodedPath}?${uriBuild.encodedQuery}"
+                chapter_number = 1F
+                name = "Chapter"
+            })
+    }
+
+    override fun pageListParse(response: Response): List<Page> {
+        val archivePage = Gson().fromJson<ArchivePage>(response.body()!!.string())
```
arkon: Same here re: `Gson()`

src/all/lanraragi/src/eu/kanade/tachiyomi/extension/all/lanraragi/LANraragi.kt
```diff
@@ -0,0 +1,255 @@
...
+
+        val uri = getApiUriBuilder("/api/extract")
+        uri.appendQueryParameter("id", id)
+
+        return listOf(
+            SChapter.create().apply {
+                val uriBuild = uri.build()
+
+                url = "${uriBuild.encodedPath}?${uriBuild.encodedQuery}"
+                chapter_number = 1F
```
arkon: I commented on this in an earlier revision too, but is this right? Seems weird that a manga only has one chapter?

src/all/lanraragi/build.gradle
Seflyx: I'm putting it back to 1.

src/all/lanraragi/src/eu/kanade/tachiyomi/extension/all/lanraragi/LANraragi.kt
Seflyx: Done

src/all/lanraragi/src/eu/kanade/tachiyomi/extension/all/lanraragi/LANraragi.kt
Seflyx: Done

src/all/lanraragi/src/eu/kanade/tachiyomi/extension/all/lanraragi/LANraragi.kt
Seflyx: Yes, there's a demo for lanraragi that you can check at https://lrr.tvc-16.science/ to see that every title is single chapter

arkon: Seems weird to me, but okay.

PR #1428, opened 2019-08-26
Genre Filtering on MRM

happywillow0:
New Features: Added filter lists to the MRM extension. 

This pull request pulls the home page tag cloud and category list and is able to parse it into the filter section of the extension. You can now browse with filters including:
- Genres
- Popular tags (tags that show up on the bottom of the website; sadly not all the tags available but you can search )
- Categories (dj sources) 

The list is live parsed off the website so it should be up to date with whatever the admins are using.


happywillow0:
Side note, just in case, I'll be off the grid on vacation for a week so I'll be unavailable to change anything if needed in case something actually happens with this pull request.

happywillow0:
I do not believe I will be adding anything more to this from my end. It works pretty well in my testing. 

Please let me know if there is anything you would like me to change or add.

PR #1434, opened 2019-08-28
Make uses-feature:tachiyomi.extension optional, and Allow backups.

Efreak:
1. I am trying to republish the Tachiyomi Extensions in an fdroid repository, and uses-feature:tachiyomi.extension is breaking it. In order to install tachiyomi extensions at all via fdroid, I have to enable "include incompatible versions", and updating the ~~app~~ extension is entirely impossible.

2. allowBackup is almost certainly not intended to be enabled. This isn't a banking app, and the extensions don't store private data. I noticed this during above.

### DIFF
--- a/AndroidManifest.xml
+++ b/AndroidManifest.xml
@@ -4,7 +4,7 @@
 
     <uses-feature android:name="tachiyomi.extension" android:required="false"/>
 
-    <application android:icon="@mipmap/ic_launcher" android:allowBackup="false" android:label="${appName}">
+    <application android:icon="@mipmap/ic_launcher" android:label="${appName}">
 
         <meta-data android:name="tachiyomi.extension.class" android:value="${extClass}" />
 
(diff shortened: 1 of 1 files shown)


Efreak:
> > I am trying to republish the Tachiyomi Extensions in an fdroid repository
> 
> Why?

> Yea i am lost why you are trying to push these to fdroid

I had a long explanation, and I accidentally ctrl+shift+w.. :(

Basically, fdroid allows background updates similar to the play store. Making the feature optional should preserve the ability to check for plugins via the requested feature while also allowing it to be installed automatically. I'm not sure why reqFeatures was used here instead of some other method of finding extensions (a shareduserid, custom permissions, etc). Therefore I'm trying to push them to a custom fdroid repo so they can be updated automatically in the background with zero interaction.

> > allowBackup is almost certainly not intended to be enabled
> 
> You mean disabled?
yes

arkon:
I'd prefer users only getting things from one place since it simplifies supporting them.

PR #1460, opened 2019-09-04
MangaRock chapter delay fix 2

SnakeDoc83:
It's interesting, the previous push _does_ fix the delay issue for some manga, but not all.  I've reverted that change and I'm pushing a different one.  I've checked the latest 20-25 manga at different points the last few days with this change and they've all been good; so I think this might be the one.  The solution was easy to overlook, needed to capitalize a parameter for chapter requests.

### DIFF
--- a/src/en/mangarock/build.gradle
+++ b/src/en/mangarock/build.gradle
@@ -5,7 +5,7 @@ ext {
     appName = 'Tachiyomi: Manga Rock'
     pkgNameSuffix = 'en.mangarock'
     extClass = '.MangaRock'
-    extVersionCode = 13
+    extVersionCode = 14
     libVersion = '1.2'
 }
 
--- a/src/en/mangarock/src/eu/kanade/tachiyomi/extension/en/mangarock/MangaRock.kt
+++ b/src/en/mangarock/src/eu/kanade/tachiyomi/extension/en/mangarock/MangaRock.kt
@@ -12,7 +12,12 @@ import eu.kanade.tachiyomi.source.model.Page
 import eu.kanade.tachiyomi.source.model.SChapter
 import eu.kanade.tachiyomi.source.model.SManga
 import eu.kanade.tachiyomi.source.online.HttpSource
-import okhttp3.*
+import okhttp3.MediaType
+import okhttp3.OkHttpClient
+import okhttp3.Request
+import okhttp3.RequestBody
+import okhttp3.Response
+import okhttp3.ResponseBody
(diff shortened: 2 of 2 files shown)


joseph619:
Some Manga from Manga Rock Definitive version are not appearing on browse (for example, Solo Leveling)

brassica77:
Same for Youjo Senki.

SnakeDoc83:
The website/API won't provide licensed content based on a user's
country.  Definitive
displays licensed manga regardless of that.  I don't know if there's a
specific country the extension could be set to that would have 0
restrictions.

joseph619:
@SnakeDoc83 Set it to Canada, I can see everything

Code027:
It's apparent that a lot more of manga has been updating properly but there are still quite a handful that gets delayed. There's definitely some improvement from fix 1.

SnakeDoc83:
Which manga still have an issue?

Code027:
I was browsing through the Latest and found out that:
The Story of an Onee-San Who Wants to Keep a High School Boy;
The Teacher Can Not Tell Me Love;
and AoT to add, still has the issue.
I've checked in web view and they are already updated there, though I believe that those might not be the only manga that still have the issue.

pavan3999:
> I was browsing through the Latest and found out that:
> The Story of an Onee-San Who Wants to Keep a High School Boy;
> The Teacher Can Not Tell Me Love;
> and AoT to add, still has the issue.
> I've checked in web view and they are already updated there, though I believe that those might not be the only manga that still have the issue.

for me Shikkaku Mon no Saikyou Kenja, Ore dake Haireru Kakushi Dungeon: Kossori Kitaete Sekai Saikyou (~11hrs delay both) holy ancestor (~15hrs delay) notably found

SnakeDoc83:
Tested a couple manga that were mentioned, setting the `Country=` for API requests to something rather than leaving it blank seems to help.  I've been reluctant to change that in the extension though, thinking that there was a reason it was written that way.

CarlosEsco:
@SnakeDoc83   you could make it a configurable property so the user has option to change country

joseph619:
Please close this, Manga Rock site will be taken down next week

SnakeDoc83:
@CarlosEsco Yeah, that would make sense.

@joseph619 This has been closed since before you first commented.

joseph619:
@SnakeDoc83 Manga Rock officially stated they taking down the site next week, that's why I said that. Nvm

PR #1467, opened 2019-09-05
Heavenmanga date parse fix

SnakeDoc83:
Closes https://github.com/inorichi/tachiyomi-extensions/issues/1461

They've changed URLs twice since an issue was opened; I've updated baseUrl to what they're using currently.


ruuno123:
Can i ask where i can fin now the haven &holy manga ?? (What extansion.

SnakeDoc83:
How are you finding this old PR but not the one that renamed the extension to zBulu?  Anyway, zBulu is the extension for them now.

Issue #1469, opened 2019-09-06
Genre filtering for Mangakakalot


joseph619:
Currently, Mangakakalot 1.2.6 lacks Genre filtering. Please add this option as it's an excellent replacement source for Manga Rock.

rama982:
ok i will add filtering

joseph619:
Thanks bro

joseph619:
Thanks bro

PR #1471, opened 2019-09-06
MangaKakalot: Add comics filtering

rama982:
#1469


joseph619:
Has the extension been updated already? I don't see any update on Extension page

joseph619:
@rama982 The extension still doesn't have filtering (like Mangadex or Mangasee), just tags.

PR #1474, opened 2019-09-07
Mangakakalot: replicate query normalization

### LINKED ISSUE #1464: [Issue]: Mangakakalot special character '
Tachiyomi 0.8.4 - Mangakakalot English 1.2.6

When trying to search a manga with special character ' (apostrophe), Mangakakalot loads endlessly not giving ever any result, you have to write word without it or replacing it with a space (It's => Its or It s - but maybe sometimes it won't find anything with these workaround).

arkon:
Closes #1464

### DIFF
--- a/src/en/mangakakalot/build.gradle
+++ b/src/en/mangakakalot/build.gradle
@@ -5,7 +5,7 @@ ext {
     appName = 'Tachiyomi: MangaKakalot'
     pkgNameSuffix = 'en.mangakakalot'
     extClass = '.Mangakakalot'
-    extVersionCode = 7
+    extVersionCode = 8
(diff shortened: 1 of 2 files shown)


SnakeDoc83:
The Manganelo extension has the same issue, this could probably be applied to it as well.  As an aside, there's 2 source requests open that have the same layout as those 2.  They're all more-or-less clones of each other.  Could make a factory, could just close them.

https://github.com/inorichi/tachiyomi-extensions/issues/1222

https://github.com/inorichi/tachiyomi-extensions/issues/1037

arkon:
Agreed, should be a factory.

PR #1486, opened 2019-09-09
ReadMangaToday update

SnakeDoc83:
Closes https://github.com/inorichi/tachiyomi-extensions/issues/1485

Fixed the issue with chapters, added getting thumbnails for popularMangaRequest, made a few tweaks to get rid of some warnings.

### DIFF
--- a/src/en/readmangatoday/build.gradle
+++ b/src/en/readmangatoday/build.gradle
@@ -5,7 +5,7 @@ ext {
     appName = 'Tachiyomi: ReadMangaToday'
     pkgNameSuffix = 'en.readmangatoday'
     extClass = '.Readmangatoday'
-    extVersionCode = 6
+    extVersionCode = 7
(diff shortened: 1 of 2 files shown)


Black-Seraph:
Seems to still be broken here.

SnakeDoc83:
Which chapters of which manga are you having an issue with?

Black-Seraph:
Attack on Titan, Chapter 121.

But any chapter I don't have a local copy of works for testing this really.

SnakeDoc83:
It's working for me, but I have noticed it can be a little wonky at times.  Are you using 1.2.8 of the extension?  I ask because this is the PR for 1.2.7.  If you are on the latest version, try back backing out of the chapter, wait a few seconds, and open it again.  Beyond that you can try talking with the folks on discord and maybe something will come to light that will lead to an improvement for the extension.

PR #1504, opened 2019-09-14
RawLH: Update page list parse

### LINKED ISSUE #1501: RawLH is only displaying credit pages
BUGS:
RawLH only showing single page of credit page only

Tachiyomi 0.8.4 Build 13 April 2019 9:16 pm
RawLH 1.2.9

rama982:
Fix #1501

Fyi, when i updating code they just changed chapter img url mechanism.
this is latest changes on the site.

### DIFF
--- a/src/ja/rawlh/build.gradle
+++ b/src/ja/rawlh/build.gradle
@@ -5,7 +5,7 @@ ext {
     appName = 'Tachiyomi: RawLH'
     pkgNameSuffix = 'ja.rawlh'
     extClass = '.Rawlh'
-    extVersionCode = 9
+    extVersionCode = 10
     libVersion = '1.2'
 }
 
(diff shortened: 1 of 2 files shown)


SnakeDoc83:
`document.select("script")`  
There are dozens of scripts you're selecting by using "script" on its own.  You could be more specific with something like `document.select("script:containsData(imgsrc)")`

`if (encodedUrl != "")` 
If there are cases where encodedUrl would be empty, shouldn't there be an else statement to go with this?  If there are no cases where encodedUrl is empty, then why check?  Might be something to consider.

rama982:
@SnakeDoc83 thanks for review, i will recheck the code.

rama982:
i already update the code and tested with > 5 titles manga and all work good.

rama982:
just when this PR merged they updated the mechanism again, so better wait some days until they not changed stuff or update again?

PR #1510, opened 2019-09-16
all/webtoons: Add additonal languages

### LINKED ISSUE #1492: [SOURCE REQUEST : Line Webtoon (Indonesian Language)
The site is https://www.webtoons.com/id , I know there's already webtoon extension for tachiyomi. So maybe instead of adding new extension, can we have a selection for selecting the language? Because every language afaik have different sets of comics.
image

mook:
This adds more languages where webtoons has official comics:

- Chinese (Traditional)
- Indonesian
- Thai

Closes #1492

Note that Webtoons also has Chinese (Simplified) pages, but those go off into a separate host name, so I'm not touching that for now.

Also, as far as I can tell the French (unofficial translations) support is currently busted anyway; this PR does not fix it (since there's already quite a bit of changes here).  I might do a follow-up PR if that seems useful... though I note that there hasn't been an issue filed for it.

I'm not familiar with Kotlin (or Android development in general); please let me know if there is anything I can improve, I'd be happy to change things.

Thanks for the app and the various extensions!

### DIFF
--- a/src/all/webtoons/src/eu/kanade/tachiyomi/extension/th/webtoons/WebtoonsThai.kt
+++ b/src/all/webtoons/src/eu/kanade/tachiyomi/extension/th/webtoons/WebtoonsThai.kt
@@ -1,0 +1,11 @@
+package eu.kanade.tachiyomi.extension.th.webtoons
+
+import eu.kanade.tachiyomi.extension.all.webtoons.WebtoonsDefault
+import java.text.SimpleDateFormat
+import java.util.*
+
+class WebtoonsThai: WebtoonsDefault("th") {
+    override fun chapterParseDate(date: String): Long {
+        return SimpleDateFormat("d MMM yyyy", Locale("th")).parse(date).time
+    }
+}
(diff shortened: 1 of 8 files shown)


SnakeDoc83:
Just letting you know to check your PR, your build is failing.

mook:
Sorry about that, I missed a file in the commit. Updated, plus bumped the extVersionCode as documented.  Thanks for the prompt response (to my broken PR)!

src/all/webtoons/src/eu/kanade/tachiyomi/extension/th/webtoons/WebtoonsThai.kt
```diff
@@ -0,0 +1,11 @@
+package eu.kanade.tachiyomi.extension.th.webtoons
+
+import eu.kanade.tachiyomi.extension.all.webtoons.WebtoonsDefault
+import java.text.SimpleDateFormat
+import java.util.*
+
+class WebtoonsThai: WebtoonsDefault("th") {
+    override fun chapterParseDate(date: String): Long {
+        return SimpleDateFormat("d MMM yyyy", Locale("th")).parse(date).time
+    }
```
arkon: Is there no constant for this locale?

mook: Correct; there is no constant listed for Thai in the [documentation](https://developer.android.com/reference/java/util/Locale).

PR #1522, opened 2019-09-17
FoolSlide: Update Hentai Cafe

rama982:
* Fix genres and artist parsing
* Set status as completed
* Set chapters number to 1


rama982:
* Fix latest update load if there is no thumbnail

rama982:
Do not merge yet

Issue #1543, opened 2019-09-21
MyReadingManga Extension has nonfunctioning manga covers after restoring a backup


Eminane:
After restoring a backup, I noticed that the manga covers from the MyReadingManga Extensions are missing.
![image](https://user-images.githubusercontent.com/55626138/65373903-85fe2d80-dc51-11e9-847c-84e3865b5f3b.png)
The manga covers from other sources loaded normally.
![image](https://user-images.githubusercontent.com/55626138/65373910-aded9100-dc51-11e9-8391-8fb6763e4223.png)
There was no change after restoring the backup twice more, restarting the phone twice, uninstalling and installing the extension, clearing chapter cache data, clearing cookies, and refreshing library metadata.
When viewing the MyReadingManga catalog, the manga that have not been added to the library yet had working covers. After adding the manga with working covers from the MyReadingManga Extension to the library, the covers continued to work.
![image](https://user-images.githubusercontent.com/55626138/65373973-8945e900-dc52-11e9-96b2-e41d3b2e8910.png)
I tried manually changing the cover but received an error message.
![image](https://user-images.githubusercontent.com/55626138/65373988-b72b2d80-dc52-11e9-972c-4dfae97a4796.png)

Also, unrelated to the cover errors, creating shortcuts to manga restored from the backup from the MyReadingManga Extension results in an error as well.
![image](https://user-images.githubusercontent.com/55626138/65374013-f9ed0580-dc52-11e9-982a-a8340fcb1ba0.png)

joseph619:
@happywillow0

happywillow0:
Hm, interesting. I've been working on the extension of late but I don't think have made any drastic changes on how things actually work. My changes don't include changes to the cover images. 

I remember having an issue a while back where the covers were failing to update. My solution at the time was to remove the items from my library, cleared the database (deleting manga not in my library) and adding it back in. 

I'm not totally sure what is included as part of the backup and restore process and what is stored in the database. I'm assuming you tried the "Refresh library metadata" option?

MRM did have a period of time where the images were lost and they had to re-upload but I'm not sure if that affects anything with the app's database. 

I'm going to test the backup and restore on one of my other phones.

happywillow0:
Also, I remember and found again this open issue on the main app side about error setting covers which may or may not be related.
https://github.com/inorichi/tachiyomi/issues/1663

Eminane:
[...]

PR #1546, opened 2019-09-22
MRM  - mangaDetailsParse Cover enhancement

happywillow0:
Added Conditional Cover parsing into the mangaDetailsParse 
Fixes Issue #1543 

There is no thumbnail data available form the manga "detials" page. The extension will conditional parsing in the following way: 

- No Change: If user browses the catalog and adds a manga to the library, Tachiyomi uses the proper cover.
- Fix: If user restores a backup, the initial cover will be the 1st image. (This may or may not be a cover aka scanlator credits). This must be done after opening the app but before opening the catalog. 
- Breaks: The cover will change to the 1st image only if user requests a metadata update before they view the catalog.

Also included is a minor enhancement that adds the scanlator group to the chapter list.

### DIFF
--- a/src/all/myreadingmanga/src/eu/kanade/tachiyomi/extension/all/myreadingmanga/MyReadingManga.kt
+++ b/src/all/myreadingmanga/src/eu/kanade/tachiyomi/extension/all/myreadingmanga/MyReadingManga.kt
@@ -142,15 +142,16 @@ open class MyReadingManga(override val lang: String) : ParsedHttpSource() {
         val date = parseDate(document.select(".entry-time").text())
         val mangaUrl = document.baseUri()
         val chfirstname = document.select(".chapter-class a[href*=$mangaUrl]")?.first()?.text()?.ifEmpty { "Ch. 1" }?.capitalize() ?:"Ch. 1"
+        val scangroup= document.select(".entry-terms a[href*=group]")?.first()?.text()
         //create first chapter since its on main manga page
-        chapters.add(createChapter("1", document.baseUri(), date, chfirstname))
+        chapters.add(createChapter("1", document.baseUri(), date, chfirstname, scangroup))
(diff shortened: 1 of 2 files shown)


src/all/myreadingmanga/src/eu/kanade/tachiyomi/extension/all/myreadingmanga/MyReadingManga.kt
```diff
@@ -130,6 +132,8 @@ open class MyReadingManga(override val lang: String) : ParsedHttpSource() {
             "Completed" -> SManga.COMPLETED
             else -> SManga.UNKNOWN
         }
+        // Set first image as thumbnail only if user has not entered from catalog section.
```
arkon: Is the thumbnail normally the first image anyway? If so, I don't think you really need to check for something like `openbrowse`. You're fetching the data here regardless, it doesn't hurt to set the image again unless it actually changes.

src/all/myreadingmanga/src/eu/kanade/tachiyomi/extension/all/myreadingmanga/MyReadingManga.kt
```diff
@@ -37,9 +37,11 @@ open class MyReadingManga(override val lang: String) : ParsedHttpSource() {
 
     override fun latestUpdatesRequest(page: Int) = popularMangaRequest(page)
 
+    private var openbrowse : String? = null // Int a string to test if user opened "Browse" or not.
```
arkon: Use a boolean for this sort of thing.

happywillow0:
The approach is a bit flawed but at the time I couldn't think of a better way. In my experience, the 1st page is a credits page from the scan group rather than the cover. 

However, hold off a small bit. I'm looking to the API of wordpress and am making some progress.

PR #1547, opened 2019-09-22
Webtoons: Fan translations

### LINKED ISSUE #1517: [SOURCE REQUEST] : Line Webtoon (Portuguese Language)
Line Webtoon site has a translation area made by fans, and I would like to know if the possibility of adding the extension "Webtoon all" if not, patience

[Webtoon Translate](https://translate.webtoons.com/translate) 

I could not find a page of this with all of them in Portuguese, so it may not be possible to add this extension, but it costs nothing to ask.

mook:
all/webtoons: update fan-translation languages
    
This reworks the Webtoons fan translation languages to actually work now, as well as add all the previously unlisted languages (we just had French before).

Fixes #1517

This includes so many languages, we might want to consider splitting it off into its own extension instead.  I'm happy to do that if it seems like a good idea.

### DIFF
--- a/src/all/webtoons/build.gradle
+++ b/src/all/webtoons/build.gradle
@@ -5,7 +5,7 @@ ext {
     appName = 'Tachiyomi: Webtoons'
     pkgNameSuffix = 'all.webtoons'
     extClass = '.WebtoonsFactory'
-    extVersionCode = 8
+    extVersionCode = 9
     libVersion = '1.2'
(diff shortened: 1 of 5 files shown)


arkon:
> This includes so many languages, we might want to consider splitting it off into its own extension instead. I'm happy to do that if it seems like a good idea.

I'm fine with keeping it together too.

PR #1550, opened 2019-09-23
GetManhwa - new extension

SnakeDoc83:
Closes https://github.com/inorichi/tachiyomi-extensions/issues/1237


src/en/getmanhwa/src/eu/kanade/tachiyomi/extension/en/getmanhwa/GetManhwa.kt
```diff
@@ -0,0 +1,254 @@
...
+    }
+
+    override fun popularMangaParse(response: Response): MangasPage {
+        return parseMangaDocument(response.asJsoup())
+    }
+
+    private fun parseMangaDocument(document: Document): MangasPage {
+        val mangas = mutableListOf<SManga>()
+
+        document.select(popularMangaSelector()).map{ mangas.add(popularMangaFromElement(it)) }
```
arkon: Could just do something like:

```kotlin
val mangas = document.select(popularMangaSelector()).map { popularMangaFromElement(it) }
```

SnakeDoc83: I did it this way to use parseMangaDocument in searchMangaRequest

arkon: Yeah, I just mean you can directly use the result of `map` instead of using it like a `forEach`, but doesn't really matter.

PR #1551, opened 2019-09-23
MyReadingManga Thumbnail from API

### LINKED ISSUE #1543: MyReadingManga Extension has nonfunctioning manga covers after restoring a backup

happywillow0:
2nd attempt to fix #1543 

Cover information is pulled from the wordpress API. User however must be logged in to the website for the API to return data. There is a very basic (possibly flawed) check to see if the user is logged in or not. 

Also included is a minor enhancement that adds the scanlator group to the chapter list.

Please feel free to point out how it can be optimized or flaws fixed.

### DIFF
--- a/src/all/myreadingmanga/build.gradle
+++ b/src/all/myreadingmanga/build.gradle
@@ -5,7 +5,7 @@ ext {
     appName = 'Tachiyomi: MyReadingManga'
     pkgNameSuffix = 'all.myreadingmanga'
     extClass = '.MyReadingMangaFactory'
-    extVersionCode = 25
+    extVersionCode = 26
     libVersion = '1.2'
 }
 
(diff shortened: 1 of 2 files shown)


SnakeDoc83:
An alternative approach: https://github.com/inorichi/tachiyomi-extensions/compare/master...SnakeDoc83:mrm

PR #1562, opened 2019-09-26
Add Toomics global

### LINKED ISSUE #1120: [Source Request] [English] toomics.com
please add 
https://toomics.com/en

alessandrojean:
Add the global version of Toomics. Should close #1120.
Since the main website is different and I don't know Korean, I did not included it in the extension.

The VIP chapters can't be read in the extension. Maybe with the login done via web view, they can be read, but I have not tried since I don't have a VIP account. Need someone with such account to try.

To disable the "family mode" it's necessary to open in web view and disable it on the site to set the cookie. Having done this, just open the catalog again in the app. When this PR is merged, I will add this information to the website FAQ.


SnakeDoc83:
>To disable the "family mode" it's necessary to open in web view and disable it on the site to set the cookie. 

Looks like you can skip family mode by using "https://global.toomics.com/en/index/set_display/?display=A&return=/en" for `baseUrl`

Also, like a couple other sources, this one has free and locked chapters.  Inevitable, people will open issues asking what's "wrong" with this extension.  Last extension I made that had locked chapters I added "(Limited free chapters)" in `name` and kept the chapter list to only the free chapters.  Something to consider.

alessandrojean:
> Looks like you can skip family mode by using "https://global.toomics.com/en/index/set_display/?display=A&return=/en" for `popularMangaRequest()`

I don't know if this will work because in the first time it redirects to another page asking if the person is an adult. Also, I think the ideal would be that the family mode is enabled by default, allowing the person to disable it if they want in the extension settings.

> Last extension I made that had locked chapters I added "(Limited free chapters)" in `name` and kept the chapter list to only the free chapters.

I had initially thought of doing that, I will change it and filter only the free chapters.

arkon:
> I don't know if this will work because in the first time it redirects to another page asking if the person is an adult.

Doesn't seem to do that for me.

alessandrojean:
> Doesn't seem to do that for me.

Maybe because sometime in the past you already disabled it.

There are three urls in this process:

1. https://global.toomics.com/en/index/set_family_mode/?family_mode=N&return=/en
2. https://global.toomics.com/en/age_verification?cancel_return=L2Vu&return_url=L2Vu
3. https://global.toomics.com/en/index/set_display/?display=A&return=/en

At the first time a request is made to (1), it returns a `302` and redirects to (2), that prompts the user with [this page](https://imgur.com/o9zFZIB). When the person clicks on "Yes", it redirects to (3) with `display=A`, and in "No" to `display=N`. This process create in the cookies two variables in the section (`GTOOMICScisession`, that is encoded), `tw_adultYN` that has the date when the person informed that is over 18, and `family_mode` with the value `N`. This allow the server to show the R18+ titles.

What I've discovered now is that this behavior only happens in the first time, and if (in the first time) a request is made directly to (3), it sets the two necessary variables. However, to re-enable (or disable) the family mode, the (3) does not work anymore for a while, it's needed to use the (1). The site behavior is confusing. Beyond that, the `tw_adultYN` has some kind of expire date, so in some time to disable family mode (again) it's needed to follow the previous process again.

What I can do is always make a request in `popularMangaRequest` and `latestUpdatesRequest` to (3), and always show the R18+ titles. I will change to that now, and later someone can change the extension to allow enable or disable the family mode.

alessandrojean:
Now all R18+ titles are always displayed and only free chapters are listed.

joseph619:
Will using the extension cause the Toomics account to get banned? Toomics is extremely aggressive about piracy.

alessandrojean:
@joseph619 It's better that you don't login into your account (using webview) and use the extension only to read the free chapters that does not require you to login. Since they are free and public to anyone to read in the website, I don't think this can cause some kind of trouble.

joseph619:
Not logging in should be fine I think. However logging in is something I will recommend against, even for experiment. Not worth getting account banned for a premium service.

src/all/toomics/src/eu/kanade/tachiyomi/extension/all/toomics/ToomicsFactory.kt
```diff
@@ -0,0 +1,31 @@
...
+
+import eu.kanade.tachiyomi.source.Source
+import eu.kanade.tachiyomi.source.SourceFactory
+import java.util.*
+
+class ToomicsFactory : SourceFactory {
+    override fun createSources(): List<Source> = getAllToomics()
+}
+
+class ToomicsEnglish : ToomicsGlobal("en", "MMM dd, yyyy", Locale.ENGLISH)
```
arkon: I feel like it actually makes more sense to pass in a `SimpleDateFormat` rather than the arguments separately.

src/all/toomics/src/eu/kanade/tachiyomi/extension/all/toomics/ToomicsFactory.kt
```diff
@@ -0,0 +1,31 @@
...
+class ToomicsSimplifiedChinese : ToomicsGlobal("sc", "yyyy.MM.dd", Locale.SIMPLIFIED_CHINESE, "zh", "简体")
+class ToomicsTraditionalChinese : ToomicsGlobal("tc", "yyyy.MM.dd", Locale.TRADITIONAL_CHINESE, "zh", "繁體")
+class ToomicsSpanishLA : ToomicsGlobal("mx", "d MMM, yyyy", Locale("es", "419"), "es", "LA")
+class ToomicsSpanish : ToomicsGlobal("es", "d MMM, yyyy", Locale("es", "419"), "es", "ES")
+class ToomicsItalian : ToomicsGlobal("it", "d MMM, yyyy", Locale.ITALIAN)
+class ToomicsGerman : ToomicsGlobal("de", "d. MMM yyyy", Locale.GERMAN)
+class ToomicsFrench : ToomicsGlobal("fr", "dd MMM. yyyy", Locale.ENGLISH)
+class ToomicsPortuguese : ToomicsGlobal("por", "d 'de' MMM 'de' yyyy", Locale("pt", "BR"), "pt")
+
+fun getAllToomics(): List<Source> = listOf(
```
arkon: This function is a bit unnecessary, could just do it directly in `createSources()`

src/all/toomics/src/eu/kanade/tachiyomi/extension/all/toomics/ToomicsGlobal.kt
```diff
@@ -0,0 +1,152 @@
...
+
+    override fun chapterFromElement(element: Element): SChapter = SChapter.create().apply {
+        val num = element.select("div.cell-num span.num").text()
+        val isVip = element.attr("onclick").contains("login")
+        val numText = if (num.isNotEmpty()) "$num - " else ""
+
+        name = numText + element.select("div.cell-title strong.tit").first().ownText()
+        chapter_number = num.toFloatOrNull() ?: 0f
+        date_upload = parseChapterDate(element.select("div.cell-time time").text()!!)
+        scanlator = "Toomics"
```
arkon: Is this really necessary?

src/all/toomics/src/eu/kanade/tachiyomi/extension/all/toomics/ToomicsGlobal.kt
```diff
@@ -0,0 +1,152 @@
...
+    override fun fetchChapterList(manga: SManga): Observable<List<SChapter>> {
+        return super.fetchChapterList(manga)
+            .map { it.reversed().filter { c -> !c.url.startsWith("VIP") } }
+    }
+
+    override fun chapterListSelector(): String = "section.ep-body ol.list-ep li.normal_ep a"
+
+    override fun chapterFromElement(element: Element): SChapter = SChapter.create().apply {
+        val num = element.select("div.cell-num span.num").text()
+        val isVip = element.attr("onclick").contains("login")
```
arkon: Could these potentially be filtered out earlier?

src/all/toomics/src/eu/kanade/tachiyomi/extension/all/toomics/ToomicsFactory.kt
alessandrojean: Yeah, I will change.

src/all/toomics/src/eu/kanade/tachiyomi/extension/all/toomics/ToomicsGlobal.kt
alessandrojean: I fell that when there's no scanlator set, the chapter in the listing at the UI becomes kinda "empty", so I set it manually.

src/all/toomics/src/eu/kanade/tachiyomi/extension/all/toomics/ToomicsGlobal.kt
alessandrojean: Hmm, probably yes, will try too.

PR #1568, opened 2019-09-27
Add shqqaa.com

yshalsager:


src/ar/shqqaa/src/eu/kanade/tachiyomi/extension/ar/shqqaa/Shqqaa.kt
```diff
@@ -0,0 +1,117 @@
...
+            manga.title = element.select("small").first().text().split(", ")[0]
+        }
+        manga.thumbnail_url = element.select("img").first().attr("data-src")
+        return manga
+    }
+
+    override fun latestUpdatesNextPageSelector(): String? = null
+
+    // Search
+    override fun searchMangaRequest(page: Int, query: String, filters: FilterList): Request = throw Exception("Not used")
```
arkon: Return an empty observable instead, otherwise it may cause issues with global search. See Xkcd for an example.

yshalsager: @arkon May you check the new changes?

PR #1570, opened 2019-09-27
add westmanga.info

arifr007:


as280093:
Here is the Icons based on their Android app
Use this instead
[westmanga.zip](https://github.com/inorichi/tachiyomi-extensions/files/3665042/westmanga.zip)

arifr007:
@as280093 done

PR #1572, opened 2019-09-27
fix pecintakomik.net

arifr007:
fix bugs blank retry each image


arkon:
Update version in build.gradle

PR #1573, opened 2019-09-27
add komikindo.web.id

arifr007:


SnakeDoc83:
These sources you're adding and fixing all use the wordpress-mangastream CMS.  I was planning on making a factory extension for them after I finish a different one I'm working on, but you could do that instead if you're interested.

src/id/komikindo/src/eu/kanade/tachiyomi/extension/id/komikindo/KomikIndo.kt
```diff
@@ -0,0 +1,237 @@
...
+    override fun searchMangaNextPageSelector() = popularMangaNextPageSelector()
+
+    override fun mangaDetailsParse(document: Document): SManga {
+        val infoElm = document.select(".listinfo > ul > li")
+        val manga = SManga.create()
+        infoElm.forEachIndexed { index, element ->
+            val infoTitle = element.select("b").text()
+            var infoContent = element.text()
+            when {
+                infoTitle.toLowerCase().contains("type") -> println("position is 1")
```
arkon: Are you going to use these values?

arifr007: damn, so sorry for got to remove it, i will update it later, 
about factory, ill do it later i want to optimized the source first

PR #1583, opened 2019-09-30
Merge HolyManga and HeavenManga

SnakeDoc83:
They're effectively the same site under different domains.  Updated a couple functions and changed icons.  Potentially closes https://github.com/inorichi/tachiyomi-extensions/issues/1575

Assuming this is accepted, is it preferable to remove HeavenManga in this PR or in a separate PR?


arkon:
> is it preferable to remove HeavenManga in this PR or in a separate PR?

Yup.

I'm wondering if it'll be an issue having this as `.en` at some point.

SnakeDoc83:
I think it's fine.  My view is that this is one source operating on two domains.  If at some point someone wants to add another source then this extension would be treated like a single-source extension: remove it and add a new one to `.all` with a new name.

PR #1584, opened 2019-09-30
Update MangaAe.kt

ellpatron:
We get http error 404 when open chapters page


arkon:
Update `build.gradle` too.

PR #1585, opened 2019-09-30
Add vgperson extension

ObserverOfTime:
This PR adds an extension for [vgperson's manga viewer](http://vgperson.com/other/mangaviewer.php). The site doesn't show manga covers so I substituted them with covers from MU, but Tachiyomi isn't loading them. Any idea how I can fix that?


as280093:
Use this Icons 
[ic_launcher.zip](https://github.com/inorichi/tachiyomi-extensions/files/3674818/ic_launcher.1.zip)

Issue #1267, opened 2019-07-06
mangahub.io


princess15mm:
(CATALOGUE) SOURCE REQUESTS:
  Please add https://mangahub.io/ as an extension!

as280093:
Here the icons
[mangahub.zip](https://github.com/inorichi/tachiyomi-extensions/files/3614402/mangahub.zip)

PR #1593, opened 2019-10-01
Add Mangahub.io extension

Weasy666:
My first time using Kotlin, so if there is something i should change to be more ideomatic, just tell me.
If there is anything else about my implementation, just tell me, i tried to orientate myself with the Mangaeden and Mangahere extensions.
This addresses #1267.

And here the MangaHub icon as svg
[mangahub.zip](https://github.com/inorichi/tachiyomi-extensions/files/3678623/mangahub.zip)
that i created and used, before i remembered the icons in the issue 😅


as280093:
Your missing web_hi_res_512.png in res folder

Weasy666:
Hm...after taking a look again...something seems to have gone horribly wrong with the pr

Weasy666:
Now this looks better! Correct content in kotlin file and the missing png is also there.

src/en/mangahub/build.gradle
```diff
@@ -0,0 +1,18 @@
...
+    pkgNameSuffix = 'en.mangahub'
+    extClass = '.Mangahub'
+    extVersionCode = 1
+    libVersion = '1.2'
+}
+
+dependencies {
+    compileOnly project(':duktape-stub')
+    sourceCompatibility = "1.6"
+    targetCompatibility = "1.6"
```
arkon: Why?

src/en/mangahub/src/eu/kanade/tachiyomi/extension/en/mangahub/Mangahub.kt
```diff
@@ -0,0 +1,272 @@
...
+import org.jsoup.nodes.Element
+import java.net.URL
+import java.text.ParseException
+import java.text.SimpleDateFormat
+import java.util.*
+import kotlin.collections.ArrayList
+
+class Mangahub : ParsedHttpSource() {
+
+    override val id: Long = 2
```
arkon: Remove this.

src/en/mangahub/src/eu/kanade/tachiyomi/extension/en/mangahub/Mangahub.kt
```diff
@@ -0,0 +1,272 @@
...
+
+    override val baseUrl = "https://www.mangahub.io"
+
+    override val lang = "en"
+
+    override val supportsLatest = true
+
+    override fun popularMangaSelector() = "#mangalist div.media-manga.media"
+
+    override fun latestUpdatesSelector() = "#mangalist div.media-manga.media"
```
arkon: Could just call `popularMangaSelector`

src/en/mangahub/src/eu/kanade/tachiyomi/extension/en/mangahub/Mangahub.kt
```diff
@@ -0,0 +1,272 @@
...
+
+    override fun popularMangaNextPageSelector() = "ul.pager li.next > a"
+
+    override fun latestUpdatesNextPageSelector() = popularMangaNextPageSelector()
+
+    override fun searchMangaNextPageSelector() = popularMangaNextPageSelector()
+
+    override fun mangaDetailsParse(document: Document): SManga {
+        val manga = SManga.create()
+        manga.title = document.select("h1._3xnDj").first().text()
```
arkon: I'm wary of the classes like `_3xnDj` -- they seem autogenerated and prone to changing.

src/en/mangahub/build.gradle
Weasy666: Guess this is some leftover. Deleted this.

src/en/mangahub/src/eu/kanade/tachiyomi/extension/en/mangahub/Mangahub.kt
Weasy666: Done.

src/en/mangahub/src/eu/kanade/tachiyomi/extension/en/mangahub/Mangahub.kt
Weasy666: Done.

src/en/mangahub/src/eu/kanade/tachiyomi/extension/en/mangahub/Mangahub.kt
Weasy666: Yeah...it is a React webpage (i think), so had to use this heavily. If it is ok with you, let's try with this classes and if they change really often i will think about something else i can do.

SnakeDoc83: You could do `val infoElement = document.select("div.row").first()`

Then `manga.title = infoElement.select("h1").text()` and the rest as well by selecting from `infoElement` instead of `document`.

Weasy666: I thought about that at first and went with the classes, because when the classes change i will just end up with an empty manga or an empty list of pages and so on. But when the layout of the page changes and i'm just using the tags, i could end up selecting something i don't want and potentially break the application. Maybe i'm thinking to much ¯\_(ツ)_/¯

PR #1615, opened 2019-10-04
add otakufile

arifr007:


src/id/otakufile/src/eu/kanade/tachiyomi/extension/id/otakufile/OtakuFile.kt
```diff
@@ -0,0 +1,134 @@
...
+
+    override fun popularMangaRequest(page: Int): Request {
+        return GET("$baseUrl/komik-populer", headers)
+    }
+
+    override fun latestUpdatesRequest(page: Int): Request {
+        return GET("$baseUrl/", headers)
+    }
+
+    override fun searchMangaRequest(page: Int, query: String, filters: FilterList): Request = latestUpdatesRequest(page)
```
arkon: You can probably make a request to the endpoint that the site uses for that dropdown via AJAX.

arifr007: @arkon ill add search feature in next update

arifr007: > You can probably make a request to the endpoint that the site uses for that dropdown via AJAX.

can you give me example request using ajax? i haven't use ajax in kotlin before ._.

arkon: I mean the website hits an endpoint via AJAX. It'd just be another regular request in the extension.

Issue #1578, opened 2019-09-29
[Source Request][English] Boommanga.com


KuroViLolitannia:
https://www.boommanga.com/

Has some webtoon not available in other sites

waptik:
Do you have the url where mangas can be viewed on the web or an api url?

KuroViLolitannia:
Yes, it's from here https://m.boommanga.com/category

You can see their website by using the mobile version too the https://m.boommanga.com/

PR #1626, opened 2019-10-09
Source: BoomManga

happywillow0:
A start to request #1578

Added:
Source Added for BoomManga
Supports Latest
Supports Most Popular
Supports Search

Missing: 
Icons


happywillow0:
I knew this looked similar to another extension. I didn't know which one at first but it is using the same base as manmanga.

happywillow0:
Found Three that look to be running on a similar platform. 
BoomManga 
ManManga
TwinsComics

Someone can make a factory for the three. 
Boommanga however does use some different CSS which my code is based on.

happywillow0:
Made a factory instead of sleeping. 😪

as280093:
[ic_launcher.zip](https://github.com/inorichi/tachiyomi-extensions/files/3716700/ic_launcher.zip)
If it's not a factory here the icon

PR #1629, opened 2019-10-10
[New] BoomManga Factory

### LINKED ISSUE #1578: [Source Request][English] Boommanga.com

happywillow0:
Adds new source factory based off of an app platform from what I'm guessing to be China
Willing to rename to what makes sense. 
Willing to change to individual extensions instead of factory. 

Supports both EN and ZH

Source list:
BoomManga - closes #1578 
ManManga - deprecates an older extension
TwinsComics - new


as280093:
Are you keeping the same name or any name suggestion
[ic_launcher.zip](https://github.com/inorichi/tachiyomi-extensions/files/3716735/ic_launcher.zip)

happywillow0:
Thanks for the icons!

I'll take any name suggestion. The build currently has it as BoomManga, ManManga, TwinsComics similar to HolyManga & HeavenManga. 

<details><summary>I tried to find out more info about the websites but I couldn't find too much. 
</summary>All three domains registered with GoDaddy point to the same IP at Amazon. While all three have each individual SSL certs from Letsencrypt, the cert that comes up from the  “private manage” section is configured and and all three sites point to the *.boommanga.com certificate. </details> That is why I kept the same name.

joseph619:
Holy and Heavenmanga are exactly the same source with just different domain. Is it the same for BoomManga and ManManga? If not you should name the factory extension by their layout name (like Madara, Foolslide etc).

happywillow0:
You know, I never actually opened Holy or Heaven to actually look at the websites. 
Man, Boom, and Twin all use the same layout, but I can't tell if they are run by the same people or not. From what I can tell, they do have slightly different catalogs. 

> you should name the factory extension by their layout name

The problem is that I can't find what the layout is actually called.

joseph619:
If they have different catalaogues then you should use the layout name for factory extension, otherwise it would be misleading (since Manmanga is not a clone of Boommanga). You can ask Snakedoc or Arkon for layout name first.

happywillow0:
Thanks for your insight. I'll update this when I get more info.

arkon:
I don't really have any suggestions for a name. ¯\_(ツ)_/¯

joseph619:
@happywillow0  Manmanga group source shows less number of manga than Manmanga extension

Issue #1592, opened 2019-10-01
KOMGA SLOWING DOWN CATALOGUE SEARCH


shinh28:
Just like what the title says, komga slowed down catalogue searching for me. When I still had it installed searching would take longer than usual, but once it was gone, the search became fast again just like the way it did. Just want y'all to know. I'm using the latest version of tachiyomi.

arkon:
Was your Komga instance running at the time?

gotson:
@shinh28 if you have more details on this issue i am interested, being the developer of Komga and its Tachiyomi extension.

- Did you had a Komga server configured in the extension ?
- Was the server running at the time ?
- Did you try searching in Komga only, and not from the catalogue view ?

If no server is setup or if the server is unreachable, the search method for Komga in Tachiyomi will throw an exception straight away.

I don't know exactly how the catalogue search is done, it may be slow because of the exception.

I tried it on my side, with an unreachable and reachable Komga server, and the Komga search is always the fastest, compared to other extensions.

SnakeDoc83:
@arkon This can be closed.  Tachi handles global search better in .8.5; so if Komga really was slowing down search in .8.4 it shouldn't be as much of a hindrance now.  Plus, gotson's made some changes since then to the extension.

PR #1638, opened 2019-10-12
[Komga] add filters by library

gotson:
Libraries depend on each server, and can be dynamic. Retrieving once at startup seems the best option, but i am happy to get feedback in case there is a better option!

### DIFF
--- a/src/all/komga/src/eu/kanade/tachiyomi/extension/all/komga/Komga.kt
+++ b/src/all/komga/src/eu/kanade/tachiyomi/extension/all/komga/Komga.kt
@@ -7,15 +7,15 @@ import android.support.v7.preference.PreferenceScreen
 import android.widget.Toast
 import com.github.salomonbrys.kotson.fromJson
 import com.google.gson.Gson
-import eu.kanade.tachiyomi.extension.all.komga.dto.BookDto
-import eu.kanade.tachiyomi.extension.all.komga.dto.PageDto
-import eu.kanade.tachiyomi.extension.all.komga.dto.PageWrapperDto
-import eu.kanade.tachiyomi.extension.all.komga.dto.SerieDto
+import eu.kanade.tachiyomi.extension.all.komga.dto.*
 import eu.kanade.tachiyomi.network.GET
 import eu.kanade.tachiyomi.source.ConfigurableSource
 import eu.kanade.tachiyomi.source.model.*
 import eu.kanade.tachiyomi.source.online.HttpSource
 import okhttp3.*
+import rx.Single
+import rx.android.schedulers.AndroidSchedulers
+import rx.schedulers.Schedulers
 import uy.kohesive.injekt.Injekt
 import uy.kohesive.injekt.api.get
 import java.text.SimpleDateFormat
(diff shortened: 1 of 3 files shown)


gotson:
Hi @arkon I was wondering if you could have a look at this PR? Thanks!

src/all/komga/src/eu/kanade/tachiyomi/extension/all/komga/Komga.kt
```diff
@@ -119,6 +138,33 @@ open class Komga : ConfigurableSource, HttpSource() {
...
+    private var libraries = emptyList<LibraryDto>()
+
+
+    init {
+        Single.fromCallable {
+            client.newCall(GET("$baseUrl/api/v1/libraries", headers)).execute()
+        }
+            .subscribeOn(Schedulers.io())
+            .observeOn(AndroidSchedulers.mainThread())
+            .subscribe({
```
arkon: Can't you do something like `.asObservableSuccess()`?

src/all/komga/src/eu/kanade/tachiyomi/extension/all/komga/Komga.kt
```diff
@@ -132,6 +178,7 @@ open class Komga : ConfigurableSource, HttpSource() {
     override fun headersBuilder(): Headers.Builder =
         Headers.Builder()
             .add("Authorization", Credentials.basic(username, password))
+            .add("User-Agent", "Tachiyomi Komga v3")
```
arkon: You could probably just get the version from `Build.Config`

src/all/komga/src/eu/kanade/tachiyomi/extension/all/komga/Komga.kt
gotson: I tried that, but i can't get it to subscribe and observe on the correct thread, so it fails at initialisation and the extension doesn't load properly :(

src/all/komga/src/eu/kanade/tachiyomi/extension/all/komga/Komga.kt
gotson: Indeed, thanks for the tip!

gotson: fixed in https://github.com/inorichi/tachiyomi-extensions/pull/1638/commits/b0c56632a6b82542e43bd25ee0899bf4e44962c4

src/all/komga/src/eu/kanade/tachiyomi/extension/all/komga/Komga.kt
arkon: Actually, would it be possible to lazily load all this stuff when needed? Users have noted that Komga affects the app because it may be slow or failing to fetch things at times.

gotson: Yes why not, but I don't know so well the extension lifecycle to do that. I understand `init` will be called at application startup, o indeed that may nto be the best place.

Where else could I call this, as it's needed only in filters, and given libraries don't change so often loading it only once and not every time filters are shown is enough?

I'm also interested in the user's report, as I have about 10 friends using the extension almost daily and none have reported anything, would you be able to share more about it?

arkon: There's #1592 and there was some discussion on Discord at some point, although I admittedly don't pay much attention to Discord.

I actually don't remember if there's proper lifecycle hooks to do this, so I'm cool with leaving it until later, if it's possible.

gotson: Thanks, I'll monitor the cossue you mentioned. Please don't hesitate to mention me on any issues related to Komga, I'll be happy to help!

Issue #1641, opened 2019-10-13
[Source Request] [English] mangakisa


randw0:
### Source Website
The website for this source is https://mangakisa.com/.

happywillow0:
Looks like it could be a easy add. 

I'm not sure it is uses a common platform / layout or not but I didn't think it looked like the others.

PR #1646, opened 2019-10-14
[New] MangaKisa

happywillow0:
New Source for #1641 

Supports

- Popular (This Week or All Time set via preferences)
- Latest (Popular Updates or All Updates)
- Search
- Genres

Missing

- ~~Icons~~ Thanks as always @as280093 

Known bugs:

- Certain chapters with one specific domain is failing to render (rare)
- ~~Searching generates an extra entry at the end and I'm not sure how the css selector is generating it.~~ Thanks @SnakeDoc83


SnakeDoc83:
Use `"div.iepbox a.an"` for `searchMangaSelector`

as280093:
Here is the Icon
[ic_launcher.zip](https://github.com/inorichi/tachiyomi-extensions/files/3724158/ic_launcher.2.zip)

PR #1654, opened 2019-10-15
Add kotlinx.serialization to MangaPlus

alessandrojean:
Currently, MangaPlus is a little slow because it uses Duktape to deserialize Protobuf and decode the pages of the chapter. This PR removes Duktape completely by replacing it with [kotlinx.serialization](https://github.com/Kotlin/kotlinx.serialization) and a native function to decode the images. Should fix the problem with the thumbnails not loading after some time too.

I had to add one `classpath` to the repository Gradle file, but I think this is not a problem.

**Edit:** The compile log will have some warnings about `@Optional`, but I couldn't make it work without it, since using a default value to make a property optional only works with Kotlin >= 1.3.30.

### DIFF
--- a/src/all/mangaplus/build.gradle
+++ b/src/all/mangaplus/build.gradle
@@ -1,11 +1,12 @@
 apply plugin: 'com.android.application'
 apply plugin: 'kotlin-android'
+apply plugin: 'kotlinx-serialization'
 
 ext {
     appName = 'Tachiyomi: MANGA Plus by SHUEISHA'
     pkgNameSuffix = 'all.mangaplus'
     extClass = '.MangaPlusFactory'
-    extVersionCode = 1
+    extVersionCode = 2
     libVersion = '1.2'
 }
 
@@ -13,6 +14,7 @@ dependencies {
     compileOnly 'com.google.code.gson:gson:2.8.2'
     compileOnly 'com.github.salomonbrys.kotson:kotson:2.5.0'
     compileOnly project(':duktape-stub')
+    implementation 'org.jetbrains.kotlinx:kotlinx-serialization-runtime:0.13.0'
 }
 
 apply from: "$rootDir/common.gradle"
--- a/src/all/mangaplus/build.gradle
+++ b/src/all/mangaplus/build.gradle
@@ -11,9 +11,6 @@ ext {
 }
 
 dependencies {
-    compileOnly 'com.google.code.gson:gson:2.8.2'
-    compileOnly 'com.github.salomonbrys.kotson:kotson:2.5.0'
-    compileOnly project(':duktape-stub')
     implementation 'org.jetbrains.kotlinx:kotlinx-serialization-runtime:0.13.0'
 }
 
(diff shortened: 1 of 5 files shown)


src/all/mangaplus/build.gradle
```diff
@@ -1,18 +1,17 @@
...
-    extVersionCode = 1
+    extVersionCode = 2
     libVersion = '1.2'
 }
 
 dependencies {
-    compileOnly 'com.google.code.gson:gson:2.8.2'
-    compileOnly 'com.github.salomonbrys.kotson:kotson:2.5.0'
-    compileOnly project(':duktape-stub')
+    implementation 'org.jetbrains.kotlinx:kotlinx-serialization-runtime:0.13.0'
```
CarlosEsco: I believe this needs to be compileOnly and the implementation dependency added to Tachiyomi's gradle.  @arkon  thoughts?

alessandrojean: It will take a while to the PR on the other repo be merged, the new stable version be launched and the users update the app. Isn't it better to launch this version with `implementation` and, after this process finishes, release a new version with `compileOnly`?

CarlosEsco: @inorichi  thoughts?

arkon: There's issues with how backwards compatibility between the app and extensions is supposed to work that I haven't really thought through. I suspect the older way of handling the version codes was meant to deal with that, but I don't really remember...

src/all/mangaplus/src/eu/kanade/tachiyomi/extension/all/mangaplus/MangaPlusApi.kt
```diff
@@ -0,0 +1,117 @@
...
+
+@Serializable
+data class Popup(
+    @SerialId(1) val subject: String,
+    @SerialId(2) val body: String
+)
+
+@Serializable
+data class SuccessResult(
+    @Optional @SerialId(1) val isFeaturedUpdated: Boolean? = false,
```
CarlosEsco: With this version of kotlinx i am getting a warning that its deprecated to use @Optional

alessandrojean: It gives a warning, yes. But I couldn't make it work without it, since to make the property optional without the `@Optional` requires a newer Kotlin version, as the kotlinx.serialization repo's says:

> Property will be considered optional if it has default value (kotlin 1.3.30 or higher is required).

Since the repo uses Kotlin 1.3.29, I did not update it as it potentially could cause some problems as said on #1047.

src/all/mangaplus/build.gradle
CarlosEsco: @alessandrojean  have you built the extension as apk and tried installing it?  If that works i dont mind merging this, and you add a pull request to tachiyomi to add kotlinx.

alessandrojean: @CarlosEsco I tried now, and it built the apk with success (generating a ~1mb one, however). It installed and worked as expected.

CarlosEsco: @inorichi  are you fine with this being merged until kotlinx is added to 0.x stable?

inorichi: I think it should be fine. I'll probably merge the Kotlin upgrade on the main app, then we could add the serialization lib there.

PR #1680, opened 2019-10-21
[Mangadex] Remove eu/eu2 image server options and update strings to match current website options

FlaminSarge:
Confirmed by MD staff

```
[4:01 PM] FlaminSarge: quick q: the 'eu' and 'eu2' server query parameter options are completely gone and we should use 'na' and 'na2' only now, right?
[4:01 PM] FlaminSarge: or 'row'
[4:07 PM] Plykiya ⭐: @FlaminSarge eu isn't gone
[4:07 PM] Plykiya ⭐: the na and eu options are merged because Path automatically directs your traffic to their closest cache server which is either in the NA or EU
[4:37 PM] FlaminSarge: Sorry, right, the query param values only accept 'na' or 'na2' or 'row' now though, right?
[4:38 PM] FlaminSarge: last I checked, using 'eu' or 'eu2' values didn't work
[4:45 PM] Holo: yesw
[4:53 PM] FlaminSarge: ty
```

### DIFF
--- a/src/all/mangadex/build.gradle
+++ b/src/all/mangadex/build.gradle
@@ -5,7 +5,7 @@ ext {
     appName = 'Tachiyomi: MangaDex'
     pkgNameSuffix = 'all.mangadex'
     extClass = '.MangadexFactory'
-    extVersionCode = 69
+    extVersionCode = 70
     libVersion = '1.2'
 }
 
--- a/src/all/mangadex/src/eu/kanade/tachiyomi/extension/all/mangadex/Mangadex.kt
+++ b/src/all/mangadex/src/eu/kanade/tachiyomi/extension/all/mangadex/Mangadex.kt
@@ -584,8 +584,8 @@ abstract class Mangadex(
         val serverPref = ListPreference(screen.context).apply {
             key = SERVER_PREF_Title
             title = SERVER_PREF_Title
-            entries = arrayOf("Auto", "North America", "North America 2", "Europe", "Europe 2", "Rest of the World")
-            entryValues = arrayOf("0", "na", "na2", "eu", "eu2", "row")
+            entries = SERVER_PREF_ENTRIES
+            entryValues = SERVER_PREF_ENTRY_VALUES
             summary = "%s"
 
             setOnPreferenceChangeListener { _, newValue ->
@@ -603,7 +603,11 @@ abstract class Mangadex(
 
     private fun getShowR18(): Int = preferences.getInt(SHOW_R18_PREF, 0)
     private fun getShowThumbnail(): Int = preferences.getInt(SHOW_THUMBNAIL_PREF, 0)
-    private fun getServer(): String = preferences.getString(SERVER_PREF, "0")
+    private fun getServer(): String {
+        val default = SERVER_PREF_ENTRY_VALUES.first()
+        return preferences.getString(SERVER_PREF, default).takeIf { it in SERVER_PREF_ENTRY_VALUES }
+            ?: default
+    }
 
     private class TextField(name: String, val key: String) : Filter.Text(name)
     private class Tag(val id: String, name: String) : Filter.TriState(name)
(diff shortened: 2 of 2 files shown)


src/all/mangadex/src/eu/kanade/tachiyomi/extension/all/mangadex/Mangadex.kt
```diff
@@ -599,7 +599,14 @@ open class Mangadex(override val lang: String, private val internalLang: String,
 
     private fun getShowR18(): Int = preferences.getInt(SHOW_R18_PREF, 0)
     private fun getShowThumbnail(): Int = preferences.getInt(SHOW_THUMBNAIL_PREF, 0)
-    private fun getServer(): String = preferences.getString(SERVER_PREF, "0")
+    private fun getServer(): String {
+        var server = preferences.getString(SERVER_PREF, SERVER_PREF_ENTRY_VALUES.first())
+        if (server !in SERVER_PREF_ENTRY_VALUES) {
+            server = SERVER_PREF_ENTRY_VALUES.first()
+            preferences.edit().putString(SERVER_PREF, server).commit()
+        }
```
FlaminSarge: Unsure whether I should be overriding+storing the value here or merely ignoring it (skipping the preferences.edit()), or even removing the SERVER_PREF value entirely with .remove(). Particularly not sure if the preference screen's selection updates in tandem with preferences.edit() calls.

src/all/mangadex/src/eu/kanade/tachiyomi/extension/all/mangadex/Mangadex.kt
```diff
@@ -580,8 +580,8 @@ open class Mangadex(override val lang: String, private val internalLang: String,
         val serverPref = ListPreference(screen.context).apply {
             key = SERVER_PREF_Title
             title = SERVER_PREF_Title
-            entries = arrayOf("Auto", "North America", "North America 2", "Europe", "Europe 2", "Rest of the World")
-            entryValues = arrayOf("0", "na", "na2", "eu", "eu2", "row")
+            entries = SERVER_PREF_ENTRIES
+            entryValues = SERVER_PREF_ENTRY_VALUES
             summary = "%s"
 
             setOnPreferenceChangeListener { _, newValue ->
```
FlaminSarge: Trying to figure out if this can just use 'selected' instead of getting index and then pulling value from the entryValues array again.

src/all/mangadex/build.gradle
```diff
@@ -5,7 +5,7 @@ ext {
     appName = 'Tachiyomi: MangaDex'
     pkgNameSuffix = 'all.mangadex'
     extClass = '.MangadexFactory'
-    extVersionCode = 68
+    extVersionCode = 69
```
arkon: Nice.

src/all/mangadex/src/eu/kanade/tachiyomi/extension/all/mangadex/Mangadex.kt
arkon: Did you ever manage to check if you could?

src/all/mangadex/src/eu/kanade/tachiyomi/extension/all/mangadex/Mangadex.kt
arkon: You could alternatively just switch to an entirely new preference key and notify people to update it if needed.

FlaminSarge: I'm going to update this to ignore the server pref value if the value is invalid, rather than reset it; in the event that eu/eu2 comes back, users with old values will just already have it set, while still ignoring any invalid values that may come along. That seems like it involves the least friction and least 'overriding what a user has selected'.

src/all/mangadex/src/eu/kanade/tachiyomi/extension/all/mangadex/Mangadex.kt
FlaminSarge: I'm going to leave this alone for now and try to check its behavior after this is pushed (I can't seem to get dev extensions to be picked up by the debug app). If anything, stricter safety around entryValues[index] will allow us to set it back to "0" if an invalid value comes in, though I'm not sure how Kotlin handles invalid indices e.g. -1.

src/all/mangadex/build.gradle
arkon: I think you actually need to merge the latest changes into your branch, this should be `70` now.

FlaminSarge: Will rebase and update.

FlaminSarge: (Why'd you have to go and take that version?)

PR #1696, opened 2019-10-23
[Enhancement] TuMangaOnline De-dup Preference

happywillow0:
Addresses https://github.com/inorichi/tachiyomi/issues/2297

New:
Provides preference option to remove duplicate chapter entries by only adding first scanlator on list. 

To be fixed later:
Strings are in English, someone should translate them properly.

### DIFF
--- a/src/es/tumangaonline/build.gradle
+++ b/src/es/tumangaonline/build.gradle
@@ -5,12 +5,14 @@ ext {
     appName = 'Tachiyomi: TuMangaOnline'
     pkgNameSuffix = 'es.tumangaonline'
     extClass = '.TuMangaOnline'
-    extVersionCode = 10
+    extVersionCode = 11
     libVersion = '1.2'
 }
 
 dependencies {
     implementation project(':lib-ratelimit')
+    compileOnly project(':preference-stub')
+    compileOnly 'com.github.inorichi.injekt:injekt-core:65b0440'
 }
 
 apply from: "$rootDir/common.gradle"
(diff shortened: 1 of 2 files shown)


arkon:
What if you don't want the first scanlator? I feel like this needs to be addressed at the app level.

MerradOklom:
This is just an idea, but how about blocking scans by name like MangaDex does now. For example in preferences shows "write scan name to exclude/hide", click and text field pop up and user must input the scan name, or something similar. I'm not a developer and I know this must have some degree of complications, but if it is feasible that would be nice.

happywillow0:
In theory, there are a lot of scanners for a lot of series so "blocking" scans per group is a bit hard.

This should replicate basically what the extension was doing before I split the scanners up. Users didn't have a choice when it existed, nor did the some users really care which scan they were reading.  

I'm looking at the previous code, what I pretty sure was happening was url `attr("href"))` grabbed the first <a> link while the scanlator grabbed both names at once. 
```
    override fun chapterListParse(response: Response): List<SChapter> {
        (...)
        return document.select(regularChapterListSelector()).map { regularChapterFromElement(it) }
    }
    private fun regularChapterListSelector() = "div.chapters > ul.list-group li.p-0.list-group-item"
    private fun regularChapterFromElement(element: Element) = SChapter.create().apply {
        setUrlWithoutDomain(element.select("div.row > .text-right > a").attr("href"))
        name = element.select("div.col-10.text-truncate").text()
        scanlator = element.select("div.col-md-6.text-truncate")?.text()
        date_upload = element.select("span.badge.badge-primary.p-2").first()?.text()?.let { parseChapterDate(it) } ?: 0
    }
```

But, we can also just say deal with it as we did in the issue we skip / delete this change.

MerradOklom:
Yup. Personally I don't feel like I'll use this but as you said some people don't care about which scan is better and are too lazy to do what I explained in the issue.

arkon:
I'll merge this anyway since it's configurable anyway.

PR #1752, opened 2019-11-06
[Rebuilt] Mangakawaii

happywillow0:
No open issue on GitHub but user reported error 503 on Discord. 
Added headers but testing found no results. 
Code was based on a forked copy of a MMRCMS

Instead of trying to follow old code and patching issues, I thought it was easier to just re-write the extension from the ground up.


arkon:
Would it make sense to focus on updating the MMRCMS code later and putting this in there too?

SnakeDoc83:
@arkon I think the only issue with MMRCMS is that when the generator script runs it gets boned by sites using Cloudflare.

PR #1770, opened 2019-11-12
Mangahub - parse images from JSON

SnakeDoc83:
Closes https://github.com/inorichi/tachiyomi-extensions/issues/1766

### DIFF
--- a/src/en/mangahub/src/eu/kanade/tachiyomi/extension/en/mangahub/Mangahub.kt
+++ b/src/en/mangahub/src/eu/kanade/tachiyomi/extension/en/mangahub/Mangahub.kt
@@ -1,15 +1,25 @@
 package eu.kanade.tachiyomi.extension.en.mangahub
 
+import com.github.salomonbrys.kotson.fromJson
+import com.github.salomonbrys.kotson.get
+import com.github.salomonbrys.kotson.string
+import com.google.gson.Gson
+import com.google.gson.JsonObject
 import eu.kanade.tachiyomi.network.GET
+import eu.kanade.tachiyomi.network.POST
 import eu.kanade.tachiyomi.source.model.*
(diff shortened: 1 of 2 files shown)


src/en/mangahub/src/eu/kanade/tachiyomi/extension/en/mangahub/Mangahub.kt
```diff
@@ -128,23 +135,33 @@ class Mangahub : ParsedHttpSource() {
...
+    override fun pageListRequest(chapter: SChapter): Request {
+        val jsonHeaders = headers.newBuilder().add("Content-Type", "application/json").build()
 
-        val pages = document.select("div#mangareader img.PB0mN")
-        val pageUrl = pages.first().attr("src")
-        val pageRoot = pageUrl.replaceAfterLast("/", "")
-        val numPages = pages.first().nextElementSibling().text().split("/").last().toInt()
+        val slug = chapter.url.substringAfter("chapter/").substringBefore("/")
+        val number = chapter.url.substringAfter("chapter-").removeSuffix("/")
+        val body = RequestBody.create(null, "{\"query\":\"{chapter(x:m01,slug:\\\"$slug\\\",number:$number){id,title,mangaID,number,slug,date,pages,noAd,manga{id,title,slug,mainSlug,author,isWebtoon,isYaoi,isPorn,isSoftPorn,unauthFile,isLicensed}}}\"}")
```
arkon: You probably don't need to query for everything here

SnakeDoc83: That's what the request looked like in Firefox.  It's been my experience in the past that altering a body that looks like that messes up the request; so I went with what I saw.

Tried altering the request just now and I get a 400 error.  I think it needs to be as is unless you know something I don't.

PR #1789, opened 2019-11-16
MMRCMS update

SnakeDoc83:
Added minimal ability for the generator to bypass Cloudflare along with a little cleanup.

Added Utsukushii as requested in https://github.com/inorichi/tachiyomi-extensions/issues/1690
Added Tumangaonline.co, which closes https://github.com/inorichi/tachiyomi-extensions/issues/1535 (The requested site is forwarding to this)

Zxcomic has been down since at least Oct 3, can be considered dead.
SOS Scanlation was down when I ran the generator, don't know if it's temporary or not.  Should probably check again in a day or so to see if it comes back.
Mangatreat should be moved to Madara.

Tried adding Mangakawaii and it's not a clean fit, not worth trying to make that happen at this time.


happywillow0:
The old comments for Mangakawaii said that was why it was forked off into it's own extension anyways. (classes names, custom endpoints, etc)

PR #1795, opened 2019-11-18
[FIx] Allporncomic

### LINKED ISSUE #1769: Extension under Madara not working
### Device information
* Tachiyomi extension version: 1.2.26
* Tachiyomi version: 0.8.4
* Android version: 7.0 (NRD90M XIAOMI REDMI NOTE 4X)

## Steps to reproduce
Opening the "Allporncomic" catalogue results in error even after force closing and opening the app. I've also tried restarting the phone before using the app. Although the individual comics of the aforementioned site is accessible from its page via webpage view NOT the catalogue "browse" or "latest" option.

### Expected behavior
Selecting the catalogue/extension should open all the comics of "allporncomic", but it shows error. (However, I can view individual bookmarked comics of that extension via web view)

### Actual behavior
Error message pops up: 
"java.lang.Exception:  Challenge not found"

happywillow0:
Fixes #1769

Allporncomic might still be using cloudflare but most of the major browser checking and ddos protections look to be turned off. 

Not sure why this edit works but it works on my test phone. I not even adding any headers.

### DIFF
--- a/src/all/madara/build.gradle
+++ b/src/all/madara/build.gradle
@@ -5,7 +5,7 @@ ext {
     appName = 'Tachiyomi: Madara (multiple sources)'
     pkgNameSuffix = "all.madara"
     extClass = '.MadaraFactory'
-    extVersionCode = 26
+    extVersionCode = 27
     libVersion = '1.2'
 }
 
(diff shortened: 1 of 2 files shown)


fam007e:
Unfortunately, @happywillow0 the problem which seemed to have been solved after new version installation is still showing now. 😭

PR #1796, opened 2019-11-18
[Fix] My Reading Manga - fix for no cover

happywillow0:
Reported on discord although I see it sometimes myself. 

Fixes a null error thrown on the off chance a search result doesn't have a cover.

### DIFF
--- a/src/all/myreadingmanga/src/eu/kanade/tachiyomi/extension/all/myreadingmanga/MyReadingManga.kt
+++ b/src/all/myreadingmanga/src/eu/kanade/tachiyomi/extension/all/myreadingmanga/MyReadingManga.kt
@@ -131,7 +131,7 @@ open class MyReadingManga(override val lang: String) : ParsedHttpSource() {
         val manga = SManga.create()
         manga.setUrlWithoutDomain(titleElement.attr("href"))
         manga.title = cleanTitle(titleElement.text())
-        if (thumbnailElement !=null) manga.thumbnail_url = getThumbnail(getImage(thumbnailElement)) // else manga.thumbnail_url="none"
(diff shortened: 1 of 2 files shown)


src/all/myreadingmanga/src/eu/kanade/tachiyomi/extension/all/myreadingmanga/MyReadingManga.kt
```diff
@@ -125,13 +125,13 @@ open class MyReadingManga(override val lang: String) : ParsedHttpSource() {
...
-    override fun searchMangaFromElement(element: Element) = buildManga(element.select("a").first(), element.select("img").first())
+    override fun searchMangaFromElement(element: Element) = buildManga(element.select("a").first(), element.select("img")?.first() )
 
-    private fun buildManga(titleElement: Element, thumbnailElement: Element): SManga {
+    private fun buildManga(titleElement: Element, thumbnailElement: Element?): SManga {
         val manga = SManga.create()
         manga.setUrlWithoutDomain(titleElement.attr("href"))
         manga.title = cleanTitle(titleElement.text())
-        manga.thumbnail_url = getThumbnail(getImage(thumbnailElement))
+        if (thumbnailElement !=null) manga.thumbnail_url = getThumbnail(getImage(thumbnailElement)) else manga.thumbnail_url="none"
```
arkon: Does `none` actually resolve to something, or should it just be left uninitialized?

happywillow0: Due to an an additional check in the fetchmangadetails to check if there is a cover or not, it was an attempt to quickly bypass the missing cover and the resulting error in the details screen. 

Though really, it should be relatively rare (should be reported and fixed at the source) and it doesn't affect the chapter list fetcher.

PR #1800, opened 2019-11-19
[Fix v3] All Porn Comics

happywillow0:
I really hate this website now and maybe I should just give up on it. 
I'm leaving this as a draft so I can at least wait a day between testing. 

I've been trying a whole bunch of combinations between sending different headers, switching the clients, and even tried adding a rate limiter. 
Is there something different about my unsigned build vs the official build? 
Unless the site is doing something stupid on their end. 

Anyone want to confirm, review, comment, or have any other ideas?


happywillow0:
No, I wanted a draft. Why did the conflict resolution mark it as ready?

PR #1801, opened 2019-11-19
Madara Fixes, - revert AllPornComics, ZinManga fix

happywillow0:
**ZinManga** https://github.com/inorichi/tachiyomi/issues/2335
Added Referrer to resolve 403 forbidden from images hosted by CDN

**AllPornComics** https://github.com/inorichi/tachiyomi-extensions/issues/1769
503 errors should have come from WordFence. However, due to also running though the cloudflare cdn, it triggered Tachiyomi's cloudflare bypass. Wordfence seems to be behaving now so, I reverted the changes. 

Draft pending continued testing against AllPornComics to ensure the fix actually sticks.

### DIFF
--- a/src/all/madara/build.gradle
+++ b/src/all/madara/build.gradle
@@ -5,7 +5,7 @@ ext {
     appName = 'Tachiyomi: Madara (multiple sources)'
     pkgNameSuffix = "all.madara"
     extClass = '.MadaraFactory'
-    extVersionCode = 27
+    extVersionCode = 28
     libVersion = '1.2'
 }
 
--- a/src/all/madara/build.gradle
+++ b/src/all/madara/build.gradle
@@ -5,7 +5,7 @@ ext {
     appName = 'Tachiyomi: Madara (multiple sources)'
     pkgNameSuffix = "all.madara"
     extClass = '.MadaraFactory'
-    extVersionCode = 28
+    extVersionCode = 29
     libVersion = '1.2'
 }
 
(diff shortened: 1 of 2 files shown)


fam007e:
Fixes are not working. I mean initial use after an update works but stops after restarting the app and persists. However, some "info" pages still able to update the metadata of the following comic but error message pops whenever you try to open the chapter. In webview it shows wordfence protection with error 503. Is there a way to bypass that... Could the use of "webclient" in place of "weburlconnect" in the http line work..
It's really weird about "allporncomic" WordPress site. **feeling confused**

happywillow0:
By chance, did wordfence say why it was giving you a 503 such as rate limit or banned?

I'm testing my code and the latest iteration is holding for me.

SnakeDoc83:
@happywillow0 I'm still on Madara 1.2.26 (so the version before you tried to fix APC) and APC is now working.  Try reverting your changes and see if it works for you.  Interestingly, on my PC I'm getting a 503 from Wordfence, their page reads

```Your access to this site has been limited

Your access to this service has been temporarily limited. Please try again in a few minutes. (HTTP response code 503)

Reason: Advanced blocking in effect.
```

I tried changing my User-Agent and then I was able to access it with my PC.  Seems like they're messing around with their blocking filters.

happywillow0:
Thanks for the confirmation. I guess this is done then? AllPornComics is reverted because it seems to be working and I think I fixed ZinManga.

fam007e:
Problem again for "Allporncomic" catalogue in Madara extension... 
Wordfence error message shows error 503.
☹️

fam007e:
Madara version 1.2.34 doesn't solve ongoing "allporncomic" issues. It's so frustrating for me. Then it must be much more frustrating for you dev members. I mean what's their problem with wordfence.

PR #1858, opened 2019-12-07
Update README.md

as280093:


arkon:
Should update the issue template(s) instead, I don't think many (non-developer) people read the README.

PR #1876, opened 2019-12-14
Fix readmanga/mintmanga npe on chapter parse

Pavkazzz:


arkon:
Update build.gradle files too.

Pavkazzz:
Forgot. Thanks.

Issue #927, opened 2019-03-17
[Source Request] [Spanish] vercomicsporno


Credits125:
URL:

[vercomicsporno](https://vercomicsporno.com)

Issue #926, opened 2019-03-17
[Source Request] [Spanish] vermangasporno


Credits125:
URL:

[vermangasporno](https://vermangasporno.com)

SnakeDoc83:
This can be closed.

PR #1879, opened 2019-12-15
Added VCP & VMP

edgarMejia:
Closing #926 & #927

Abbreviated names because they are too explicit.


src/es/vcpvmp/build.gradle
```diff
@@ -0,0 +1,12 @@
+apply plugin: 'com.android.application'
+apply plugin: 'kotlin-android'
+
+ext {
+    appName = 'Tachiyomi: VCPVMP'
+    pkgNameSuffix = "es.vcpvmp"
+    extClass = '.VCP; .VMP'
```
arkon: Can just use the factory class.

edgarMejia: Ok, give me a moment

PR #1882, opened 2019-12-16
Toonly Cleanup

happywillow0:
Deleted the source files.

Is there any reason to leave the apks in case anyone is looking for them? 
Am I doing this right?


arkon:
> Am I doing this right?

Yeah.

> Is there any reason to leave the apks in case anyone is looking for them?

Nope, should do another PR to the `repo` branch to remove the APKs and icons.

PR #1893, opened 2019-12-22
Mangakakalot description parsing

Guiorgy:
In manga description:
replace __\<br\>__ tags with __'\n'__ (line break)
remove any other tags (such as __\<b\>__ or __\</b\>__ for example)
some descriptions start with __"[_manga title_] summary:"__, remove it as it adds nothing to the context


arkon:
Update build.gradle

PR #1945, opened 2019-12-31
Cyanide & Happiness - new extension

SnakeDoc83:
Closes https://github.com/inorichi/tachiyomi-extensions/issues/1299


src/en/explosm/src/eu/kanade/tachiyomi/extension/en/explosm/Explosm.kt
```diff
@@ -0,0 +1,104 @@
...
+
+class Explosm : HttpSource() {
+
+    override val name = "Cyanide & Happiness"
+
+    override val baseUrl = "http://explosm.net"
+
+    override val lang = "en"
+
+    override val supportsLatest = false
```
arkon: Could potentially use their RSS feed: http://feeds.feedburner.com/Explosm

PR #1964, opened 2020-01-03
mangabox: fix alt titles included in title

az4521:
fix applies to tachiEH and other forks with updated deps.
I tested it dw.


arkon:
Update the version in `build.gradle`

az4521:
alright, done

PR #1970, opened 2020-01-05
Mangadex: update to cc domain

ZaneHannanAU:
quick fix to move mangadex to the cc domain
https://mangadex.cc/thread/200708

### DIFF
--- a/src/all/mangadex/AndroidManifest.xml
+++ b/src/all/mangadex/AndroidManifest.xml
@@ -11,6 +11,14 @@
                 <category android:name="android.intent.category.BROWSABLE" />
                 <data
                     android:scheme="https"
+                    android:host="mangadex.org"
+                    android:pathPattern="/title/..*" />
+                <data
+                    android:scheme="https"
+                    android:host="mangadex.org"
+                    android:pathPattern="/manga/..*" />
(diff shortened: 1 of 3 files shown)


CarlosEsco:
Not merging. Old chapters won't work until .org returns

ZaneHannanAU:
for the time being, may as well drop a temporary fix:
> temp tachi fix: uninstall old, install via
> - adb: https://cdn.discordapp.com/attachments/413328536389287946/663290240990969886/tachiyomi-all.mangadex-v1.2.74-debug.apk
> - download -> files/open: https://cdn.discordapp.com/attachments/413328536389287946/663290926231453706/tachiyomi-all.mangadex-v1.2.74-release.apk
>
> Then trust in tachi -> extensions -> mangadex
> 
> note that not all chapters are available and won't be until the change comes back
> once mangadex.org comes back immediately switch back to the officially signed application btw 

You should only try it if you trust me + discord + cf wholeheartedly to not mislead you as well

ZaneHannanAU:
Majority of chapters are now loading through tachiyomi-dev, but now opening source config crashes tachi. Unsure if this is an r1582 issue or an issue with this extension.

PR #1981, opened 2020-01-05
fix 45ca32d

Guiorgy:
fixes a mistake (I made) in 45ca32d by not using __.toRegex()__

### DIFF
--- a/src/all/mangabox/build.gradle
+++ b/src/all/mangabox/build.gradle
@@ -5,7 +5,7 @@ ext {
     appName = 'Tachiyomi: MangaBox (Mangakakalot and others)'
     pkgNameSuffix = 'all.mangabox'
     extClass = '.MangaBoxFactory'
-    extVersionCode = 9
+    extVersionCode = 10
     libVersion = '1.2'
 }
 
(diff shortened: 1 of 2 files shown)


arkon:
Whoops, I missed that too.

Can you also update the version number? Could also pull out the regexes as constants.

Guiorgy:
```kotlin
   val HTML_BR_REGEX = """<\s*br\s*\/?>""".toRegex()
   val HTML_ELEMENT_REGEX = "<[^>]*>".toRegex()

    override fun mangaDetailsParse(document: Document): SManga {
        val infoElement = document.select(mangaDetailsMainSelector)
```
__...__
```kotlin
            description = document.select(descriptionSelector)?.first()?.ownText()
                ?.replace("""^$title summary:\s""".toRegex(), "")
                ?.replace(HTML_BR_REGEX , "\n")
                ?.replace(HTML_ELEMENT_REGEX , "")
            thumbnail_url = document.select(thumbnailSelector).attr("abs:src")
        }
    }
```

Will this be fine? I actually have little experience with Kotlin (mainly use Java and C#), so don't exactly know what the equivalent of _private static final_ is

Guiorgy:
before
&nbsp;|
V
after

arkon:
> I actually have little experience with Kotlin (mainly use Java and C#), so don't exactly know what the equivalent of private static final is

It should be in the `companion object { }` block, but that's fine. It's not really a big deal.

PR #1988, opened 2020-01-06
Several https updates

NotTsunami:
All of these sources are now serving their content over https.


SnakeDoc83:
Changing the name might be an issue, that might change the source ID.

NotTsunami:
Alright, I'll leave it out then.

PR #1989, opened 2020-01-06
MangaBox - MangaOnl update

SnakeDoc83:
Closes https://github.com/inorichi/tachiyomi-extensions/issues/1987

### DIFF
--- a/src/all/mangabox/build.gradle
+++ b/src/all/mangabox/build.gradle
@@ -5,7 +5,7 @@ ext {
     appName = 'Tachiyomi: MangaBox (Mangakakalot and others)'
     pkgNameSuffix = 'all.mangabox'
     extClass = '.MangaBoxFactory'
-    extVersionCode = 10
+    extVersionCode = 11
     libVersion = '1.2'
 }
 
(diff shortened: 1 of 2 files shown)


NotTsunami:
Awesome, was just looking at the site myself, but you beat me to it!

SnakeDoc83:
Yeah, they just changed some of their selectors; except they're not site-wide.  So there's a mix of old selectors and new ones, which always seems silly.

happywillow0:
I don't think this isn't actually a change in selectors. It should be the same as kakalot where some manga entries are redirecting to the manganelo url. (I'm not sure why they are doing that).

SnakeDoc83:
No, in this case it's new selectors for some manga.

NotTsunami:
> I don't think this isn't actually a change in selectors. It should be the same as kakalot where some manga entries are redirecting to the manganelo url. (I'm not sure why they are doing that).

Actually, this is right. Search Trinity Seven, click on the first result on the website and it redirects to manganelo.

SnakeDoc83:
Fair enough, different selectors because it's a different site.

PR #1992, opened 2020-01-06
Fix latest ManaMoa extension mistake

DitFranXX:
I didn't pull androidx update.
So it need to increase version number for new update. 

Sorry :P


DitFranXX:
#1994

PR #2004, opened 2020-01-07
Mangabat pagelist Empty

### LINKED ISSUE #2000: [Bug] MangaBat also has some manganelo redirects
The people behind Mangabox are consolidating some mangas back into Manganelo

MangaBat needs some `${super.mangaDetailsMainSelector}` as well. 

Better yet, we might as well add the super to the rest of them just in case.

happywillow0:
Reported via discord 
Fixes #2000

The only examples I could find are Mangabat hosting the manga details page themselves but link to Nelo for the chapter / reader. I don't think there is a need to change anything else. 

Also added a quick comment to help explain what is going on.

### DIFF
--- a/src/all/mangabox/build.gradle
+++ b/src/all/mangabox/build.gradle
@@ -5,7 +5,7 @@ ext {
     appName = 'Tachiyomi: MangaBox (Mangakakalot and others)'
     pkgNameSuffix = 'all.mangabox'
     extClass = '.MangaBoxFactory'
-    extVersionCode = 11
+    extVersionCode = 12
     libVersion = '1.2'
 }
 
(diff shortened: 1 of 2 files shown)


SnakeDoc83:
Do pageListSelector too, pages aren't loading.

happywillow0:
Just to make sure we're on the same page, was there another `pageListSelector` that I missed? 
The PR is for the `pageListSelector` on Mangabat.

I had taken a quick look at the other sources, and they didn't have any connection that I saw with Nelo or at least were not exhibiting this type of behavior as Kakalot / Bat / Onl

SnakeDoc83:
My fault, I saw one change but not the other, should be all good.

PR #2015, opened 2020-01-08
Mangadex: add CC domain to be opened directly.

ZaneHannanAU:
Adds CC to be openable via activities

This fixes an issue with #2013 which was addressed in #1970. 

Examples:
https://mangadex.org/manga/47
https://mangadex.cc/manga/47


ZaneHannanAU:
May as well show the difference:

mangadex.org

mangadex.cc

PR #2021, opened 2020-01-09
MangaLife update

SnakeDoc83:
Still need to do filters.

### DIFF
--- a/src/en/mangalife/src/eu/kanade/tachiyomi/extension/en/mangalife/mangalife.kt
+++ b/src/en/mangalife/src/eu/kanade/tachiyomi/extension/en/mangalife/mangalife.kt
@@ -1,191 +1,225 @@
 package eu.kanade.tachiyomi.extension.en.mangalife
 
-import eu.kanade.tachiyomi.network.POST
+import com.github.salomonbrys.kotson.fromJson
+import com.github.salomonbrys.kotson.get
+import com.github.salomonbrys.kotson.string
+import com.google.gson.GsonBuilder
+import com.google.gson.JsonArray
+import com.google.gson.JsonElement
+import eu.kanade.tachiyomi.network.GET
+import eu.kanade.tachiyomi.network.asObservableSuccess
 import eu.kanade.tachiyomi.source.model.*
-import eu.kanade.tachiyomi.source.online.ParsedHttpSource
-import okhttp3.FormBody
+import eu.kanade.tachiyomi.source.online.HttpSource
+import eu.kanade.tachiyomi.util.asJsoup
 import okhttp3.Headers
-import okhttp3.HttpUrl
+import okhttp3.OkHttpClient
 import okhttp3.Request
-import org.jsoup.nodes.Document
-import org.jsoup.nodes.Element
+import okhttp3.Response
+import rx.Observable
 import java.text.SimpleDateFormat
-import java.util.regex.Pattern
+import java.util.Locale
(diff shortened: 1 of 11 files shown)


happywillow0:
Not that we had called it but if you wanted to take a look at my version which I just finished too:
https://github.com/happywillow0/tachiyomi-extensions/commit/3e54ae4c4acd3355dc810e7860a5595ffd3dc294

SnakeDoc83:
I think my approach towards the catalogs would be a bit more efficient.

`chapterURLEncode()` may be a better way to go than what I've done to build chapter URLs.  At the moment I don't intend to refactor to make it work, but I wouldn't mind if you did after this is committed.  I might steal steal some code for `chapterListParse()`.

`pageListParse()` is kind of a push on which would be better, I think.

As for search, I think the filters from v1 are usable, just need to rewrite the for() loop.

src/en/mangalife/src/eu/kanade/tachiyomi/extension/en/mangalife/mangalife.kt
```diff
@@ -1,206 +1,243 @@
...
+        directory = gson.fromJson<JsonArray>(directoryFromResponse(response))
+            .sortedByDescending { it["v"].string }
+        return parseDirectory(1)
+    }
+
+    private fun parseDirectory(page: Int): MangasPage {
+        val mangas = mutableListOf<SManga>()
+        val endRange = ((page * 24) - 1).let { if (it <= directory.lastIndex) it else directory.lastIndex }
+
+        for (i in (((page - 1) * 23) .. endRange)){
```
happywillow0: I know it is 2am EST / GMT-5 but the math doesn't look right to me here. I double checked with excel but shouldn't the multiplayer be the same? With one x23 and the other x24, the two values diverge and you get duplicate listings if I am thinking straight.

happywillow0: Page | Start | End
-- | -- | --
1 | 0 | 23
2 | 23 | 47
3 | 46 | 71
4 | 69 | 95
5 | 92 | 119
6 | 115 | 143
7 | 138 | 167
8 | 161 | 191
9 | 184 | 215
10 | 207 | 239
11 | 230 | 263
12 | 253 | 287
13 | 276 | 311
14 | 299 | 335
15 | 322 | 359
16 | 345 | 383
17 | 368 | 407
18 | 391 | 431
19 | 414 | 455
20 | 437 | 479
21 | 460 | 503
22 | 483 | 527
23 | 506 | 551
24 | 529 | 575
25 | 552 | 599
26 | 575 | 623
27 | 598 | 647

SnakeDoc83: Yeah, got things mixed up in my head.

SnakeDoc83:
Should be good to go.  As discussed above, although I think my `chapterListParse()` should work fine, if @happywillow0 wants to follow up with a PR to change it, that's okay with me.

arkon:
Closes #2023

happywillow0:
If @SnakeDoc83 is willing, I created a pull request for the refactor into his branch directly so he can pull it in first before this one gets merged.

happywillow0:
Oh cool, that worked! (sense my newness to github?)
I guess we're all set then.

PR #2028, opened 2020-01-09
TMO Updates

happywillow0:
Another day, another update. 
I always feel like it is targeted towards us but I know it is more likely towards Ninemanga and other aggregation sites .

### DIFF
--- a/src/es/tumangaonline/build.gradle
+++ b/src/es/tumangaonline/build.gradle
@@ -5,7 +5,7 @@ ext {
     appName = 'Tachiyomi: TuMangaOnline'
     pkgNameSuffix = 'es.tumangaonline'
     extClass = '.TuMangaOnline'
-    extVersionCode = 16
+    extVersionCode = 17
     libVersion = '1.2'
 }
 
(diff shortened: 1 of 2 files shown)


MerradOklom:
Thank you for your continuous efforts on maintaining this source, but you shouldn't bother fixing it all the time, the users can use LectorManga instead (as long as it works).

happywillow0:
I know I should give up on it but it is sort of fun to solve the challenge they throw. 
Kind of like a daily puzzle. Not that it is good use of time. 

Code updated but currently untested.

happywillow0:
I noticed that they were just changing the /cascade page because scraping the page gets all the images at once. The cascade page is easy to scrape though but it is also easy for them to change currently.  

I switched it to the paginated option which they haven't modified.

happywillow0:
Updated Cascade Code
Added Preferences selector for Cascade vs Paginated

Tested 1/18/20

happywillow0:
I think this is good to review and pull. 
They haven't made any significant changes for a while now. 

The extension is set for the same primary method while there is a back up method built into the preferences in case things change.

PR #2030, opened 2020-01-09
MangaRock - Update Error Message

happywillow0:
Changed P`age List Empty` error to `Manga Rock has shut down. Please migrate to another source` because people still are still asking us in Discord Support 💢


CarlosEsco:
can you add the error to the browse/latest section too?

happywillow0:
Done

I'm at work (not able to pay attention) so I'm editing directly via github. Because, I'm not at home,  nothing is actually tested but the exceptions are in the right place and should work.

CarlosEsco:
I'll merge it now.  Worst case it doesn't work and you can fix it later.

PR #2053, opened 2020-01-13
MerakiScans update

SnakeDoc83:
I think adding `Locale.US` should fix https://github.com/inorichi/tachiyomi-extensions/issues/2052.  Dates were showing up for me already; so I'm guessing that's the issue.  Cleaned up some other stuff while I had it open.


happywillow0:
Not that this isn't the case but doesn't a bad date parse normally spit out an "unparseable date" error?

SnakeDoc83:
Yeah, which is why I'm not 100% sure this will be the fix, but I haven't thought of anything else to try.

happywillow0:
I suddenly remembered. There is a try and catch on the date parser. The error wasn't caught on the other extensions I was thinking about.

Yes, the local will fix it.

PR #2060, opened 2020-01-14
1stKiss - CND Revert

happywillow0:
They went back to the old CND url.

I still have to build / test. 

I need to test if they just need a referrer. Might as well just add it to all the headers and skip the whole cdn check.


happywillow0:
Merged #2066

PR #2063, opened 2020-01-14
Mangadex: Use preferences to specify domain

ZaneHannanAU:
This is future proofing so that if org goes down once more, or if more domains are added, it can be changed (default) or set specifically through the defaults in updates.

- [x] Built
- [x] Tested
- [ ] Ready to merge

### DIFF
--- a/src/all/mangadex/build.gradle
+++ b/src/all/mangadex/build.gradle
@@ -5,7 +5,7 @@ ext {
     appName = 'Tachiyomi: MangaDex'
     pkgNameSuffix = 'all.mangadex'
     extClass = '.MangadexFactory'
-    extVersionCode = 78
+    extVersionCode = 79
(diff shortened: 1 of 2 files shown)


arkon:
Tbh this seems overkill.

ZaneHannanAU:
tbh probably is
but avoids issues later on and less version bumps overall

ZaneHannanAU:
Hrmm. Seems to only show three settings.

ZaneHannanAU:
Whatever. Current one works for now. May be worth ensuring we're back on mangadex.org & mangadex.org

happywillow0:
> Hrmm. Seems to only show three settings.

What Android version are you on? Code got updated to Android X which is a separate code block.

ZaneHannanAU:
Android 9, Google Pixel (original)

PR #2064, opened 2020-01-14
Fix Rawkuma date parsing

### LINKED ISSUE #2061: [Bug] Bug Report - unknown pattern character 'x'
**The extension that have problem:** Rawkuma extension
**Tachiyomi extension version:** 1.2.1
**Tachiyomi version:** 0.8.4
**Android version:** 6.0

**Bug:** Everytime I tap into a manga, the word " unknown pattern character 'x' "appears, and then no chapter can be displayed(completely empty).

I haven't met this bug in any other extensions on this device.
Also, on my Android 8.0 device, Rawkuma extension doesn't have this bug.

Hoping that you can fix it. Thanks a lot :D

happywillow0:
Fixes #2061 

Pattern X is only part of API 24+ (Android 7)
X | Time zone | ISO 8601 time zone | -08; -0800; -08:00 | 24+

I removed the time zone because we are really looking for the date only. 
I added a try and catch just in case something changes.

Tested OK to ensure I didn't break anything.

### DIFF
--- a/src/ja/rawkuma/src/eu/kanade/tachiyomi/extension/ja/rawkuma/Rawkuma.kt
+++ b/src/ja/rawkuma/src/eu/kanade/tachiyomi/extension/ja/rawkuma/Rawkuma.kt
@@ -157,13 +157,13 @@ class Rawkuma: ParsedHttpSource() {
         document.select(chapterListSelector()).map { chapters.add(chapterFromElement(it)) }
         // Add date for latest chapter only
         document.select("time[itemprop=dateModified]").attr("datetime")
-            .let { chapters[0].date_upload = parseDate(it) }
+            .let { chapters[0].date_upload = parseDate(it.substringBefore("+")) }
         return chapters
     }
 
     private fun parseDate(date: String): Long {
         return try {
-            SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ssXXX", Locale.US).parse(date).time
+            SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss", Locale.US).parse(date).time
         } catch (e: ParseException) {
             0L
         }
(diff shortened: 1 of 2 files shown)


src/ja/rawkuma/src/eu/kanade/tachiyomi/extension/ja/rawkuma/Rawkuma.kt
```diff
@@ -157,13 +157,13 @@ class Rawkuma: ParsedHttpSource() {
...
         document.select("time[itemprop=dateModified]").attr("datetime")
-            .let { chapters[0].date_upload = parseDate(it) }
+            .let { chapters[0].date_upload = parseDate(it.substringBefore("+")) }
         return chapters
     }
 
     private fun parseDate(date: String): Long {
         return try {
-            SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ssXXX", Locale.US).parse(date).time
+            SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss", Locale.US).parse(date).time
```
ObserverOfTime: Wouldn't it be better to only remove the time zone if parsing it is not supported?
Otherwise, the time will appear off for everyone who's in a different time zone.

happywillow0: I didn't know exactly how exact we wanted to be. I felt like we didn't need to be this exact on when the page was last updated. 

Unless there is a setting somewhere that I overlooked, the time isn't even shown.

ObserverOfTime: `23:01 +0000` and `23:01 -0100` are two different days, aren't they?

happywillow0: Alright, I changed it. 

I was thinking that overall, it wouldn't be that bad in this use case for the date to be off by a little. Or we could also just switch to parse the text date.

a7410449155:
@happywillow0 
Thanks for your help~
However, the problem doesn't seem to be fixed yet.....

Even though I Updated the rawkuma extension to1.2.2,
Even though I clear tachiyomi app's data and reinstall all the extensions,
Everytime I tap into RawKuma category,I still see the desperate sentence
"unknown pattern character 'x' "............OTL

Hoping that you can help fix it again.
Thanks a lot.....

PR #2082, opened 2020-01-18
Mangahere Update Page List for Webtoons

### LINKED ISSUE #2107: [Bug] Mangahere : Webtoons Last Pages Not Loading
### Device information
* Tachiyomi extension version: 1.2.11
* Tachiyomi version: 0.8.4
* Android version: 9

## Steps to reproduce
1. Install Mangahere as an extension
2. Search for a manga that is published in webtoons view (like Solo Leveling, Peerless Dad, Berserk or Chouka Kou)
3.1 Download one chapter or 
3.2 go to the last page of a chapter

### Expected behavior
3.1 When downloading one chapter, it should finish without problems.
3.2 When viewing the last page of a chapter it should show the page.

### Actual behavior
3.1 Everything gets downloaded but the last page. There it stops and reports a failure.
3.2 There will be a error loading that page and "Try again" button will be displayed.

### Other details
The last page of these Mangas are also not available in the web view of Mangahere. It seems that the mangas this happens with, are shown in a webtoon view. Scrolling from up to down, instead of clicking through pages.
I also noticed that even though "I am the Sorcerer King" is also a webtoon, is published as such in Mangahere and the last page can not be loaded in the webview, there is no problem in Tachiyomi with its last page.
The last pages do not seem to be actual pages of the manga/mahnwa, but pages from the translators or so.
Try with recent chapters if it does not appear on other.

happywillow0:
Fixes #2107 

Mangahere normally has a insert image for the mangazone app as the last page. However, it doesn't get inserted properly for the webtoon viewer type. 

The invalid last image is now bypassed by reducing the total images by 1. It should not affect actual pages that are important. 

Also, changed to https

### DIFF
--- a/src/en/mangahere/build.gradle
+++ b/src/en/mangahere/build.gradle
@@ -5,7 +5,7 @@ ext {
     appName = 'Tachiyomi: Mangahere'
     pkgNameSuffix = 'en.mangahere'
     extClass = '.Mangahere'
-    extVersionCode = 11
+    extVersionCode = 12
     libVersion = '1.2'
 }
 
--- a/src/en/mangahere/src/eu/kanade/tachiyomi/extension/en/mangahere/Mangahere.kt
+++ b/src/en/mangahere/src/eu/kanade/tachiyomi/extension/en/mangahere/Mangahere.kt
@@ -19,7 +19,7 @@ class Mangahere : ParsedHttpSource() {
 
     override val name = "Mangahere"
 
-    override val baseUrl = "http://www.mangahere.cc"
+    override val baseUrl = "https://www.mangahere.cc"
 
     override val lang = "en"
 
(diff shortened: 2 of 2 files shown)


arkon:
Is there a way to actually tell if it's the ad page? I don't want to accidentally lose the last page sometimes.

happywillow0:
Loocool2 in discord might see the more in discord with people mentioning it. 

On the website end, the img url is encoded like the rest. However, even online, it doesn't load and stays as a "Loading..." image. For our reader, it should say retry, but it mostly affects downloads leaving one last image per chapter in the download queue.

I'm not sure if there is an easy check or not. I'll think about it.

happywillow0:
Another user posted a link to the mobile site. The mobile site doesn't use obfuscation.

Refactored code to drop de-obfuscation to just use plain scraping from the mobile site.

PR #2087, opened 2020-01-18
Update Preference Stubs & Update Manamoa extension

DitFranXX:
preference-stubs
-------------------
* Add `TwoStatePreference`
* Add `CheckBoxPreference`

Manamoa
-----------
* Code clean up.
* Simplify decoder by using kotlin functions.
* Add Experimental, Unstable feature: New domain auto fetcher 
    - Requires Android Nougat or newer as it needs Java 8's `CompletableFuture`
    - Using main thread black magic 🤷‍♂ 
        - It will causes a freeze when enter Manamoa Browse first time on each time app launched

### DIFF
--- a/src/ko/mangashowme/src/eu/kanade/tachiyomi/extension/ko/mangashowme/MangaShowMe.kt
+++ b/src/ko/mangashowme/src/eu/kanade/tachiyomi/extension/ko/mangashowme/MangaShowMe.kt
@@ -8,7 +8,6 @@ import android.os.Build
 import android.support.v7.preference.CheckBoxPreference
 import android.support.v7.preference.EditTextPreference
 import android.support.v7.preference.PreferenceScreen
-import android.util.Log
 import android.widget.Toast
 import eu.kanade.tachiyomi.extension.BuildConfig
 import eu.kanade.tachiyomi.network.GET
@@ -28,7 +27,6 @@ import java.text.SimpleDateFormat
 import java.util.*
 import java.util.concurrent.CompletableFuture
 import java.util.concurrent.TimeUnit
-import java.util.logging.Logger
 
 
 /**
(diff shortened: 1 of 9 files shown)


DitFranXX:
It need to bypass/scrape [DDoS-GUARD](https://ddos-guard.net/). It will be very hard for now. I will reopen this PR when I found some new way to autofetch new domain.

DitFranXX:
They had some changes in their system. It will now works without DDOS-Guard

src/ko/mangashowme/src/eu/kanade/tachiyomi/extension/ko/mangashowme/MangaShowMe.kt
```diff
@@ -322,17 +345,82 @@ class ManaMoa : ConfigurableSource, ParsedHttpSource() {
...
+                        super.completeExceptionally(e)
+                    }
+                }
+
+                val request: Request = Request.Builder().get()
+                    .url("http://13.229.223.203")
+                    .build()
+
+                val future = CallbackFuture()
+                okhttp3.OkHttpClient().newCall(request).enqueue(future)
```
arkon: Can you use the regular network client here instead of instantiating another one?

DitFranXX: It doesn't work with `network.cloudflareClient`. (Null returns or Freeze forever)
I think CF solver causes that on main thread blackmagic thingy...

happywillow0: Do you need cloudflare resolving for this part of the code? I'm not sure if it is exactly what arkon is requesting but you can also use `network.client`

DitFranXX: I thought regular client means `client` variable one.
I'm very sleepy now. Sorry :(

PR #2104, opened 2020-01-20
fix chapter list not reflecting blocked groups from mangadex

CarlosEsco:
Force network instead of using cache when getting manga info and chapter list for mangadex.

### DIFF
--- a/src/all/mangadex/src/eu/kanade/tachiyomi/extension/all/mangadex/Mangadex.kt
+++ b/src/all/mangadex/src/eu/kanade/tachiyomi/extension/all/mangadex/Mangadex.kt
@@ -347,13 +347,19 @@ abstract class Mangadex(
     }
 
     private fun apiRequest(manga: SManga): Request {
-        return GET(baseUrl + API_MANGA + getMangaId(manga.url), headers)
+        return GETNoCache(baseUrl + API_MANGA + getMangaId(manga.url), headers)
     }
 
     private fun searchMangaByIdRequest(id: String): Request {
-        return GET(baseUrl + API_MANGA + id, headers)
+        return GETNoCache(baseUrl + API_MANGA + id, headers)
     }
-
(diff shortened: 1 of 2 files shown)


src/all/mangadex/src/eu/kanade/tachiyomi/extension/all/mangadex/Mangadex.kt
```diff
@@ -347,13 +347,19 @@ abstract class Mangadex(
     }
 
     private fun apiRequest(manga: SManga): Request {
-        return GET(baseUrl + API_MANGA + getMangaId(manga.url), headers)
```
happywillow0: Doesn't the GET request have a third comma separated value for cache control?
GET(`url` , `headers`, `cache control`) ?

CarlosEsco:
Oh true

joseph619:
Would this also prevent chapters from blocked groups being automatically downloaded? Preferably they would be blocked from downloads if Autodownload new chapters is enabled.

happywillow0:
It should in theory. It all depends on what Mangadex is giving us. 
If they do not show up the Mangadex api / site, then they should not get downloaded. 

The only thing I can think of is what happens if a user gets logged out without them knowing and the chapter list gets refreshed. _Then_ the blocked group may re-appear and will get downloaded.

CarlosEsco:
Yes. This does prevent their API from sending it back.  It should currently do that also though. This just bypasses the sometimes cached call

src/all/mangadex/src/eu/kanade/tachiyomi/extension/all/mangadex/Mangadex.kt
CarlosEsco: fixed.  This is why i waited for a review :)  knew i was probably overlooking something simple

PR #2108, opened 2020-01-21
Fix TMO

happywillow0:
I thought it looked good but I double checked the code again and noticed that they changed the `height: 0px` changed to `height: 0` 
That broke the selector to filter out "extra" pages.


happywillow0:
Luckily, they are only adding a few extra dummy pages per chapter so it isn't that big of a deal. 

Built and tested the updated style selector in order to clean up the page list. Should be good to merge.

PR #2126, opened 2020-01-24
Remove deprecated Optional annotations on MangaPlus

alessandrojean:
Since the repo has upgraded the Kotlin version, the deprecated ```@Optional``` annotations can be removed on the MangaPlus extension.

Also, as discussed on #1654, @arkon, do you think you can add the ```kotlinx.serialization``` and the ```kotlinx-serialization-runtime``` dependencies to the main app, so we can change the ```implementation``` on MangaPlus' Gradle to ```compileOnly``` after a while when a new stable version of the app is released?

### DIFF
--- a/src/all/mangaplus/build.gradle
+++ b/src/all/mangaplus/build.gradle
@@ -6,7 +6,7 @@ ext {
     appName = 'Tachiyomi: MANGA Plus by SHUEISHA'
     pkgNameSuffix = 'all.mangaplus'
     extClass = '.MangaPlusFactory'
-    extVersionCode = 3
+    extVersionCode = 4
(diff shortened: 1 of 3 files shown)


arkon:
> Since the repo has upgraded the Kotlin version

Well not really, this would break stable users. Not entirely sure if we have a good plan for that. I'd love to force everyone to dev instead, but then we'd be dropping Android 4.x users completely if we decide to break all the extensions too.

alessandrojean:
> This would break stable users

I've not considered that, mistake on my part :sweat_smile:.

I'll close the PR for now, and if/when the support for KitKat is dropped, I can reopen this or create a new one.

PR #2128, opened 2020-01-26
Fix Japscan

### LINKED ISSUE #2125: [BUG] Japscan not working
### Device information
Japscan extension version: 1.2.13
Tachiyomi version: 0.8.4
Android version: 10
### Steps to reproduce
Install Japscan as extension
Select a manga from its catalog
Then try to read any chapter
### Expected behavior
First page of chapter is displayed and readable

### Actual behavior
Chapter is not loaded and a retry button appears. (Even with retrying multiple times it's still not working).

### Other details
Can't download any chapter.

happywillow0:
Fixes #2125 

I'm not totally sure what encoding measures that japscan has in place so I tried not to mess with it too much.

- The data url as part of the page list (options) are already an absolute url currently. (not sure what it was before.) Fixed it so there is no duplicate path interted. 
- Also fixed thumbnail urls which was broken for me because the img files were case sensitive for whatever reason.

### DIFF
--- a/src/fr/japscan/build.gradle
+++ b/src/fr/japscan/build.gradle
@@ -5,7 +5,7 @@ ext {
     appName = 'Tachiyomi: Japscan'
     pkgNameSuffix = 'fr.japscan'
     extClass = '.Japscan'
-    extVersionCode = 13
+    extVersionCode = 14
     libVersion = '1.2'
 }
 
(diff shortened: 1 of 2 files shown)


o01210o:
Here is how they encode their images (not all manga) https://github.com/inorichi/tachiyomi-extensions/issues/698#issuecomment-467929831

amar33:
Some news ? This is still not working...

nanartoon:
serch not working http error 404 add-on 1.2.14 release 0.8.4  android 8

amar33:
The research is still not working on japscan. " HTTP 404 ERROR "
1.2.14

PR #2141, opened 2020-01-28
TMO / Lector Update

happywillow0:
TMO and Lector broke. 
Use NineMangaES in the meantime.


happywillow0:
Looked at the source web pages re-tested,
Ready whenever you are.

PR #2163, opened 2020-02-01
Manga Park - chapter list preferences

SnakeDoc83:
Users can now filter their chapter lists.  The default is to show all sources and all chapters; which is what users see currently.  Other options are to show:

- Source with most chapters (plus any chapters that source doesn't have but others do)

- Smart list (try not to show duplicate chapters, no specific source is prioritized)

- Source priority (show only a specific source plus any chapters that source is missing; if a manga doesn't have any chapters from the prioritized source then all sources are shown).

I may have missed some of the source/scanlator names, but I didn't want to spend more time hunting down what those may be.  It'll be easy enough to add more if needed.

### DIFF
--- a/src/en/mangapark/src/eu/kanade/tachiyomi/extension/en/mangapark/MangaPark.kt
+++ b/src/en/mangapark/src/eu/kanade/tachiyomi/extension/en/mangapark/MangaPark.kt
@@ -1,22 +1,34 @@
 package eu.kanade.tachiyomi.extension.en.mangapark
 
+import android.annotation.SuppressLint
+import android.app.Application
+import android.content.SharedPreferences
+import android.support.v7.preference.ListPreference
+import android.support.v7.preference.PreferenceScreen
 import android.net.Uri
 import eu.kanade.tachiyomi.network.GET
+import eu.kanade.tachiyomi.source.ConfigurableSource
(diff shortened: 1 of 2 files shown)


src/en/mangapark/src/eu/kanade/tachiyomi/extension/en/mangapark/MangaPark.kt
```diff
@@ -419,4 +469,51 @@ class MangaPark : ParsedHttpSource() {
...
+        Injekt.get<Application>().getSharedPreferences("source_$id", 0x0000)
+    }
+
+    override fun setupPreferenceScreen(screen: androidx.preference.PreferenceScreen) {
+        val myPref = androidx.preference.ListPreference(screen.context).apply {
+            key = SOURCE_PREF_TITLE
+            title = SOURCE_PREF_TITLE
+            entries = arrayOf("All sources, all chapters", "Source with most chapters", "Smart list", "Prioritize source: Rock",
+                "Prioritize source: Duck", "Prioritize source: Mini", "Prioritize source: Fox", "Prioritize source: Panda")
+            entryValues = arrayOf("0", "1", "2", "3", "4", "5", "6", "7")
```
arkon: I generally find it less fragile to not use numbers for the values.

PR #2178, opened 2020-02-03
Update TMO / Lector

happywillow0:
They implemented the dummy image code into Lector and in the process, broke the selectors / div classes

Lector

- [x] - Fixes referrer 
- [x] - Fixes selectors + dummy filter
- [x] - Adds New Alternative Image Method

 
TMO

- [x] Fixes popular url that was giving 403
- [x] Fixes Search 403


happywillow0:
Tested 2/3/2020
Ready

Issue #2180, opened 2020-02-03
[Bug] Source Name : mangaswat


baytout00:
Mangaswat not working 
Arabic source

happywillow0:
If it is saying no source found, open it up in webview first to bypass the browser check. They are using another service not cloudflare but similar to protect their website. 

*You will want to be on the dev version or a fork to open up webview directly from the catalog screen.

baytout00:
I'm sorry, but it doesn't show anything.

https://drive.google.com/file/d/1mhc9ENmH0L_37cl2zvow5JeJtzaxT-I-/view?usp=sharing

happywillow0:
Install the dev version. http://tachiyomi.kanade.eu/latest
At the same screen ("Browse"), you can use the three dots menu and click "Open in webview"
Once the page loads, you can go back and open the catalog again. 

You can migrate your manga over by backing up from 0.8.4 stable and restoring it into dev.

baytout00:
Thanks, bro.

baytout00:
Hey, bro, the last version of the app (tachiyomi) doesn't work on my phone.Is there any way to run it in a Kitkat 4.4.4 version?

CarlosEsco:
No there is not

PR #2184, opened 2020-02-04
Add MangaSwat Warning

happywillow0:
Addresses #2180

Checks headers for sucuri bypass (similar to cloudflare). 

If not bypassed, prompts user to open webview. This attempts to replace error "no results found".

### DIFF
--- a/src/all/wpmangastream/build.gradle
+++ b/src/all/wpmangastream/build.gradle
@@ -5,7 +5,7 @@ ext {
     appName = 'Tachiyomi: WP Manga Stream'
     pkgNameSuffix = 'all.wpmangastream'
     extClass = '.WPMangaStreamFactory'
-    extVersionCode = 4
+    extVersionCode = 5
     libVersion = '1.2'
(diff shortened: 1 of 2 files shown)


happywillow0:
Not currently. 

Manual search, you have to open browse
Everything else I suppose would silently fail as with what is happening currently. 

One problem is that Sucuri doesn't give a http error code so it will only fail at the any of the page parsing steps. 

Do you want me to try to insert it into more places?

I'm looking at refactoring the code but I don't want to break global search with a thrown exception.

happywillow0:
Update: 

All functions except for search should have the check now. Global search will fail cleanly and silently similar to Mangadex with no results. Any other methods will result in the pop up message.

arkon:
I was going to suggest overriding the network client to have an interceptor to do the check. Might still be a good idea? I'm fine with the current implementation too.

happywillow0:
I had to learn how to build a custom interceptor but here we go.

PR #2187, opened 2020-02-04
Update TMO Lector

happywillow0:
Further fixes for the ever changing TMO / Lector sites.


happywillow0:
Have to make more changes

PR #2189, opened 2020-02-05
Fix TMO / Lector

happywillow0:
Further fixes to keep up to date with TMO and Lector.

Updated time method to emulate website methods.


CarlosEsco:
When you gonna finally throw in the towel?

PR #2192, opened 2020-02-05
Remove TMO (1/3)

happywillow0:
Site actively does not want us. I will accept them working with us if they want to have app support. 
I think I might even accept them to inject their own ads that say visit us on the web. 

Part 1/3 - Remove build so it doesn't build. 
Part 2/3 - Rest of code
Part 3/3 - Repo / apk cleanup

Otherwise, comments welcome below.


arkon:
Why have parts 1 and 2 separately?

happywillow0:
I'm currently only on github web. There is no easy way of selecting multiple items to delete at once. 
I was going to hit the rest when I get home and can use the desktop version.

happywillow0:
Also, the previous build stalled so I was thinking to get it marked as `Outdated `before the previous PR gets pushed out.

arkon:
@edgarMejia Not sure if you have any input on this?

PR #2196, opened 2020-02-06
Remove TMO (1/2)

happywillow0:
Option to remove TMO

TMO has been and continues to be hostile to scrapers including third party apps.
I am willing to compromise and add features that would benefit owners if they so wish. 

If site continues to cause intended disruptions, this PR is to drop support of site.   

@edgarMejia, we wanted to check if you had any other thoughts that you wish to share.


edgarMejia:
First of all thank you for your support.

From the beginning TMO gave us too many problems so I think that at the moment it is best to remove the extension.

There are currently good options in Spanish in the application like **NinemangaEs** and **VCP & VMP** for NSFW content.

There are also good options to add like MangaMx:
https://github.com/inorichi/tachiyomi-extensions/issues/1098
(I don't know why this request was closed).

I don't think they can keep this war against scrapers for a long time so I will be pending to add the extension when the time comes ...

happywillow0:
> There are also good options to add like MangaMx:

I actually created that myself! 😄  
https://github.com/inorichi/tachiyomi-extensions/blob/master/src/es/mangamx/src/eu/kanade/tachiyomi/extension/es/mangamx/MangaMx.kt

> I don't think they can keep this war against scrapers for a long time so I will be pending to add the extension when the time comes ...

I was thinking about keeping the code around but I think GitHub does still keep the history in our commit history.

edgarMejia:
> I actually created that myself! 😄
> MangaMx.kt

Nani?? hahahahaha xd
I had started to create the extension I didn't see the one you published, only mine appeared (I never finished it).

>I was thinking about keeping the code around but I think GitHub does still keep the history in our commit history.

Yes the code is always kept in the commits history,  it happened with another extension that died (LeoManga).

PR #2214, opened 2020-02-08
[Japscan] Fix search

oktapodia:
Hello,

This PR is fixing the search on Japscan (the previous url does not exist anymore on japscan).

However, this is taking ages to perform a search (130 sequential calls).

Might be better to add a bubble?

Feel free to make your suggestions

Fix #2128


arkon:
As mentioned on Discord, definitely not a fan of this approach.

Issue #2222, opened 2020-02-09
[Bug] Japscan: Search not working


dryx01:
### Device information
* Tachiyomi extension version: 1.2.14
* Tachiyomi version: 0.8.4 and dev R1705
* Android version: 10

## Steps to reproduce
1. Install Tachiyomi
2. In the app, install japscan extension.
3. Make a search in the "Catalogues"

### Expected behavior
The search should be successful.

### Actual behavior
No matter what research, it doesn't work

### Other details
I tested the extension with stable and dev version, both doesn't work.
An exemple here:
During the search
![Screenshot_20200209-162827](https://user-images.githubusercontent.com/6872170/74108767-d41eb980-4b7d-11ea-87fb-1ba015f9060c.jpg)
After
![Screenshot_20200209-162832](https://user-images.githubusercontent.com/6872170/74108792-f9abc300-4b7d-11ea-8560-750564a40b65.jpg)

Thanks in advance.

happywillow0:
Just FYI, 

Japscan themselves does not have a search function built into their website. The previous system is broken too (as you noted) because they also took out a alphabetical list of all manga.

dryx01:
So then, there is a method/procedure do add manually manga from japscan without this search function ? Like a conf file to modify ?
Currently there is no way to add more from this source.

happywillow0:
Two of us were talking about different ways to get what you need in discord. 
I might try to type up my method when I have time and test it out. 

If it is early in the alphabet, you can scroll your way though browse currently, or what I do for discovering new manga is to pick it up when it shows up in the Latest section as things get updated. (Still doesn't help for old finished manga though)

We'll link to this issue any workarounds that we might try.

dryx01:
Perhaps a method where the user find the manga in a webview and click on a button when the good manga is displayed.

Thanks for the fast reply.

If the issue is known, then I close the subject.

PR #2226, opened 2020-02-10
Japscan Browse via Filters

happywillow0:
Workaround for #2222
Related / Alternative to #2214

### DIFF
--- a/src/fr/japscan/src/eu/kanade/tachiyomi/extension/fr/japscan/Japscan.kt
+++ b/src/fr/japscan/src/eu/kanade/tachiyomi/extension/fr/japscan/Japscan.kt
@@ -4,24 +4,17 @@ import android.graphics.Bitmap
 import android.graphics.BitmapFactory
 import android.graphics.Canvas
 import android.graphics.Rect
+import android.net.Uri
 import eu.kanade.tachiyomi.network.GET
-import eu.kanade.tachiyomi.network.asObservableSuccess
-import eu.kanade.tachiyomi.source.model.FilterList
-import eu.kanade.tachiyomi.source.model.MangasPage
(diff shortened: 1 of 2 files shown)


src/fr/japscan/src/eu/kanade/tachiyomi/extension/fr/japscan/Japscan.kt
```diff
@@ -262,4 +243,31 @@ class Japscan : ParsedHttpSource() {
...
+    override fun getFilterList():FilterList {
+        val totalPages = pageNumberDoc?.select("li.page-item:last-child a")?.text()
+        val pagelist = mutableListOf<Int>()
+        var filterList:FilterList = if (!totalPages.isNullOrEmpty()) {
+            for (i in 0 until totalPages.toInt()) {
+                pagelist.add(i+1)
+            }
+            FilterList(
+                Filter.Header("Alphabetical Page"),
+                Filter.Header("Page alphabétique"),
```
arkon: What's up with the English and French of the same thing?

happywillow0: ¯\_(ツ)_/¯
I code in English and thought I it would make more sense to use french because it is a french extension. I'll remove it.

arkon: Yeah, it makes sense to have the French. I'm just not sure why there's both.

happywillow0: Changed another string while I was at it.

nanartoon:
the solution is adding all the mangas of the source to my libray for the search option

PR #2234, opened 2020-02-11
Add MangaYabu

### LINKED ISSUE #2232: [Source Request] mangayabu [Brazil]
### Source Website
Website can be found at 
https://mangayabu.com

### Other Details
updates manga faster than other Brazilian sources

alessandrojean:
Closes #2232.


src/pt/mangayabu/src/eu/kanade/tachiyomi/extension/pt/mangayabu/MangaYabu.kt
```diff
@@ -0,0 +1,144 @@
...
+
+    override fun searchMangaFromElement(element: Element): SManga = SManga.create().apply {
+        title = element.select("div.contento span.search-name")!!.text()
+        thumbnail_url = "$baseUrl/${element.select("img.search-thumb")!!.attr("src")}"
+        setUrlWithoutDomain(element.attr("href"))
+    }
+
+    override fun searchMangaNextPageSelector(): String? = null
+
+    override fun mangaDetailsRequest(manga: SManga): Request = GET(baseUrl + manga.url, headers)
```
happywillow0: It doesn't matter too much but isn't the request the same as the default? 
There isn't a need to have the override in here for `mangaDetailsRequest` `chapterListRequest` `pageListRequest`

alessandrojean: I haven't checked the default methods, but probably yes, I'll take a look.

alessandrojean:
I can't find a better solution for the different slugs used for the chapters and the mangas, but I think there are very few manga that have this problem and this solution will work for now.

arkon:
Seems like the site has CF protection on so I can't even access it (or is it region-blocked?)...

alessandrojean:
Probably region-blocked as many others Brazilian sources.

src/pt/mangayabu/src/eu/kanade/tachiyomi/extension/pt/mangayabu/MangaYabu.kt
```diff
@@ -0,0 +1,150 @@
...
+
+    override fun imageUrlParse(document: Document) = ""
+
+    /**
+     * Some mangas doesn't use the same slug from the chapter url, and
+     * since the site doesn't have a proper popular list yet, we have
+     * to deal with some exceptions and map them to the correct
+     * slug manually.
+     *
+     * It's a bad solution, but it's a working one for now.
```
arkon: You might be able to do some janky network calls to get stuff if you need to.

PR #2264, opened 2020-02-18
MangaLib: added "Accept" header, should fix main image server problems

DumbCodeGenerator:


src/ru/libmanga/src/eu/kanade/tachiyomi/extension/ru/libmanga/LibManga.kt
```diff
@@ -45,6 +45,11 @@ class LibManga : ConfigurableSource, HttpSource() {
         else -> "https://img3.mangalib.me"
     }
 
+    override fun headersBuilder() = Headers.Builder().apply {
+        add("User-Agent", "Mozilla/5.0 (Windows NT 6.3; WOW64)")
+        add("Accept", "*/*")
```
happywillow0: Not that it matters too much but unless you want to completely override the the original function, you can use 
`super.headersBuilder().add("Accept", "*/*")`
That way, you don't need to re-add the user agent. 

Also in general, because it is a builder, instead of using `.apply{}`, you should be able to just use
`.add()`

arkon: +1, let's add to the existing builder.

PR #2269, opened 2020-02-20
Add MangaMainac

happywillow0:
Not that anyone really asked for it but as280093 mentioned it so I threw together this extension which bundles all the MangaMainac based sites into one. 

I think rather than calling them single manga sites, they are just manga pages using different base url as part of a aggregator but without a overall home page / manga list.


happywillow0:
Icons needed

as280093:
**Icons:**
[ic_launcher.zip](https://github.com/inorichi/tachiyomi-extensions/files/4229017/ic_launcher.1.zip)

src/en/mangamainac/build.gradle
```diff
@@ -0,0 +1,12 @@
+apply plugin: 'com.android.application'
+apply plugin: 'kotlin-android'
+
+ext {
+    appName = 'Tachiyomi: MangaMainac '
```
arkon: Extra space at the end. Also spelled "Maniac" wrong.

src/en/mangamainac/src/eu/kanade/tachiyomi/extension/en/mangamainac/MangaMainac.kt
```diff
@@ -0,0 +1,143 @@
...
+            Pair("Berserk","http://berserkmanga.net"),
+            Pair("The Quintessential Quintuplets","https://5-toubunnohanayome.net"),
+            Pair("Kaguya Wants to be Confessed To","https://kaguyasama.net"),
+            Pair("Domestic Girlfriend","https://domesticgirlfriend.net"),
+            Pair("Black Clover","https://w5.blackclovermanga.com"),
+            Pair("One Piece","https://1piecemanga.net"),
+            Pair("The Promised Neverland","https://neverlandmanga.net"),
+            Pair("Shingeki no Kyojin","https://readshingekinokyojin.com"),
+            Pair("Nanatsu no Taizai","https://w1.readnanatsutaizai.net")
+        ).sortedBy { it.first }.distinctBy { it.second }
```
arkon: Would there be duplicates?

src/en/mangamainac/build.gradle
happywillow0: Should we change the spelling even though I copied it from the site?
> Theme: mangamainac by Animemaniac.co.

src/en/mangamainac/src/eu/kanade/tachiyomi/extension/en/mangamainac/MangaMainac.kt
happywillow0: Since there was no main list, I was just copying pasting from all the tabs I opened. 

There shouldn't be duplicates as I hope I know what I'm doing but since the list didn't look clean, I added the sort and de-duplicate just in case I (or anyone else) added to the list and did duplicate one by accident. 

I can take it out if you like.

src/en/mangamainac/build.gradle
arkon: Ah... nevermind then. Should still remove the extra space though.

src/en/mangamainac/src/eu/kanade/tachiyomi/extension/en/mangamainac/MangaMainac.kt
arkon: I think it's fine, we'll try to catch duplicates in the PRs.

src/en/mangamainac/build.gradle
happywillow0: Done (I thought I put it earlier but I guess I didn't hit the right button)

PR #2270, opened 2020-02-20
Add Search to JapScan

happywillow0:
Use Duck Duck Go to search within JapScan


amar33:
Some news about that ? There is so long time we can't research with japscan.

happywillow0:
It not the best search but you now have two options. You can browse by page similar to how you would use the website using the filter. 

The app extension now runs a search though duckduckgo which helps provide some results. I don't think they are the best results but it works most of the time, hopefully.

amar33:
Oh i didnt knew that thanks you!!

PR #2280, opened 2020-02-22
webtoon age gate fix

E3FxGaming:
* can browse popular webtoons again

* can update webtoon chapter lists again

* can search for webtoons again

* version number updated

Webtoon implemented an age gate that interfered with scraping.

We now have to send a cookie with every request to make it work again.

I've used code from commit #1029 which fixed a similar issue.

Finding the problem, adapting the fix and testing it is what I did.

### DIFF
--- a/src/all/webtoons/src/eu/kanade/tachiyomi/extension/all/webtoons/Webtoons.kt
+++ b/src/all/webtoons/src/eu/kanade/tachiyomi/extension/all/webtoons/Webtoons.kt
@@ -19,15 +19,16 @@ abstract class Webtoons(override val lang: String, open val langCode: String = l
 
     override val client = super.client.newBuilder()
         .cookieJar(object : CookieJar {
-            override fun saveFromResponse(url: HttpUrl, cookies: MutableList<Cookie>) {}
-            override fun loadForRequest(url: HttpUrl): MutableList<Cookie> {
-                return ArrayList<Cookie>().apply {
-                    add(Cookie.Builder()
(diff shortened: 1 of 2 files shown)


src/all/webtoons/src/eu/kanade/tachiyomi/extension/all/webtoons/Webtoons.kt
```diff
@@ -20,6 +17,22 @@ abstract class Webtoons(override val lang: String, open val langCode: String = l
...
+        .cookieJar(object : CookieJar {
+            override fun saveFromResponse(url: HttpUrl, cookies: MutableList<Cookie>) {}
+            override fun loadForRequest(url: HttpUrl): MutableList<Cookie> {
+                return ArrayList<Cookie>().apply {
+                    add(Cookie.Builder()
+                        .domain("www.webtoons.com")
+                        .path("/")
+                        .name("ageGatePass")
+                        .value("true")
+                        .build()) }
```
ObserverOfTime: You could write this like so:
```kotlin
return listOf<Cookie>(
    Cookie.Builder()
    .domain("www.webtoons.com")
    .path("/")
    .name("ageGatePass")
    .value("true")
    .build()
)
```

E3FxGaming:
I've implemented the requested change.

I also changed the cookies parameter of saveFromResponse to List instead of MutuableList in order to stay closer to what the okhttp3 super class CookieJar.java contains.

PR #2289, opened 2020-02-23
Update Manamoa Extension

DitFranXX:
v.1.2.19
=====
* Updated: Search function - now support order, and artist search (previously  was Exact Search)
* Updated: BaseUrl
* New feature: ID Search
* New feature: Url Activity to open manga directly
* New experiment feature: Latest updated manga


DitFranXX:
PS. 'Recent' in search filter doesn't means 'Latest'. It means 'Recently added manga'. (So extension  not support latest (for now.))

DitFranXX:
Ready to merge.

PR #2293, opened 2020-02-24
Move Hatigarm Scans

SnakeDoc83:
Closes https://github.com/inorichi/tachiyomi-extensions/issues/2291

Genkan changes:
* Added Hatigarm Scans (formerly part of MMRCMS) and bumped its `versionId`.  Kind of went back and forth on whether I should, let me know if you think it's unnecessary to change `versionId`.  
* Removed Psycho Play, they got DMCAed earlier this month and decided to shutdown their website.
* Deleted a couple lines I forgot to delete when I removed KaguyaDex and KomiScans in a previous update.

MMRCMS changes:
* Removed Hatigarm Scans


arkon:
I guess it'd be good to bump the versionId if it's not backwards compatible?

SnakeDoc83:
URLs for the manga are different and I don't think the old ones will forward.  They only have 5 manga up right now (was 3 when I opened the PR, I think their catalog was more than that before) and only new chapters for them.  So, without the bump, if users have old chapters or older manga from MMRCMS-Hatigarm in their database they won't realize they aren't available currently unless they manually refresh their chapter list.  If old chapters do get uploaded eventually they won't match the old URLs.

PR #2294, opened 2020-02-25
MangaFree - Remove https

happywillow0:
Reported on Discord:

MangaFree doesn't support https 
`Unable to connect` / refused connection.


CarlosEsco:
sigh why would they not use https

PR #2299, opened 2020-02-27
Update Kissmanga.kt

Light-Wing:
making Kissmanga source load faster by loading more info on the results page.

this is done with a cookie to show the "Full List" instead of the compact list...

try to run the current version, then this version and you'll see the difference clearly

### DIFF
--- a/src/en/kissmanga/src/eu/kanade/tachiyomi/extension/en/kissmanga/Kissmanga.kt
+++ b/src/en/kissmanga/src/eu/kanade/tachiyomi/extension/en/kissmanga/Kissmanga.kt
@@ -3,12 +3,18 @@ package eu.kanade.tachiyomi.extension.en.kissmanga
 import com.squareup.duktape.Duktape
 import eu.kanade.tachiyomi.network.GET
 import eu.kanade.tachiyomi.network.POST
+import eu.kanade.tachiyomi.network.asObservableSuccess
 import eu.kanade.tachiyomi.source.model.*
 import eu.kanade.tachiyomi.source.online.ParsedHttpSource
+import eu.kanade.tachiyomi.util.asJsoup
 import okhttp3.*
 import org.jsoup.nodes.Document
 import org.jsoup.nodes.Element
+import org.jsoup.select.Elements
+import rx.Observable
+import java.net.URLEncoder
 import java.text.SimpleDateFormat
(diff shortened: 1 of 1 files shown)


happywillow0:
Actually, the information is already there in the "tool tip"

All you really needed to do is add one line to the popularMangaFromElement()
`manga.thumbnail_url = element.select("td").attr("title").substringAfter("src=\"").substringBefore("\"")`

Light-Wing:
cool, when i looked at the source code of the page i couldn't figure out how to get the image from the tool tip, as it appears at the end of the page...
can you explain to me what Jsoup.parseBodyFragment(element.select()).select() is doing? and how is it matching each tooltip to the right source?

happywillow0:
The tool tips at the bottom is actually generated by javascript and added to the page. The data that javascript uses is already part of the table. 

`popularMangaSelector() or latestUpdatesSelector() = "table.listing tr:gt(1)"` is already pulling the the following elements:

```html
<tr class="odd">
<td title='
                <img width="120" height="165" src="https://kissmanga.com/Uploads/Etc/8-24-2011/5569412cover.jpg" style="float: left; padding-right: 10px" />
                <div style="float: left; width: 300px">
                    <a class="bigChar" href="/Manga/One-Piece">One Piece</a>
                    <p>
                        Greatness, Glory, Gold. A Pirate Named Gold Roger also known as the Pirate King has conquered this all. He was executed with an unknown reason but before he died he had revealed his last word about The ...
                    </p>
                </div>'>
<a href="/Manga/One-Piece">
One Piece</a>
&nbsp;&nbsp;<img style="vertical-align: middle" title="Most Popular Manga" src="../../Content/images/hot.png" />
</td>
<td>
<a href="/Manga/One-Piece/Chapter-972?id=595147">Chapter 972</a>
</td>
</tr>
```

Therefor, in order to extract the image, we do:

1. `element.select("td").attr("title")` to select the TD element and extract the title attribute as a string. 
2. `Jsoup.parseBodyFragment()` to apply jsoup parsing to the string because it is a html fragment (not a full page)
3. `select("img").attr("abs:src")`  to select the img element and extract the image url. 

Do you want me to submit my PR?

Light-Wing:
Thanks for explaining, and definitely submit it, who doesn't like a
one-line fix? 😄

PR #2305, opened 2020-02-28
Update MRM Filter Code

happywillow0:
Tried to improve filter code by utilizing existing network calls. That didn't work. 

Found out that they changed part of the website and categories broke. 
Fixed categories and additional additional new filters found on website. 

Also ran the default code reformat in Android Studio to clean up some of my code.


happywillow0:
*sigh* sorry. I just realized I just broke the extension a little bit and why it was sort of working before. 
I need to update it again with some cache control so expect a new update sometime soon. 

Good news though, if what I'm planning works, it will get smooth out some of the network calls.

PR #2314, opened 2020-02-29
Fix MRM Filters

happywillow0:
I broke the filters with the last update. 

- Re-organized code 
-- Popular, Latest, Search, Details, Chapters, Pages, Filters
- Fixed / Updated filters. 
-- No longer attempts to parse filter on startup. 
-- Now waits until user enters browse and bypasses cloudflare captcha.


src/all/myreadingmanga/src/eu/kanade/tachiyomi/extension/all/myreadingmanga/MyReadingManga.kt
```diff
@@ -256,32 +264,31 @@ open class MyReadingManga(override val lang: String) : ParsedHttpSource() {
     override fun imageUrlRequest(page: Page) = throw Exception("Not used")
     override fun imageUrlParse(document: Document) = throw Exception("Not used")
 
+    //Filter Parsing, grabs pages as document and filters out Genres, Popular Tags, and Categories, Parings, and Scan Groups
+    private var filtesrCached = false
```
arkon: Typo

happywillow0: Ah darn. Thanks for catching that. 
Done / Renamed all instances

PR #2317, opened 2020-02-29
[ru-mangalib] Realtime calculate of image server

Pavkazzz:
Remove manual settings for configure image sever, because of large community report of doesn't working chapters in main(default) server

### DIFF
--- a/src/ru/libmanga/src/eu/kanade/tachiyomi/extension/ru/libmanga/LibManga.kt
+++ b/src/ru/libmanga/src/eu/kanade/tachiyomi/extension/ru/libmanga/LibManga.kt
@@ -1,9 +1,5 @@
 package eu.kanade.tachiyomi.extension.ru.libmanga
 
-import android.app.Application
-import android.content.SharedPreferences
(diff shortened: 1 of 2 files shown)


src/ru/libmanga/src/eu/kanade/tachiyomi/extension/ru/libmanga/LibManga.kt
```diff
@@ -239,6 +178,11 @@ class LibManga : ConfigurableSource, HttpSource() {
 
         val chapInfoJson = jsonParser.parse(chapInfo).obj
 
+        val imageServerUrl: String = when(chapInfoJson["imgServer"].string){
+            "compress" -> "https://img3.mangalib.me"
+            else -> "https://img3.mangalib.me"
```
happywillow0: Are these supposed to be the same?

Pavkazzz: Thx for notice. Fix

PR #2318, opened 2020-02-29
Add warning to Mangadex search

happywillow0:
Adds a check for the login button during a search call and if exists, throws a message to login. 
Would this be useful?


bbtora:
Can you perhaps also work on showing this message during Global Search as well?

happywillow0:
So far, global search will hide any extensions with no results or error results. The extension is programmed to just throw an error. 

You can submit a request over on the main app issue tracker if this is important to you.

PR #2319, opened 2020-02-29
Add KissManga Trending Manga option

happywillow0:
I don't know if this is bad feature creep or or a good thing giving users what they are looking for

- Adds preference to switch to Trending manga on KissManga rather than most popular.


CarlosEsco:
Could this not just be a filter option?  Press here to sort by trending

happywillow0:
Yes! That would be a good idea. Let me take a look.

happywillow0:
Re-written with filters rather than the preference screen.

PR #2324, opened 2020-03-01
Manga Park - chapter list preferences v2

happywillow0:
It has been a while since the pending review on #2163

Since this is a collaborative effort, I patched on top of previous patch so snakedoc83 should still get credit.

Original Description:  
Users can now filter their chapter lists. The default is to show all sources and all chapters; which is what users see currently. Other options are to show:

- Source with most chapters (plus any chapters that source doesn't have but others do)

- Smart list (try not to show duplicate chapters, no specific source is prioritized) 

- Source priority (show only a specific source plus any chapters that source is missing; if a manga doesn't have any chapters from the prioritized source then all sources are shown).


as280093:
@happywillow0  can you update icons too
I am fixing my previous mistakes while making icons without increasing version count 
[ic_launcher.zip](https://github.com/inorichi/tachiyomi-extensions/files/4271415/ic_launcher.4.zip)

happywillow0:
Icons updated. 👍

src/en/mangapark/src/eu/kanade/tachiyomi/extension/en/mangapark/MangaPark.kt
```diff
@@ -419,4 +469,54 @@ class MangaPark : ParsedHttpSource() {
...
+                val entry = entryValues.get(index) as String
+                preferences.edit().putString(SOURCE_PREF, entry).commit()
+            }
+        }
+        screen.addPreference(myPref)
+    }
+    private fun getSourcePref(): String? = preferences.getString(SOURCE_PREF, "all")
+    
+    companion object {
+        private const val SOURCE_PREF_TITLE = "Chapter List Source Pref"
```
arkon: Nitpick: remove "Pref"

src/en/mangapark/src/eu/kanade/tachiyomi/extension/en/mangapark/MangaPark.kt
```diff
@@ -419,4 +469,54 @@ class MangaPark : ParsedHttpSource() {
...
+        Injekt.get<Application>().getSharedPreferences("source_$id", 0x0000)
+    }
+
+    override fun setupPreferenceScreen(screen: androidx.preference.PreferenceScreen) {
+        val myPref = androidx.preference.ListPreference(screen.context).apply {
+            key = SOURCE_PREF_TITLE
+            title = SOURCE_PREF_TITLE
+            entries = arrayOf("All sources, all chapters", "Source with most chapters", "Smart list", "Prioritize source: Rock",
+                "Prioritize source: Duck", "Prioritize source: Mini", "Prioritize source: Fox", "Prioritize source: Panda")
+            entryValues = arrayOf("all", "most", "smart", "rock", "duck", "mini", "fox", "panda")
```
arkon: You could extract the arrays to avoid potentially messing up keeping them in sync in the future.

src/en/mangapark/src/eu/kanade/tachiyomi/extension/en/mangapark/MangaPark.kt
```diff
@@ -109,20 +121,58 @@ class MangaPark : ParsedHttpSource() {
...
-    //TODO Maybe make it possible for users to view the other versions as well?
-    override fun chapterListSelector() = ".stream .volume .chapter li"
+    override fun chapterListSelector() = ".volume .chapter li"
 
-    override fun chapterFromElement(element: Element) = SChapter.create().apply {
+    private fun chapterFromElement(element: Element, source: String) = SChapter.create().apply {
         url = element.select(".tit > a").first().attr("href").replaceAfterLast("/", "")
         name = element.select(".tit > a").first().text()
+        // Get the chapter number or create a unique one if it's not available
+        chapter_number = Regex("""\b\d+\.?\d?\b""").find(name)?.value?.toFloatOrNull() ?: ".${name.hashCode().absoluteValue}".toFloat()
```
arkon: I feel like using the hash is even worse than just not setting it. It'd lead to some weird sorting, wouldn't it?

SnakeDoc83: If you don't set a chapter number, all the chapters without them would be considered duplicates and be removed for some of the sorting options.  By adding a unique chapter number we can avoid this.  Yes, this may lead to a chapter being sorted improperly but it should be a rare case.

src/en/mangapark/src/eu/kanade/tachiyomi/extension/en/mangapark/MangaPark.kt
happywillow0: Done, Arrays Extracted

src/en/mangapark/src/eu/kanade/tachiyomi/extension/en/mangapark/MangaPark.kt
happywillow0: Done. Removed "Pref"

PR #2329, opened 2020-03-01
fix: zh-bainianmanga

Trim21:


Trim21:
Should the version code be bumped? The original version is broken at fetching page image.

arkon:
Shoot, I missed that (I just saw that `build.gradle` was updated).

Yes, it should be bumped.

PR #2336, opened 2020-03-01
NewToki - Fix Imports / null handling

happywillow0:
This might fixes the build error?


src/ko/newtoki/src/eu/kanade/tachiyomi/extension/ko/newtoki/NewToki.kt
```diff
@@ -246,7 +246,7 @@ open class NewToki(override val name: String, private val defaultBaseUrl: String
         screen.addPreference(baseUrlPref)
     }
 
-    private fun getPrefBaseUrl(): String = preferences.getString(BASE_URL_PREF, defaultBaseUrl)
+    private fun getPrefBaseUrl(): String = preferences.getString(BASE_URL_PREF, defaultBaseUrl) ?:defaultBaseUrl
```
arkon: I'll just push a commit that adds `!!`, since it should be returning that as a fallback anyway.

PR #2339, opened 2020-03-02
Add MyAnimeList Free Manga

### LINKED ISSUE #2302: [Source Request] My Anime List Free Manga
### Source Website
Website can be found at https://myanimelist.net/read/manga.

### Other Details
Free manga hosted legally on MAL

alessandrojean:
Should close #2302 for now.

### Some warnings

Some chapters requires login to read and they will not be shown until the login is done at the WebView. If the person isn't logged and the manga has chapters missing, a message will be shown at the description.

I don't know if the login can cause some kind of banishment in the site, at least when I tried with my account nothing happened.

Some chapters are wrong in the site, like Nanatsu no Taizai first chapter is in fact the second, and the most recent (302) is a chapter from other manga (Chihayafuru). Nothing can be done in the extension side, I think.

### Decryption of the images

I couldn't make the native way of decrypt the page image work, so I used Duktape instead. For reference, I opted to keep the native function in the code.

The decrypt functions are commented explaining the process, so someone with more knowledge that wants to help can solve this problem later.

For reference, the decryption works in the same way of the MangaPlus one.


ali486316:
Hi how to install wallet address

src/en/myanimelist/src/eu/kanade/tachiyomi/extension/en/myanimelist/MyAnimeList.kt
```diff
@@ -0,0 +1,215 @@
...
+            .map { it.toInt().toByte() }
+            .toByteArray()
+    }
+
+    /**
+     * For some odd reason, probably with the way Kotlin handles UInt is different of
+     * the Javascript way, using a native method will not work, as some bytes
+     * becomes different values.
+     *
+     * Replacing the instances of [toInt] with [toUint] don't solve the problem either.
```
arkon: You probably want `toByte()`

alessandrojean: Tried now with `toUByte().toInt()` and it worked, I will update the code. 

For some reason, use only `toByte()` or `toByte().toInt()` does not work, probably related to [this](https://stackoverflow.com/q/38651192) somehow.

alessandrojean:
Native method now works, but I used a [workaround](https://stackoverflow.com/a/38660733) to not use `UByte` and the annotations that its usage requires. Should be finished now.

PR #2373, opened 2020-03-05
[New Source]- Genkan- Edelgarde Scans (en)

as280093:


src/all/genkan/src/eu/kanade/tachiyomi/extension/all/genkan/GenkanFactory.kt
```diff
@@ -13,10 +13,10 @@ class GenkanFactory : SourceFactory {
         ZeroScans(),
         ReaperScans(),
         TheNonamesScans(),
-        HatigarmScans()
+        HatigarmScans(),
+		EdelgardeScans()
```
arkon: Indentation is off

src/all/genkan/src/eu/kanade/tachiyomi/extension/all/genkan/GenkanFactory.kt
```diff
@@ -13,10 +13,10 @@ class GenkanFactory : SourceFactory {
...
     )
 }
-
```
arkon: Revert this

PR #2389, opened 2020-03-07
[New] Ciayo Comics

### LINKED ISSUE #2296: [Source Request] Ciayo
Need extensions of https://www.ciayo.com

Thanks you

happywillow0:
Closes #2296


happywillow0:
PR is ready for review. 

Site has 2 versions, one ID and one EN. ID has more content. 
It really didn't seem right to put the site under "all" this time because it really is only 2 languages where it would bloat the "all" section. 

Should I switch it to a factory and / or "all"
Or should we add I was thinking, two separate extensions?

arkon:
The `all` factory makes the most sense to be from a maintenance standpoint. There isn't a great way to deal with the bloat, unfortunately.

happywillow0:
I'm going to re-open this.

Issue #2315, opened 2020-02-29
[Bug] Japscan : The pages dont load 


amar33:
### Device Information
* Tachiyomi extension version: 1.2.16
* Tachiyomi version: 0.8.4
* Android version: 9.00

### Actual behavior
![Screenshot_2020-02-29-13-20-41-675_eu kanade tachiyomi](https://user-images.githubusercontent.com/56921708/75607412-23aa3280-5af7-11ea-88f6-45faca1cfa07.jpg)

### Other details
I can't read all the manga i have in my library from japscan. Since Yesterday.

o01210o:
Same.

HerrCraziDev:
I'm having the same issue. Purging cache and cookies as well as trying from several dirrerent networks (different wifi, mobile service, ect) didn't helped.
Seems like the only thing not working is loading the images themselves. No exception is thrown in the logcat, but maybe that's because I'm not using the Dev version. 

The website works as usual. 

I may try looking at the code tonight but I would appreciate any help on how to debug extensions. 

(btw Kingdom is awesome! 😉)

happywillow0:
I took a quick peak on my phone earlier and the data-src url was throwing 404. Maybe some more scrambling but in url form?

happywillow0:
Instead of scrambling the images, they seem to be scrambling the links. 
I'm not sure how it works but hakuneko has pushed their own patch. 
https://github.com/manga-download/hakuneko/commit/fc9aee2065284d738d1fe341f6ecea65ad78cd0e

HerrCraziDev:
Hakuneko seems to be using a script, probably taken from Japscan itself (see [manga-download/hakuneko@fc9aee2](https://github.com/manga-download/hakuneko/commit/fc9aee2065284d738d1fe341f6ecea65ad78cd0e), line 72. Unfortunately it is obfuscated.

I took a look into Japscan's own scripts and I have identified [this obfuscated script](https://www.japscan.co/zjs/292795fd975f554504f025abdec08ea4.js) as the one that _seems_ to be responsible for generating the image URL for a given page.

Reverse-engineering these might give us more information about how Japscan mangles their image URLs.

HerrCraziDev:
I think I managed to find what was causing the issue: as we supposed above, Japscan changed their image links, but those still present in the `select#pages`'s `data-img` attributes are the old ones.

Hence, we seem to be left with two options :

- Figure out the relationship between pages and `data-img` direct URLs
[...]

PR #2404, opened 2020-03-11
Japscan not loading images dirty fix (#2315)

morcefaster:
Related issue: https://github.com/inorichi/tachiyomi-extensions/issues/2315

Japscan are encoding their image urls with an ever-changing substitution cipher concealed in an oversized and obfuscated js file.

This fix relies on an existing solved image url list. By downloading a predetermined chapter and analyzing page urls we can determine the key for the substitution cipher, which is used to decode requested chapter's pages.

As a result, the parser works again at the price of one additional GET call per chapter.

EDIT:
Updated to use two chapters for full alphabet keysheet. Now works for all manga.


morcefaster:
The provided list of pages is not full enough for a keysheet, it seems some letters were missing. 
I am going to search for a better chapter to provide a baseline.

arkon:
wutface

morcefaster:
Added multi-chapter drifting. Fully works now.

PR #2475, opened 2020-03-22
Hive Works - Manga not found

### LINKED ISSUE #2465: [hiveworks] "Completed" comics do not appear in list
Tachi ver: 0.7.8
Extension ver: 1.2.3

https://hiveworkscomics.com/completed
Any existing comics in this section give an error shown in attached screenshot when attempting to load comic information
Screenshot_20200321-192613

happywillow0:
According to [kotlin reference](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/first.html), if no such element is found for the given predicate, a  `NoSuchElementException` exception is thrown.  This exception is now caught. 

- Fixes #2465 
- Adds Original, Kids, and Completed comic lists to filters. These comics however do not get "details" updates.

### DIFF
--- a/src/en/hiveworks/src/eu/kanade/tachiyomi/extension/en/hiveworks/Hiveworks.kt
+++ b/src/en/hiveworks/src/eu/kanade/tachiyomi/extension/en/hiveworks/Hiveworks.kt
@@ -163,13 +163,9 @@ class Hiveworks : ParsedHttpSource() {
     override fun mangaDetailsParse(document: Document): SManga = throw Exception("Not Used")
     private fun mangaDetailsParse(response: Response, url: String): SManga {
         val document = response.asJsoup()
-        return try {
-            document.select(popularMangaSelector())
-                .first { url == it.select("a.comiclink").first().attr("abs:href") }
-                .let { mangaFromElement(it) }
-        } catch (e: NoSuchElementException) {
-            SManga.create()
-        }
+        return document.select(popularMangaSelector())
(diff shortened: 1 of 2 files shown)


src/en/hiveworks/src/eu/kanade/tachiyomi/extension/en/hiveworks/Hiveworks.kt
```diff
@@ -152,10 +163,12 @@ class Hiveworks : ParsedHttpSource() {
     override fun mangaDetailsParse(document: Document): SManga = throw Exception("Not Used")
     private fun mangaDetailsParse(response: Response, url: String): SManga {
         val document = response.asJsoup()
-        return document.select(popularMangaSelector()).first {
-            url == it.select("a.comiclink").first().attr("abs:href")
-        }.let {
-            mangaFromElement(it)
+        return try {
+            document.select(popularMangaSelector())
+                .first { url == it.select("a.comiclink").first().attr("abs:href") }
```
CarlosEsco: what about firstOrNull{..}?.let

happywillow0: I actually had that first but wasn't sure if it was exception or null error. I'll try it out when I get home.

happywillow0: Thanks for the suggestion. I switched it over. 
Done

PR #2525, opened 2020-03-26
Update Kotlin and kotlinx.serialization version

### LINKED ISSUE #2512: [Bug] Manga Plus: Not loading on dev
### Device information
* Tachiyomi extension version: 1.2.4
* Tachiyomi version: dev r1940
* Android version: 6.0.1

## Steps to reproduce
Go to Catalogues and press Browse/Latest on Manga Plus.

OR

Update chapter list on a manga from Manga Plus.

OR

Read a chapter from Manga Plus.

### Expected behavior
Manga and chapters should load fine.

### Actual behavior
Everything related to Manga Plus loads indefinitely with no error messages.

### Other details
The extension works fine on stable v0.8.5, this issue is only present on dev.

alessandrojean:
Update Kotlin's version to match the version used in the app (1.3.70) and the kotlinx.serialization lib.

The compiled MangaPlus apk was tested in Android 4.1 (stable), Android 5.0, 6.0 and 10.0 (stable and dev) and it's working fine with the deserialization from kotlinx.serialization.

On older devices (Android <= KitKat) the extension only works with the stable version because of some unknown incompatibility with Duktape 1.3.0 and Kotlin 1.3.70 related to the `require` function. Since the dev version isn't compatible anymore with those devices, it doesn't see to be a huge problem.

Should close #2512.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,5 +1,5 @@
 buildscript {
-    ext.kotlin_version = '1.3.60'
+    ext.kotlin_version = '1.3.70'
     repositories {
         jcenter()
         google()
--- a/src/all/mangaplus/build.gradle
+++ b/src/all/mangaplus/build.gradle
@@ -6,12 +6,13 @@ ext {
     appName = 'Tachiyomi: MANGA Plus by SHUEISHA'
     pkgNameSuffix = 'all.mangaplus'
     extClass = '.MangaPlusFactory'
-    extVersionCode = 4
+    extVersionCode = 5
     libVersion = '1.2'
 }
 
 dependencies {
-    implementation 'org.jetbrains.kotlinx:kotlinx-serialization-runtime:0.14.0'
+    implementation 'org.jetbrains.kotlinx:kotlinx-serialization-protobuf:0.20.0'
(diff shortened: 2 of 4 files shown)


build.gradle
```diff
@@ -1,5 +1,5 @@
 buildscript {
-    ext.kotlin_version = '1.3.60'
+    ext.kotlin_version = '1.3.70'
```
arkon: Can we actually do a separate PR of just this change? I'm kind of worried since it does affect _all_ the extensions.

alessandrojean: When I was updating locally, just changing the Kotlin version have caused some issues and compile errors with the old version of kotlinx-serialization-runtime (0.14.0), mostly because it has a conflict with the stdlib version required by the lib and because the protobuf was now in an separate artifact.

So, just updating the Kotlin version will probably break the MangaPlus extension temporary, but I don't know if this local error will cause some general compile error on the repo, I can try.

arkon: Hmm you're right, I forgot they were so tightly coupled. That's fine then.

arkon:
I'll merge this sometime tomorrow since it's probably not a good idea to merge it right before going to bed in case something goes wrong. :P

alessandrojean:
No problem, it's a good choice actually.

PR #2529, opened 2020-03-27
Fixed title parsing for DisasterScans's mangaDetailsParse

SkiFire13:
Shouldn't really change nothing since the title _should_ already be fetched when opening the latest/popular/search page but better safe than sorry.

### DIFF
--- a/src/all/madara/build.gradle
+++ b/src/all/madara/build.gradle
@@ -5,7 +5,7 @@ ext {
     appName = 'Tachiyomi: Madara (multiple sources)'
     pkgNameSuffix = "all.madara"
     extClass = '.MadaraFactory'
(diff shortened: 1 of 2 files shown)


arkon:
I might've missed something, but what's the motivation for this change? Is it broken?

SkiFire13:
The default `mangaDetailsParse` looks for a `h3` tag for the title while DisasterScans uses a `h1`. This PR should fix it.

Is it broken? If the title is updated on the website, before it wouldn't be reflected on the app if the user refresh the details page, now it will.

I had a nasty bug where some DisasterScans's mangas had no title even through I refreshed the page so I looked into it and found this "bug".

PR #2545, opened 2020-03-29
Mangadex description cleanup

### LINKED ISSUE #2540: [Feature Request] Mangadex: Show description in language of Sytem locale only
Extension Version: 1.2.85

Currently, Mangadex shows manga description for all languages in the INFO tab, which makes it almost unreable for many manga. Requesting to show description for language of the system locale only, so that it's not a mess anymore.

CarlosEsco:
closes #2540
instead of the system language it matches the language of the extension you are using

ran ktlint on mangadex class
descriptions cleaned up depending on internal language

main code is cleanDescription.  Not pretty but it works

### DIFF
--- a/src/all/mangadex/src/eu/kanade/tachiyomi/extension/all/mangadex/Mangadex.kt
+++ b/src/all/mangadex/src/eu/kanade/tachiyomi/extension/all/mangadex/Mangadex.kt
@@ -33,6 +33,10 @@ abstract class Mangadex(
     private val internalLang: String
 ) : ConfigurableSource, ParsedHttpSource() {
 
+    init {
+
+    }
+
     override val name = "MangaDex"
 
     override val baseUrl = "https://mangadex.org"
@@ -43,6 +47,10 @@ abstract class Mangadex(
 
     private val preferences: SharedPreferences by lazy {
         Injekt.get<Application>().getSharedPreferences("source_$id", 0x0000)
+    }
+
+    private val mangadexDescription : MangadexDescription by lazy {
+        MangadexDescription()
     }
 
     private val rateLimitInterceptor = RateLimitInterceptor(4)
@@ -381,7 +389,7 @@ abstract class Mangadex(
         val chapterJson = json.getAsJsonObject("chapter")
         manga.title = cleanString(mangaJson.get("title").string)
         manga.thumbnail_url = cdnUrl + mangaJson.get("cover_url").string
-        manga.description = cleanDescription(mangaJson.get("description").string)
+        manga.description = cleanString(mangadexDescription.clean(internalLang, mangaJson.get("description").string))
         manga.author = mangaJson.get("author").string
         manga.artist = mangaJson.get("artist").string
         val status = mangaJson.get("status").int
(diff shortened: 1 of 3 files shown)


CarlosEsco:
I can remove the ktlint format if needed. It's set to auto format when I save using save actions plugin and it formatted all unintentionally

src/all/mangadex/src/eu/kanade/tachiyomi/extension/all/mangadex/Mangadex.kt
```diff
@@ -416,6 +468,58 @@ abstract class Mangadex(
...
+            "it" -> listOf("[b][u]Italian / Italiano[/u][/b]")
+            in "es", "mx" -> listOf("[b][u]Espa&ntilde;ol / Spanish:[/u][/b]")
+            in "br", "pt" -> listOf(
+                "[b][u]Portuguese (BR) / Portugu&ecirc;s (BR)[/u][/b]",
+                "[b][u]Português / Portuguese[/u][/b]",
+                "[b][u]Portuguese / Portugu[/u][/b]"
+            )
+            "tr" -> listOf("[b][u]Turkish / T&uuml;rk&ccedil;e[/u][/b]")
+            "fr" -> listOf(
+                "[b][u]Russian / Русский[/u][/b]",
```
fuzzylights: Duplicate from "ru"

src/all/mangadex/src/eu/kanade/tachiyomi/extension/all/mangadex/Mangadex.kt
```diff
@@ -416,6 +416,49 @@ abstract class Mangadex(
...
+            "fr" -> listOf("French - Français:","[b][u]French / Fran&ccedil;ais[/u][/b]")
+            "sa" -> listOf("[b][u]Arabic / العربية[/u][/b]")
+            else -> emptyList()
+
+        }
+        return cleanString(getCleanedDescription(description, listOfLangs))
+    }
+
+    /**
+     * this is where teh description really gets cleaned
```
arkon: the

src/all/mangadex/src/eu/kanade/tachiyomi/extension/all/mangadex/Mangadex.kt
```diff
@@ -868,6 +911,22 @@ abstract class Mangadex(
 
         const val PREFIX_ID_SEARCH = "id:"
 
+        private val DESCRIPTION_LANGUAGES = arrayListOf(
```
arkon: It'd be nice if there was a way that there's just a single mapping.

src/all/mangadex/src/eu/kanade/tachiyomi/extension/all/mangadex/Mangadex.kt
```diff
@@ -416,6 +416,49 @@ abstract class Mangadex(
...
+        originalString: String,
+        langTextToCheck: List<String>
+    ): String {
+        val langList = DESCRIPTION_LANGUAGES.toMutableList();
+
+        //remove any languages before the ones provided in the langTextToCheck, if no matches or empty
+        // just uses the original description, also removes the potential lang from all lang list
+        var newDescription = originalString;
+        langTextToCheck.forEach { it ->
+            newDescription = newDescription.substringAfter(it)
```
arkon: Can you remind me what this would do if the text isn't found?

src/all/mangadex/src/eu/kanade/tachiyomi/extension/all/mangadex/Mangadex.kt
CarlosEsco: Yea I couldn't think of an easy clean way. I guess I could do something enumish

src/all/mangadex/src/eu/kanade/tachiyomi/extension/all/mangadex/Mangadex.kt
CarlosEsco: It just return original string. So basically nothing happens if not found.

CarlosEsco:
@arkon  thoughts now?

arkon:
Ideally the enum would contain the lang code info too so it's all in one place.

You can reduce the cost of the lookup to just do it once as a lazily loaded class field or something, rather than repeatedly for every description.

CarlosEsco:
@arkon check now

src/all/mangadex/src/eu/kanade/tachiyomi/extension/all/mangadex/MangadexDescription.kt
```diff
@@ -0,0 +1,66 @@
...
+            "ru" -> RUSSIAN
+            "de" -> GERMAN
+            "it" -> ITALIAN
+            in "es", "mx" -> SPANISH
+            in "br", "pt" -> PORTUGESE
+            "tr" -> TURKISH
+            "fr" -> FRENCH
+            "sa" -> ARABIC
+            else -> emptyList()
+        }
```
arkon: I meant that this lookup could be done once lazily. You can pass in `internalLang` to the class and just init this right away.

Then this function would just have to take in the description rather than doing this lookup every time.

src/all/mangadex/src/eu/kanade/tachiyomi/extension/all/mangadex/MangadexDescription.kt
```diff
@@ -0,0 +1,66 @@
...
+            "[b][u]Portuguese (BR) / Portugu&ecirc;s (BR)[/u][/b]",
+            "[b][u]Português / Portuguese[/u][/b]",
+            "[b][u]Portuguese / Portugu[/u][/b]"
+        )
+        val RUSSIAN = listOf("[b][u]Russian / Русский[/u][/b]")
+        val SPANISH = listOf("[b][u]Espa&ntilde;ol / Spanish:[/u][/b]")
+        val TURKISH = listOf("[b][u]Turkish / T&uuml;rk&ccedil;e[/u][/b]")
+
+        val ALL_LANGS =
+            listOf(ARABIC, FRENCH, GERMAN, ITALIAN, PORTUGESE, RUSSIAN, SPANISH, TURKISH).flatten()
```
arkon: Probably could've just kept an enum and then have some field in it that gets all of the keys.

CarlosEsco: will keep how it is now since once class is created the list of all languages is now static.

CarlosEsco:
@arkon  one more time?

src/all/mangadex/src/eu/kanade/tachiyomi/extension/all/mangadex/Mangadex.kt
```diff
@@ -33,6 +33,10 @@ abstract class Mangadex(
     private val internalLang: String
 ) : ConfigurableSource, ParsedHttpSource() {
 
+    init {
```
arkon: Remove this.

PR #2588, opened 2020-04-04
Incorrect parsing of HTML entities in manga author and artists on Mangadex

Xefir:


arkon:
Update the version too.

PR #2614, opened 2020-04-08
Added www Mangadex subdomain to open directly

E3FxGaming:
While Google search results lead to mangadex.org websites,
other search engines such as Duckduckgo and Bing lead to
www.mangadex.org websites.
Changed extension manifest to handle www.mangadex.org


arkon:
Would've been nice to have handled the /manga/ route too.

E3FxGaming:
I didn't exactly know where the /manga route came from and thus without a URL to test with (+ a specific use case, like coming from web search results to the mangadex website) I didn't want to add anything unnecessary.

PR #2621, opened 2020-04-09
Added Komiku

AbdullahM0hamed:
Should close https://github.com/inorichi/tachiyomi-extensions/issues/2244

Needs to be opened in webview before use due to Cloudflare.


AbdullahM0hamed:
Not sure why old commits from my other branch show - but only the Komiku files show as changed.

src/id/komiku/src/eu/kanade/tachiyomi/extension/id/komiku/Komiku.kt
```diff
@@ -0,0 +1,116 @@
...
+
+    override val supportsLatest = true
+
+    override val client: OkHttpClient = network.cloudflareClient
+
+    override fun popularMangaSelector() = "div.bge"
+
+    override fun popularMangaRequest(page: Int) = GET("$baseUrl/other/hot/page/$page/?orderby=modified", headers)
+
+    override fun popularMangaFromElement(element: Element) = SManga.create().apply {
```
arkon, line 36: You should try to get the cover here if possible, otherwise it makes extra requests for every manga.

Same for latest/search.

AbdullahM0hamed: Done

PR #2675, opened 2020-04-14
Add Manhuaid.com

AbdullahM0hamed:
Closes https://github.com/inorichi/tachiyomi-extensions/issues/2666


src/id/manhuaid/src/eu/kanade/tachiyomi/extension/id/manhuaid/ManhuaID.kt
```diff
@@ -0,0 +1,97 @@
...
+import org.jsoup.select.Elements
+import java.text.SimpleDateFormat
+import java.util.Calendar
+import java.util.Locale
+
+class ManhuaID: ParsedHttpSource() {
+
+    override val name = "ManhuaID"
+
+    override val baseUrl = "http://manhuaid.com"
```
arkon, line 24: Use https

AbdullahM0hamed: My bad. Not sure why I had it like that - fixed now.

PR #2752, opened 2020-04-22
Added option in MangaPark to exclude missing chapter search

### LINKED ISSUE #2739: [Feature Request] MangaPark: Chapter List Source needs a flag to disable checking of missing chapters
### What feature should be added to this source?
Mangapark having the option to select the list source is great but I am noticing that it doesn't work correctly when the chapter names of the different sources are wildly different. On looking through the code, I have noticed that the extension tries to look for missing chapters from other sources even when we select a particular source or select "Source with most chapters". Imo this is slightly misleading and I found the actual working of this feature only on seeing the comments in the code. 

As such, a flag option to disable checking other sources would be greatly appreciated to be prevent mixing up of chapters due to varying chapter names. 

Let me know if this is something that I can take up.

SayakMukhopadhyay:
Fixes #2739

### DIFF
--- a/src/en/mangapark/src/eu/kanade/tachiyomi/extension/en/mangapark/MangaPark.kt
+++ b/src/en/mangapark/src/eu/kanade/tachiyomi/extension/en/mangapark/MangaPark.kt
@@ -118,8 +118,13 @@ class MangaPark : ConfigurableSource, ParsedHttpSource() {
 
     override fun chapterListParse(response: Response): List<SChapter> {
         fun List<SChapter>.getMissingChapters(allChapters: List<SChapter>): List<SChapter> {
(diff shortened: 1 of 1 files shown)


SayakMukhopadhyay:
Please note that I haven't updated the version yet. I am looking for feedback regarding version.

CarlosEsco:
version is always 1.2 then you increase the version code

SnakeDoc83:
Try version 1.2.15 of MangaPark. If you're still dissatisfied and think this option is necessary then let's finish this up and get it merged. Let me know what you'd like to do.

SayakMukhopadhyay:
Hey, sorry for the super late reply but the parsing seems to be great now! I believe I will close this PR for now. Thanks!

PR #2757, opened 2020-04-22
Fix "Page list empty" for KissManga

i404788:


arkon:
Update the version code in build.gradle

arkon:
Idk if this actually fixed it or not, but you'll still get errors if your IP is banned.

armangido:
> Idk if this actually fixed it or not, but you'll still get errors if your IP is banned.

they've unbanned ppls now, idk what is there reason

i404788:
As far as I've seen from the kissmanga source this is the only thing that needed to change for it to work.

i404788:
They just seem to have changed it again, it might randomize the suffix on a timer...

PR #2764, opened 2020-04-22
Update Kissmanga.kt

armangido:
Fixes the problem temporarily


arkon:
I'd rather you find an actual solution than constantly updating it like this.

armangido:
Temporarily merge it for now, If it changes I will create an solution, but for now merge and I will monitor if something happens.

armangido:
Something's wrong I need to check the code,ignore the revert changes

PR #2766, opened 2020-04-23
Minor update Kissmanga.kt

armangido:
fixed something.
As of now, this doesn't need constant value as long 'wrapKA' exists.
This doesn't fixed if your ban in KManga, but idk what happen they've unbanned me and some users for no reason?


armangido:
Kissmanga giving false ban to tachiyomi users while browser users seems working.

PR #2772, opened 2020-04-23
 Fix "Page List Empty" with Randomized Suffix (KissManga)

i404788:
The suffix hasn't changed since my last pull request so feel free to wait on this pull request.

I've tested it with https://gist.github.com/i404788/a1a53c1c2d269de989551515b44dfb94
The previous solutions didn't work with the test program, however this does seem to work.

### DIFF
--- a/src/en/kissmanga/build.gradle
+++ b/src/en/kissmanga/build.gradle
@@ -5,7 +5,7 @@ ext {
     appName = 'Tachiyomi: Kissmanga'
     pkgNameSuffix = 'en.kissmanga'
     extClass = '.Kissmanga'
-    extVersionCode = 16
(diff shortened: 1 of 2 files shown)


armangido:
This won't fix the cloudflare issue with kissmanga they seems giving Acesss Denied to Tachiyomi users.

i404788:
@armangido Indeed it doesn't, however I haven't seen that issue yet myself

SnakeDoc83:
As an FYI, they've been intentionally banning Tachi users all month.  If people are getting banned (access denied) right now it's more than likely a continuation of that policy.

armangido:
Seems issue is coming from the cloudflare libraries

PR #2776, opened 2020-04-23
JB page list fix.

### LINKED ISSUE #2768: [Bug] Foolslide (Jaimini's Box) : chapters do not load
Just opening this to track the issue; I'll try tackling de-obfuscating the chapter pages later tonight. 

It seems like JB's moved on from base64 encoding their manga pages; the code responsible is:
```javascript
(function (_0x390ee2, _0x332043) {
  var _0x5847de = function (_0x504e99) {
    while (--_0x504e99) {
      _0x390ee2["push"](_0x390ee2["shift"]());
    }
  };
  _0x5847de(++_0x332043);
})(_0x3320, 0x153);
var _0x5847 = function (_0x390ee2, _0x332043) {
  _0x390ee2 = _0x390ee2 - 0x0;
  var _0x5847de = _0x3320[_0x390ee2];
  return _0x5847de;
};
var pages = JSON["parse"](
  atob(
    _0x5847("0x0")["replace"](/[a-zA-Z]/g, function (_0x554ef5) {
      return String[_0x5847("0x1")](
        (_0x554ef5 <= "Z" ? 0x5a : 0x7a) >=
          (_0x554ef5 = _0x554ef5["charCodeAt"](0x0) + 0xd)
          ? _0x554ef5
          : _0x554ef5 - 0x1a
      );
    })
  )
);
```

funkyhippo:
Fixes #2768  

It seems like FoolSlide has a (rather barebones) API available, but it's mostly useful for chapter pages since you'd otherwise have to source an ID for the series. 

Technically, the overrides would work in the main `FoolSlide` class, but there are some edge cases where sites don't properly set their chapter numbers correctly, causing the call to fail. I haven't seen any chapters on JB that suffer from this, so I've kept the API usage localized to that extension.

### DIFF
--- a/src/all/foolslide/src/eu/kanade/tachiyomi/extension/all/foolslide/FoolSlideFactory.kt
+++ b/src/all/foolslide/src/eu/kanade/tachiyomi/extension/all/foolslide/FoolSlideFactory.kt
@@ -7,8 +7,10 @@ import eu.kanade.tachiyomi.network.GET
 import eu.kanade.tachiyomi.source.Source
 import eu.kanade.tachiyomi.source.SourceFactory
 import eu.kanade.tachiyomi.source.model.Page
+import eu.kanade.tachiyomi.source.model.SChapter
 import eu.kanade.tachiyomi.source.model.SManga
 import okhttp3.Request
+import org.json.JSONObject
 import org.jsoup.nodes.Document
 
 class FoolSlideFactory : SourceFactory {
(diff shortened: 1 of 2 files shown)


src/all/foolslide/src/eu/kanade/tachiyomi/extension/all/foolslide/FoolSlideFactory.kt
```diff
@@ -42,17 +44,20 @@ class FoolSlideFactory : SourceFactory {
 }
 
 class JaminisBox : FoolSlide("Jaimini's Box", "https://jaiminisbox.com", "en", "/reader") {
+    override fun pageListRequest(chapter: SChapter): Request {
+        val regex = "(?:/read/)([\\w\\d-]+?)(?:/)".toRegex()
```
arkon, line 48: Pull out commonly created regexes as class fields so they don't need to be constantly recreated.

funkyhippo: Done!

PR #2782, opened 2020-04-23
DbMultiverse - additional languages

SnakeDoc83:
Closes https://github.com/inorichi/tachiyomi-extensions/issues/2778

Since it's moving from "en" to "all" I changed the version to 1, let me know if I should bump it up instead.


CarlosEsco:
So doing it this way will cause anyone using it to have orphaned results. I think you need to do it like mangadex with the separate en folder

SnakeDoc83:
I'll change it, but could you explain why that'd be the case?  The source ID should remain the same for old users.

CarlosEsco:
Let me see if I can find it

PR #2810, opened 2020-04-25
MangaBox - add Mangakakalots

SnakeDoc83:
I'd been avoiding it because I thought it might cause confusion but I think at this point it's confusing people regardless.

Also:
* closes https://github.com/inorichi/tachiyomi-extensions/issues/2804
* scanlator field now shows chapter's host name; so if you're using kakalot but it's really hosted on nelo you'll see manganelo.com in the scanlator field.
* removed old Chapter Manga code (can always find it again later if necessary)


arkon:
Nice idea with the scanlator.

PR #2855, opened 2020-04-26
Fixed SeriManga search

AbdullahM0hamed:
Had the query hardcoded instead of using the query variable, my bad😅

Fixes https://github.com/inorichi/tachiyomi-extensions/issues/2853


Agayev033:
Thnx bro :D

PR #2951, opened 2020-05-01
Change as much HTML parsing to API as possible

buskun:


CarlosEsco:
update the build.gradle

buskun:
Sorry, I forgot to notify. I have updated the gradle file.

PR #3025, opened 2020-05-05
Update Gradles and Gson

TacoTheDank:
### 1st commit
- Update gradle wrapper to 6.3 using git bash

### 2nd commit
- Update Kotlin 1.3.70 -> 1.3.72
- Use and target JDK8

### 3th commit
- Update SDK license in travis
- Fix spacing


arkon:
Can you revert the RxJava and Gson changes? They need to stay on the older versions for compatibility reasons.

TacoTheDank:
Alrighty. What would those changes specifically affect with compatibility? I'm curious. Also, I'm going to update minSdk to 21, as that's the minimum version for the Tachiyomi app itself.

CarlosEsco:
Can't do that either. The reason libraries and min Android version have to stay same is they will break the 8.4/8.5 peoples ability to use extensions. For those  who are using older versions of the library and are not sdk  21

TacoTheDank:
@CarlosEsco Oof that's right.

TacoTheDank:
@arkon Would it be possible to fix the name of the commit, considering Gson wasn't changed?

arkon:
Too late! Doesn't really matter.

TacoTheDank:
Ok lol, thanks anyway.

PR #3032, opened 2020-05-06
ReadManhwa - add filters

SnakeDoc83:
Closes https://github.com/inorichi/tachiyomi-extensions/issues/3008

Webview won't fully load the details page, users have to use "open in browser" from the webview if they want to see the actual page. Browse was even worse; so I didn't bother keeping it in after I tested.

### DIFF
--- a/src/en/readmanhwa/src/eu/kanade/tachiyomi/extension/en/readmanhwa/ReadManhwa.kt
+++ b/src/en/readmanhwa/src/eu/kanade/tachiyomi/extension/en/readmanhwa/ReadManhwa.kt
@@ -19,6 +19,7 @@ import eu.kanade.tachiyomi.source.online.HttpSource
 import java.text.SimpleDateFormat
 import java.util.Calendar
 import java.util.Locale
+import okhttp3.HttpUrl
 import okhttp3.OkHttpClient
 import okhttp3.Request
 import okhttp3.Response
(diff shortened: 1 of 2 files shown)


src/en/readmanhwa/src/eu/kanade/tachiyomi/extension/en/readmanhwa/ReadManhwa.kt
```diff
@@ -70,14 +71,33 @@ class ReadManhwa : HttpSource() {
...
     override fun searchMangaRequest(page: Int, query: String, filters: FilterList): Request {
-        return GET("$baseUrl/api/comics?page=$page&q=$query&sort=uploaded_at&order=desc&duration=day", headers)
+        var url = "$baseUrl/api/comics"
+        if (query.isNotBlank()) {
+            url += "?page=$page&q=$query&sort=uploaded_at&order=desc&duration=day"
+        } else {
+            filters.forEach { filter ->
+                when (filter) {
+                    is GenreFilter -> url += "?per_page=18&tags=${filter.toUriPart()}&page=$page&q=&sort=popularity&order=desc"
+                    is DurationFilter -> url += "&duration=${filter.toUriPart()}"
```
arkon, line 81: What happens if there's only this present? Won't it be missing the `?`?

SnakeDoc83: Seemed fine when I tested it but I've changed it anyway.

PR #3060, opened 2020-05-07
NineManga - adjust headers

SnakeDoc83:
Closes https://github.com/inorichi/tachiyomi-extensions/issues/3046


earlan123:
Is the issue been fix?because i still can't read manga in ninemanga source

PR #3067, opened 2020-05-08
Adding a manga type for j2k library filters (to sources: comx, libmanga, mangachan, mintmanga, readmanga)

e-shl:


e-shl:
Apologize for the excess Commits

arkon:
Probably not the ideal solution in the longterm, but it's fine for now.

PR #3075, opened 2020-05-09
Correct a bunch of lints from "en" to "ja"

TacoTheDank:
...yeah

### DIFF
--- a/src/en/mangakatana/src/eu/kanade/tachiyomi/extension/en/mangakatana/MangaKatana.kt
+++ b/src/en/mangakatana/src/eu/kanade/tachiyomi/extension/en/mangakatana/MangaKatana.kt
@@ -2,18 +2,14 @@ package eu.kanade.tachiyomi.extension.en.mangakatana
 
 import eu.kanade.tachiyomi.network.GET
 import eu.kanade.tachiyomi.source.model.FilterList
-import eu.kanade.tachiyomi.source.model.MangasPage
 import eu.kanade.tachiyomi.source.model.Page
 import eu.kanade.tachiyomi.source.model.SChapter
 import eu.kanade.tachiyomi.source.model.SManga
 import eu.kanade.tachiyomi.source.online.ParsedHttpSource
-import eu.kanade.tachiyomi.util.asJsoup
-import java.text.SimpleDateFormat
-import java.util.Locale
-import okhttp3.OkHttpClient
-import okhttp3.Response
 import org.jsoup.nodes.Document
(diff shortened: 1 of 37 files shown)


src/en/mangakatana/src/eu/kanade/tachiyomi/extension/en/mangakatana/MangaKatana.kt
```diff
@@ -75,7 +69,7 @@ class MangaKatana: ParsedHttpSource() {
     override fun chapterFromElement(element: Element) = SChapter.create().apply {
         setUrlWithoutDomain(element.select("a").attr("href"))
         name = element.select("a").text()
-        date_upload = dateFormat.parse(element.select(".update_time").text()).time ?: 0
```
arkon, line 78: Why removing this? I don't think it fallsback normally.

TacoTheDank:
Oh btw, I caught [this](https://github.com/inorichi/tachiyomi-extensions/commit/dbcb0117cd5b63bba8b42f0c3def0b703c624834#diff-6a3371457528722a734f3c51d9238c13R64). The repo org is supposed to be tachiyomiorg :)

src/en/mangakatana/src/eu/kanade/tachiyomi/extension/en/mangakatana/MangaKatana.kt
TacoTheDank: According to Studio, the elvis operator always returns the left operand of the non-nullable type Long (perhaps that's not actually the case, Studio mistakes stuff every so often).

arkon: I'm actually not super sure what happens if it fails to parse properly here, presumably it throws an exception. I'd just leave it for now rather than changing the logic.

AbdullahM0hamed: I've just compiled it twice (with a deliberately messed up css selector), once with the elvis operator, and once without. Both times the an unparseable date error was thrown and chapters failed to load, so I'm not sure that it's a useful failsafe either.

TacoTheDank: I'll just leave it the way it was, then.

TacoTheDank:
Fixed conflict from BoredomSociety being deleted.

PR #3083, opened 2020-05-09
Add Inkr (all the code is from MR)

AbdullahM0hamed:
You will recognise the code - it's from the old MR extension.

Worked after some slight modifications (apiUrl and /pagesv3 instead of /pagesv2) 

Here's to a new era of inkr memes.


arkon:
Do the filters need to be updated?

AbdullahM0hamed:
Filters work. Rank doesn't atm though - should I remove that?

arkon:
Yeah, remove whatever doesn't work. Do the genres still work?

AbdullahM0hamed:
Genres work fine. Some of them return no results (which I've chalked down to it having a smaller library, but I'll guess I'll remove whatever returns nothing there too)

AbdullahM0hamed:
Filters have been updated.

PR #3084, opened 2020-05-09
Kissmanga Access Denied fix

armangido:
This fixes the error, if this is patched again, someone is deliberately patching this up, and I won't release anymore.


arkon:
The extension was already removed...

PR #3088, opened 2020-05-09
Myreadingmanga - add more languages

SnakeDoc83:
Closes https://github.com/inorichi/tachiyomi-extensions/issues/3034

I didn't want to lookup the language codes I wasn't sure about; so technically I didn't enable all languages. That said, I spoke with OP on discord and they're getting their language; so good enough for now is my opinion.


happywillow0:
Thanks for optimizing this.

PR #3156, opened 2020-05-14
[Italian Sources] Remove ItaScan and Updated MangaEden domain

LightDestory:
As the title says, Itascan is now closed.
Right now the extension returns a white page inside the application. I think it is possible to remove it from the extension catalogue.

Moreover, right now the MangaEden extension doesnt work due to a new domain: https://www.mangaeden.com/ is unreachable

New domain: https://www2.mangaeden.com/

Regards, LightDestory.


src/it/mangaeden/build.gradle
```diff
@@ -5,7 +5,7 @@ ext {
     appName = 'Tachiyomi: Mangaeden'
     pkgNameSuffix = 'it.mangaeden'
     extClass = '.Mangaeden'
-    extVersionCode = 4
+    extVersionCode = 4.1
```
arkon, line 8: Change this to an integer, i.e. `5`

LightDestory: Done :)

PR #3177, opened 2020-05-16
Manga Ship extension, DataImageInterceptor

SnakeDoc83:
Closes https://github.com/inorichi/tachiyomi-extensions/issues/2974

DataImageInterceptor can be implemented for other sources that use data:image strings for their images instead of URLs.


lib/dataimage/src/main/java/eu/kanade/tachiyomi/lib/dataimage/DataImageInterceptor.kt
```diff
@@ -0,0 +1,63 @@
...
+    override fun intercept(chain: Interceptor.Chain): Response {
+        val url = chain.request().url().toString()
+        return if (url.startsWith("https://127.0.0.1/?image")) {
+            val dataString = url.substringAfter("?")
+            val byteArray = if (dataString.contains("base64")) {
+                Base64.decode(dataString.substringAfter("base64,"), Base64.DEFAULT)
+            } else {
+                dataString.substringAfter(",").toByteArray()
+            }
+            val mediaType = MediaType.parse(Regex("""(^[^;,]*)[;,]""").find(dataString)!!.value)
```
arkon, line 52: Could maybe pull the regex out of the function so it doesn't need to be rebuilt on every call?

SnakeDoc83: Changed it.

PR #3191, opened 2020-05-16
Even more lint cleanup

TacoTheDank:
Yes.

Next up on my list is to go through every extension and update all their sources.


src/zh/hanhankuman/src/eu/kanade/tachiyomi/extension/zh/hanhankuman/HanhanKuman.kt
```diff
@@ -143,8 +143,8 @@ class HanhanKuman : ParsedHttpSource() {
         val su = "www.hhimm.com"
         var b = false
 
-        for (i in 0 until sw.split("|".toRegex()).dropLastWhile { it.isEmpty() }.toTypedArray().size) {
-            if (su.indexOf(sw.split("|".toRegex()).dropLastWhile { it.isEmpty() }.toTypedArray()[i]) > -1) {
+        for (element in sw.split("".toRegex()).dropLastWhile { it.isEmpty() }.toTypedArray()) {
+            if (su.indexOf(element) > -1) {
```
arkon, line 147: I'm not sure if this is correct, either before or after...

src/th/nekopost/src/eu/kanade/tachiyomi/extension/th/nekopost/Nekopost.kt
```diff
@@ -430,7 +430,7 @@ class Nekopost : ParsedHttpSource() {
         val genreList: Array<String> = try {
             (filters.find { filter -> filter is GenreFilter } as GenreFilter).state.filter { checkbox -> checkbox.state }.map { checkbox -> checkbox.name }.toTypedArray()
         } catch (e: Exception) {
-            emptyArray<String>()
+            emptyArray()
```
AbdullahM0hamed, line 433: This line seems to be causing the following error:

`Type inference failed: not enough information to infer parameter T in inline fun <reified T> emptyArray<T>
Please specify it explicitly.`

src/zh/hanhankuman/src/eu/kanade/tachiyomi/extension/zh/hanhankuman/HanhanKuman.kt
TacoTheDank: Hmmm. The lint says "For loop over indices could be replaced with loop over elements." I'll change it back regardless if you aren't sure.

src/th/nekopost/src/eu/kanade/tachiyomi/extension/th/nekopost/Nekopost.kt
TacoTheDank: Ah, yeah. Android Studio isn't good with that lint for some reason.

TacoTheDank:
Might be good now.

PR #3198, opened 2020-05-17
Add Madokami

### LINKED ISSUE #5: Madokami
Adopting this from https://github.com/inorichi/tachiyomi/issues/140.

tesu:
Resolves #5.

Reuses most of the code from #156, and adds PreferenceScreen support.

I've never written Kotlin before, so apologies in advance if the code quality is poor.


src/en/madokami/src/eu/kanade/tachiyomi/extension/en/madokami/Madokami.kt
```diff
@@ -0,0 +1,205 @@
...
+import uy.kohesive.injekt.Injekt
+import uy.kohesive.injekt.api.get
+
+class Madokami : ConfigurableSource, ParsedHttpSource() {
+    override val name = "Madokami"
+    override val baseUrl = "https://manga.madokami.al"
+    override val lang = "en"
+    override val supportsLatest = true
+
+    private val dateFormat = SimpleDateFormat("yyyy-MM-dd HH:mm")
```
arkon, line 32: Specify the locale too.

src/en/madokami/src/eu/kanade/tachiyomi/extension/en/madokami/Madokami.kt
```diff
@@ -0,0 +1,205 @@
...
+    }
+
+    override fun latestUpdatesSelector() = "table.mobile-files-table tbody tr td:nth-child(1) a:nth-child(1)"
+
+    override fun latestUpdatesFromElement(element: Element): SManga {
+        var manga = SManga.create()
+        manga.setUrlWithoutDomain(element.attr("href"))
+        manga.title = element.text().split("/").last()
+        return manga
+    }
```
arkon, line 50: Is it possible to get the thumbnail here too?

src/en/madokami/src/eu/kanade/tachiyomi/extension/en/madokami/Madokami.kt
```diff
@@ -0,0 +1,205 @@
...
+
+    override fun latestUpdatesNextPageSelector() = null
+
+    override fun latestUpdatesRequest(page: Int) = authenticate(GET("$baseUrl/recent", headers))
+
+    override fun popularMangaSelector(): String { return latestUpdatesSelector() }
```
arkon, line 56: Nitpick for this and other functions: you can shorten it to just `foo(): Type = bar()`

src/en/madokami/src/eu/kanade/tachiyomi/extension/en/madokami/Madokami.kt
```diff
@@ -0,0 +1,205 @@
...
+
+    /**
+     * Returns the details of the manga from the given [document].
+     *
+     * @param document the parsed document.
+     */
+    override fun mangaDetailsParse(document: Document): SManga {
+        var manga = SManga.create()
+        manga.url = document.location()
+        manga.title = document.select("span.title").text()
```
arkon, line 82: It's usually unnecessary to overwrite these at this point.

src/en/madokami/src/eu/kanade/tachiyomi/extension/en/madokami/Madokami.kt
tesu: Unfortunately not, the recent changes page is pretty barebones:

image

tesu:
Thanks for the prompt feedback, I think I've addressed all of your comments.

Couple of additional questions:
1. Madokami is more of an archive for downloading manga than an online manga reader, though it works fine for the latter purpose. My assumption is that when a user downloads a chapter on Tachiyomi, its images are scraped one by one (though I haven't poked around the main codebase to verify). Is there a way for extensions to specify an alternative chapter downloading method? In this case, we'd want to fetch the zip/cbz directly instead of asking the remote server to do the unpacking.

2. Madokami's directory structure is also quite different from a traditional manga reader website. There is manga where the directories `R/RA/RAND/RANDOM MANGA` and `R/RA/RAND/RANDOM MANGA/RANDOM MANGA [alt group]` both store copies of the same manga by different scanlators, but the extension currently reports them as separate manga. This isn't really a huge problem, since users probably are only interested in one or the other, and they can add both to their library if they're inclined.  
The bigger problem is that the manga details are only available from the `R/RA/RAND/RANDOM MANGA` page, and not the subdirectory's page, so `mangaDetailsParse` fails at getting any information about the manga. As a result, all of these manga directories are missing thumbnail images. I tried adding some code into `mangaDetailsParse` to check for this common scenario and return the results from `mangaDetailsParse` from the next directory up if information isn't available in the current directory, but I was only able to build a `Request` and couldn't figure out how execute it. Is there anything I'm missing that would make this possible?

arkon:
1. Tachiyomi only supports local archives.
2. You can try overriding `mangaDetailsRequest` instead?

src/en/madokami/src/eu/kanade/tachiyomi/extension/en/madokami/Madokami.kt
```diff
@@ -0,0 +1,202 @@
...
+import org.jsoup.nodes.Document
+import org.jsoup.nodes.Element
+import uy.kohesive.injekt.Injekt
+import uy.kohesive.injekt.api.get
+
+class Madokami : ConfigurableSource, ParsedHttpSource() {
+    override val name = "Madokami"
+    override val baseUrl = "https://manga.madokami.al"
+    override val lang = "en"
+    override val supportsLatest = true
```
SnakeDoc83, line 32: If popular is effectively going to be latest, I'd argue you should set this to false since the two views would be the same.

tesu: Fair enough. I was a little bit confused since Madokami supports latest and not most popular, but there isn't a supportsPopular val for me to set to false.

tesu:
> 1. Tachiyomi only supports local archives.

Right. I guess I was hoping for a way to repurpose the "download chapter" function into a "download archive, then move it into the proper local directory" function, but I can see why extensions wouldn't be able to specify that.

> 2. You can try overriding `mangaDetailsRequest` instead?

This worked. Thanks for pointing me in the right direction.

src/en/madokami/src/eu/kanade/tachiyomi/extension/en/madokami/Madokami.kt
SnakeDoc83: Yeah, even though the function names refer to "popular" they don't necessarily need to load popular manga (popular is preferred but isn't always an available sorting option from sources). It always has to load something, though; meanwhile, "latest" is optional.

src/en/madokami/build.gradle
```diff
@@ -0,0 +1,16 @@
...
+    appName = 'Tachiyomi: Madokami'
+    pkgNameSuffix = 'en.madokami'
+    extClass = '.Madokami'
+    extVersionCode = 1
+    libVersion = '1.2'
+}
+
+dependencies {
+    compileOnly 'com.github.inorichi.injekt:injekt-core:65b0440'
+}
```
arkon, line 14: I just made a change in `master` that moves the core app-provided dependencies up a level so it's available to all extensions without explicitly defining them here.

tl;dr: you should remove this.

src/en/madokami/src/eu/kanade/tachiyomi/extension/en/madokami/Madokami.kt
```diff
@@ -0,0 +1,210 @@
...
+import okhttp3.Request
+import org.jsoup.nodes.Document
+import org.jsoup.nodes.Element
+import uy.kohesive.injekt.Injekt
+import uy.kohesive.injekt.api.get
+
+class Madokami : ConfigurableSource, ParsedHttpSource() {
+    override val name = "Madokami"
+    override val baseUrl = "https://manga.madokami.al"
+    override val lang = "en"
```
arkon, line 31: One last thing -- is Madokami actually English only? If not, we can put it under `all`.

tesu: Maybe 1% of the content on there is non-English, tucked away in the Non-English directory. Is that enough to make it `all` instead of `en`?

tesu: Decided to change it to all, accounting for both the non-English translated stuff and the raws. However, I realized that I couldn't set this lang value to all and instead need a SourceFactory class. Would it be reasonable to create a set of sources for all the different languages, even if the sources all work the same? There's no language tags I can filter by in the search, so they'd have to be identical.

arkon: Having different language sources only really makes sense if they really can be filtered out, so I'd rather not do that.

src/all/madokami/src/eu/kanade/tachiyomi/extension/all/madokami/Madokami.kt
```diff
@@ -0,0 +1,213 @@
...
+import okhttp3.Request
+import org.jsoup.nodes.Document
+import org.jsoup.nodes.Element
+import uy.kohesive.injekt.Injekt
+import uy.kohesive.injekt.api.get
+
+class Madokami : ConfigurableSource, ParsedHttpSource() {
+    override val name = "Madokami"
+    override val baseUrl = "https://manga.madokami.al"
+    override val lang = "en"
```
arkon, line 31: This should be changed too.

tesu: After I changed this to `all`, the sources list became empty, so I brought up the possibility of adding a (bad) SourceFactory. Since I can't create separate sources that properly filter by language, and most of the content is English anyway, I think keeping it `en` is the cleanest choice.

arkon:
Congrats on closing the oldest open issue. :)

ObserverOfTime:
I don't think storing the password as plain text is a particularly good idea.
It should at least use a masked password field, if that's possible.

tesu:
I'd like to do that by setting the input type to be `InputType.TYPE_CLASS_TEXT or InputType.TYPE_TEXT_VARIATION_PASSWORD`, but it doesn't seem like inputType is exposed for Preferences in the extentions library.

PR #3220, opened 2020-05-17
TurkceManga Problem Solved

Agayev033:


AbdullahM0hamed:
Testing TurkceManga before and after adding a `www` prefix - and it doesn't seem to get rid of the 401 error in Tachiyomi itself, although the site appears in the webview (which it didn't seem to before)

Agayev033:
What can i di bro?
I am not good at programming.

SnakeDoc83:
@Agayev033 Search works; so I'm guessing you need to override `popularMangaRequest()` and `latestUpdatesRequest()`.

Agayev033:
Okey thnx

Agayev033:
Okey thnx

PR #3237, opened 2020-05-19
TurkceManga Problem Solved (Probably) and MangaPhoenix changed name and domain

Agayev033:


src/all/madara/src/eu/kanade/tachiyomi/extension/all/madara/MadaraFactory.kt
```diff
@@ -694,7 +694,7 @@ class ArangScans : Madara("Arang Scans", "https://www.arangscans.xyz", "en")
 
 class MangaHentai : Madara("Manga Hentai", "https://mangahentai.me", "en")
 
-class MangaPhoenix : Madara("Manga Phoenix", "https://mangaphoenix.com", "tr")
+class MangaPhoenix : Madara("Manga Diyarı", "https://mangadiyari.com", "tr")
```
SnakeDoc83, line 697: If you're changing the name, there's an `id` variable you should override as well.

Agayev033: Okey

PR #3272, opened 2020-05-22
[Mangadex] Add Data saver option to ext preferences

FlaminSarge:
Verified that the 'server' value that gets returned with 'saver=1' is `mangadex.org/data-saver/` and that it remains the default `mangadex.org/data/` with 'saver=0'


CarlosEsco:
Oh they changed it to a query parameter?

PR #3421, opened 2020-06-06
FMReader: remove manga title from chaper names

### LINKED ISSUE #3420: [Feature Request] FMReader: remove manga titles from chapter names
### What feature should be added to this source?
with this manga from LHT: https://lhtranslation.net/manga-shikkaku-mon-no-saikyou-kenja.html
one chapter name looks like this: 
`SHIKKAKU MON NO SAIKYOU KENJA - SEKAI SAIKYOU NO KENJA GA SARA NI TSUYOKUNARU TAME NI TENSEI SHIMASHITA Chapter 34`
the only real useful information is `Chapter 34`, but it's out of reach because the chapter name is so damn long that even in landscape mode you couldnt read it
this redundancy should be removed.

AriaMoradi:
closes #3420

### DIFF
--- a/src/all/fmreader/src/eu/kanade/tachiyomi/extension/all/fmreader/FMReader.kt
+++ b/src/all/fmreader/src/eu/kanade/tachiyomi/extension/all/fmreader/FMReader.kt
@@ -181,7 +181,7 @@ abstract class FMReader(
 
     // never called
     override fun chapterFromElement(element: Element): SChapter {
-        return SChapter.create()
+        throw Exception("unreachable code was reached, you get a 1-UP!")
     }
 
     override fun chapterListSelector() = "div#list-chapters p, table.table tr"
--- a/src/all/fmreader/src/eu/kanade/tachiyomi/extension/all/fmreader/FMReader.kt
+++ b/src/all/fmreader/src/eu/kanade/tachiyomi/extension/all/fmreader/FMReader.kt
@@ -1,5 +1,6 @@
 package eu.kanade.tachiyomi.extension.all.fmreader
 
+import android.util.Log
 import eu.kanade.tachiyomi.network.GET
 import eu.kanade.tachiyomi.source.model.Filter
 import eu.kanade.tachiyomi.source.model.FilterList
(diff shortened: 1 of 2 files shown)


SnakeDoc83:
Update build.gradle, remove log.

src/all/fmreader/src/eu/kanade/tachiyomi/extension/all/fmreader/FMReader.kt
```diff
@@ -101,7 +102,8 @@ abstract class FMReader(
         val mangas = document.select(popularMangaSelector()).map { popularMangaFromElement(it) }
 
         // check if there's a next page
-        val hasNextPage = (document.select(popularMangaNextPageSelector())?.first()?.text() ?: "").let {
+        val hasNextPage = (document.select(popularMangaNextPageSelector())?.first()?.text()
```
SnakeDoc83, line 105: Is this just a newline that's been added?

AriaMoradi: looks like reformating the file changed it

AriaMoradi:
changes done

SnakeDoc83:
Just tested this--it's going to need more work if you want this committed. As it is it breaks some of the sources.

AriaMoradi:
@SnakeDoc83 
I didn't understand how it breaks sources?
the only change is: if chapter name is present remove it, if not dont do anything
would you present a few examples for me to investigate?

AriaMoradi:
also I did a test:
first I downloaded a chapter with the current version of the code, then installed my patched one and the downloaded chapter was gone, so
1- if a source changes a chapter's name but not link, tachiyomi loses the data about it, so maybe this is a bug and should be fixed inside the app (only use the link as unique data?)
2- the same problem would exist with the same patch with my patch to `Paprika`

SnakeDoc83:
Take a look at FMReaderFactory.kt and you'll see some sources were overriding or relying on `chapterFromElement` as it was. I don't have Android Studio open at the moment, but as I recall at least 2 of the non-English sources wouldn't display chapter lists.

AriaMoradi:
> Take a look at FMReaderFactory.kt and you'll see some sources were overriding or relying on `chapterFromElement` as it was. I don't have Android Studio open at the moment, but as I recall at least 2 of the non-English sources wouldn't display chapter lists.

fixed

src/all/fmreader/src/eu/kanade/tachiyomi/extension/all/fmreader/FMReader.kt
SnakeDoc83: It's a nitpick but I'd prefer you return it to how it was for readability. Also remove the one commented out log.d and then I'll merge.

AriaMoradi: done and done

Issue #3423, opened 2020-06-06
[Meta] CONTRIBUTING.md: more sophisticated documentation


AriaMoradi:
###  Why/Project's Benefit/Existing Problem
- `Useful knowledge` is really poorly written and is missing some really vital information, yet:
   - it just throws some hints about things, I had to deliberately break existing code and debug the app to uncover some of the api
    -   some of the extension api is not explained at all, i.e. clinet , chapter, manga
    -   extension workflow(how and when methods are called) is not explained well, and only having stubs to work with doesnt hep at all
- there should be a general template for a extension, reading and copying other extensions is not really usefull 

## final thoughts 
I was able to contribute to `Paprika` because I had to just change a few links and selectors, if i where to write an extension from scratch it would never be possible

arkon:
A PR for updated documentation would be helpful.

AriaMoradi:
I'm going too collect some thoughts and information here, then maybe compose them in a meaningful  manner

AriaMoradi:
## general info
- the bridge between the app and you extension is the [extention-lib](https://github.com/inorichi/tachiyomi-extensions-lib), but it only contains stubs and the actual implementations are in the [app](https://github.com/inorichi/tachiyomi) inside `eu.kanade.tachiyomi.source` package, reading the code inside there will help you in writing your extension.
- you are encouraged to clone the app itself and make your own debug build so you can attach android-studio to the app, then you can print logs in the code and debug your code from inside `Logcat`, directly debugging your extension(steeping through your extension code) is not possible, but if you keep both projects open in android studio, you can debug the app itself.
- your `extClass`(inside `build.gradle`) class should be inherited from either `SourceFactory` or one of `Source` children: `CatalogueSource` or `HttpSource` or `ParsedHttpSource`. you mostly want to use one of the last two.
- `HttpSource` as in it's name is for a online http(s) source, but `ParsedHttpSource` has a good model of work which makes writing scrapers for normal aggregator websites much easier and streamlined. (you can find the implementation of the stubs in the app as mentioned above)  

## extension workflow

AriaMoradi:
## extension workflow

CarlosEsco:
nothing should be using CatalogueSource.  It should be ParsedHttpSource 99.9% of time

AriaMoradi:
> nothing should be using CatalogueSource. It should be ParsedHttpSource 99.9% of time
noted.

AriaMoradi:
[...]

PR #3443, opened 2020-06-08
improve CONTRIBUTING.md

AriaMoradi:
related to #3423


AriaMoradi:
about `Filter`, can anyone explain this to me?
the code is really messy there

CONTRIBUTING.md
```diff
@@ -85,33 +85,49 @@ dependencies {
 ```
 
 ### Useful knowledge
+- the bridge between the app and you extension is the [extention-lib](https://github.com/inorichi/tachiyomi-extensions-lib), but it only contains stubs and the actual implementations are in the [app](https://github.com/inorichi/tachiyomi) inside `eu.kanade.tachiyomi.source` package, reading the code inside there will help you in writing your extension.
```
arkon, line 88: - Capitalize the first word in the sentence.
- "extension-lib"
- Correct link: https://github.com/tachiyomiorg/extensions-lib
- Can just directly link to https://github.com/inorichi/tachiyomi/tree/dev/app/src/main/java/eu/kanade/tachiyomi/source

CONTRIBUTING.md
```diff
@@ -85,33 +85,49 @@ dependencies {
...
+- you are encouraged to clone the app itself and make your own debug build so you can attach android-studio to the app, then you can print logs in the code and debug your code from inside `Logcat`, directly debugging your extension(steeping through your extension code) is not possible, but if you keep both projects open in android studio, you can debug the app itself.
```
arkon, line 89: - "Android Studio"
- "from inside `Logcat`" -> "using Logcat" (can link to the documentation for that too)
- Make the debugging part a separate point. Also, "stepping".

CONTRIBUTING.md
```diff
@@ -85,33 +85,49 @@ dependencies {
...
+- your `extClass`(inside `build.gradle`) class should be inherited from either `SourceFactory` or one of `Source` children: `CatalogueSource` or `HttpSource` or `ParsedHttpSource`. you shouldn't inherit from `CatalogueSource` unless you know what you are doing.
```
arkon, line 90: I don't think there's really a valid normal use case for using `CatalogueSource` anyway. I don't think we need to include that here.

CONTRIBUTING.md
```diff
@@ -85,33 +85,49 @@ dependencies {
...
+- `HttpSource` as in it's name is for a online http(s) source, but `ParsedHttpSource` has a good model of work which makes writing scrapers for normal aggregator websites much easier and streamlined. (again, you can find the implementation of the stubs in the app as mentioned above)
```
arkon, line 91: Should probably just make this its own section.

The "Http" part of the name refers to the fact that it makes HTTP calls. http/https doesn't really matter.

CONTRIBUTING.md
```diff
@@ -85,33 +85,49 @@ dependencies {
...
 
-- An extension should at least extend the [`ParsedHttpSource`](https://github.com/inorichi/tachiyomi-extensions-lib/blob/master/library/src/main/java/eu/kanade/tachiyomi/source/online/ParsedHttpSource.kt) class.
-- Do not override id!!  It will auto-generate based on the name of the extension and the language.
-- Set the thumbnail cover when possible.  When parsing the list of manga during latest, search, browse.  If not the site will get a new request for every manga that doesn't have a cover shown,  even if the user doesnt click into the manga.
-
+### Important disclaimer before you continue!
+The structure for an extension is very strict.  In the future 1.x release this will be less strict but until then this has caused some issues when some sites don't quite fit the model.  There are required overrides but you can override the calling methods if you need more general control. This will go from the highest level method to the lowest level for browse/popular, it is the same but different method names for search and latest.
```
arkon, line 94: This doesn't matter for 0.x, could just remove this.

CONTRIBUTING.md
```diff
@@ -85,33 +85,49 @@ dependencies {
...
-   - jsoup CSS selector to see if there is a another page after current one.
-
- This will provide the initial viewing once a user clicks into a manga you will need to override mangaDetailsParse and this is where you need to parse the actual manga site's manga page and parse the standard info (title, author, description etc etc).
-
- ###### Note:
- Must be overriden even if you override the parent method and it's not being called anymore  (for example I override popularMangaParse and don't need the popularMangaNextPage selector.  I would just override in the extension and throw a not used exception).
+| Field | Description |
+| ----- | ----------- |
+| `name` | Name of the source as displayed in the `sources` tab inside the app |
+| `id` | id of your source, automatically set from `HttpSource`, don't touch it |
```
arkon, line 102: Should only be set if you need to copy an existing autogenerated ID.

CONTRIBUTING.md
```diff
@@ -85,33 +85,49 @@ dependencies {
...
-
- This will provide the initial viewing once a user clicks into a manga you will need to override mangaDetailsParse and this is where you need to parse the actual manga site's manga page and parse the standard info (title, author, description etc etc).
-
- ###### Note:
- Must be overriden even if you override the parent method and it's not being called anymore  (for example I override popularMangaParse and don't need the popularMangaNextPage selector.  I would just override in the extension and throw a not used exception).
+| Field | Description |
+| ----- | ----------- |
+| `name` | Name of the source as displayed in the `sources` tab inside the app |
+| `id` | id of your source, automatically set from `HttpSource`, don't touch it | 
+| `supportsLatest` | if `true` the app adds a `latest` button to your extension |
```
arkon, line 103: "to your source"

CONTRIBUTING.md
```diff
@@ -85,33 +85,49 @@ dependencies {
...
+| `baseUrl` | base URL of the target source without any trailing slashes |
```
arkon, line 104: I don't think the trailing slash matters?

CONTRIBUTING.md
```diff
@@ -85,33 +85,49 @@ dependencies {
...
+| `lang` | as the documentation says "An ISO 639-1 compliant language code (two letters in lower case).", it will be used to catalog you extension |
```
arkon, line 105: "your extension"

CONTRIBUTING.md
```diff
@@ -85,33 +85,49 @@ dependencies {
...
+ 
+- **Notes**
+    - some time during the code of you may find yourself finding no use for some inherited methods, if so just override them and throw exceptions: `throw Exception("Not used")`
```
arkon, line 108: The grammar here is pretty broken.

CONTRIBUTING.md
```diff
@@ -85,33 +85,49 @@ dependencies {
...
+    - you probably will find `getUrlWithoutDomain` useful when parsing the target source URLs.
+    - if possible try to stick to the general workflow from`ParsedHttpSource` and `HttpSource` breaking them may cause you more headache than necessary.
```
arkon, line 110: Huh?

CONTRIBUTING.md
```diff
@@ -85,33 +85,49 @@ dependencies {
...
+    -  when reading the code documentation it helps to follow the subsequent called methods in the the default implementation from the `app`, while trying to grasp the general workflow.
```
arkon, line 111: Maybe we should just have a diagram of the call flow?

CONTRIBUTING.md
AriaMoradi: corrected all of them.
btw, the same information is present under "Extensions library", which one do you think we should keep?

CONTRIBUTING.md
AriaMoradi: done.
also improved wording and moved to it's own separate section

CONTRIBUTING.md
AriaMoradi: yeah, I figured so. maybe it's my perfectionist mind...

CONTRIBUTING.md
AriaMoradi: it's trimmed so already is

all developers don't know how the HTTP protocol works

CONTRIBUTING.md
AriaMoradi: this was there before I added my edits, are you sure?

CONTRIBUTING.md
AriaMoradi: explain a bit more please

CONTRIBUTING.md
AriaMoradi: it does, because the code is using dumb `baseUrl` + something.url
so it must be explicit, I think it hast to be without.
can someone confirm this?

AriaMoradi: if we want to make it unimportant we have to change `HttpSource` implementations and use some url builder

CONTRIBUTING.md
AriaMoradi: fixed

CONTRIBUTING.md
AriaMoradi: yes, I changed it several times and it it came out broken

CONTRIBUTING.md
AriaMoradi: also there's similar info in my notes section

AriaMoradi: I think adding " This will go from the highest level method to the lowest level for browse/popular, it is the same but different method names for search and latest." will make everything cleaner

CONTRIBUTING.md
AriaMoradi: what's wrong?

CONTRIBUTING.md
AriaMoradi: if someone made a image for this that would be gud!
but for real, it's not really necessary, i think making first-time-extension-writers jump through the code and bang their head against the wall would be better for them

CONTRIBUTING.md
AriaMoradi: merged them and put it elsewhere

CONTRIBUTING.md
AriaMoradi: removed it

CONTRIBUTING.md
AriaMoradi: fixed

CONTRIBUTING.md
AriaMoradi: reworded it and its now fixed

CONTRIBUTING.md
AriaMoradi: fixed

AriaMoradi:
completely restructured most of the file, another review can be done now.

AriaMoradi:
Todo: ConfigurableSource

PR #3464, opened 2020-06-09
[Mangaeden] Fix images italian version

PiUliano:
Based on the english version, just changed links and variables to make it work with the italian verison


SnakeDoc83:
Bump the version in build.gradle.

Issue #3292, opened 2020-05-23
[Source Request] Manga Mutiny


Samk19:
### Source Website
- Name: Manga mutiny
- Link: https://mangamutiny.org/
- Language(s): English

IP555:
No thanks

E3FxGaming:
> No thanks

@IP555 May I ask what you're trying to say?

SnakeDoc83:
Closed via #3498 

@E3FxGaming Pretty sure that's a bot or compromised account.

Samk19:
@SnakeDoc83 is the source added ? cannot find it anywhere

Samk19:
Thanks 

So it a complete seprate app 

Thanks for working on it

PR #3498, opened 2020-06-11
Adding source Manga Mutiny

E3FxGaming:
Reason for source addition: source requested in issue #3292

What does work (tested it on a virtual device - I welcome any additional testers):
- Browse/Latest
- filtering (genre, status and type) / sorting
- reading manga
- manga details (author/artist/status/description/genres)
- downloading manga

Webview doesn't work. Reason: it's an HttpSource based on an undocumented API - it's not based on web scraping

Credit:
- the concept for the Filter was taken from the MangaPark extension and changed to fit the circumstances
- the JsonObject extension function stems from stackoverflow

The entire extension is based on an undocumented API which I discovered by analyzing the network traffic after using the website. The API is stable (I didn't encounter any unexpected behavior) and the data from the JSON files is way easier to process than any web scraping on that website (let's leave that at "this website uses few descriptive/useful css class names")


SnakeDoc83:
Regarding JSON, you might want to look in to using Kotson. What you've got here is fine, but for future projects you might find it easier to work with when parsing JSON.

E3FxGaming:
> you might want to look in to using Kotson

Ah ok, I'll keep that in mind. I have basic GSON knowledge from when I used it with Java and after switching to more and more Kotlin I made myself familiar with com.squareup.moshi:moshi - however that would be an additional dependency which I wanted to avoid. Kotson is already exposed to extensions through common.gradle - so that would definitely be an option.

I'm not familiar with Kotson yet and at some point one half of this extension still relied on terrible web scraping - then I noticed that I can realize the entire extension with the API. That's when my reliance on JSON grew, but the other half of the extension already used GSON so I stuck with that.

src/mangamutiny/src/eu/kanade/tachiyomi/extension/en/mangamutiny/MangaMutiny.kt
```diff
@@ -0,0 +1,391 @@
...
+    val value: JsonElement = this.get(key) ?: return null
+
+    if (value.isJsonNull) {
+        return null
+    }
+
+    return value
+}
+
+open class MangaMutiny : HttpSource() {
```
SnakeDoc83, line 34: Nitpick: class doesn't need to be open.

src/mangamutiny/src/eu/kanade/tachiyomi/extension/en/mangamutiny/MangaMutiny.kt
```diff
@@ -0,0 +1,391 @@
...
+            responseBody.close()
+        }
+
+        return MangasPage(mangasPage, mangasPage.size == 20)
+    }
+
+    private fun mangaRequest(page: Int, filters: FilterList? = null, query: String? = null): Request {
+        val uri = Uri.parse(baseUrl).buildUpon()
+        uri.appendEncodedPath(apiMangaUrlPath)
+        if ((query != null) && query.isNotEmpty()) {
```
SnakeDoc83, line 229: Could do `if (query?.isNotBlank() == true)`

E3FxGaming:
> It seems like the folder structure is wrong, /mangamutiny/ should be in /src/en/

I apologize for that. I had to reset my fork before asking for a merge, so I ended up backing all Manga Mutiny files up into a separate directory, resetting the fork and then something seems to have gone wrong during the restoration of my backed up files.

Should be fixed with my "MangaMutiny - multiple fixes" commit.

PR #3501, opened 2020-06-12
Fix libmanga chapters page parse

Pavkazzz:
subj


SnakeDoc83:
Update build.gradle.

Pavkazzz:
@SnakeDoc83 Sure, thx!

PR #3515, opened 2020-06-13
[CI SKIP] DataImage library info added to CONTRIBUTING.md

AriaMoradi:
also added grade attribute for other code tags


AriaMoradi:
is this link better?
https://css-tricks.com/data-uris/

arkon:
> is this link better?

I prefer more technical documentation. People can look up other references if they need to.

PR #3534, opened 2020-06-14
Minor fixes for MangaPlus extension

alessandrojean:
- Possible fix to missing `viewCount` field error, now it has a default value;
- Change the source name to match the official service name, overriding the `id`;
- Change the way that authors are displayed to match the latest UI change in the preview version;
- Add missing `nonAppearanceInfo` field to Duktape parsing.

### DIFF
--- a/src/all/mangaplus/src/eu/kanade/tachiyomi/extension/all/mangaplus/MangaPlus.kt
+++ b/src/all/mangaplus/src/eu/kanade/tachiyomi/extension/all/mangaplus/MangaPlus.kt
@@ -35,15 +35,11 @@ import uy.kohesive.injekt.Injekt
 import uy.kohesive.injekt.api.get
 
 abstract class MangaPlus(
+    override val id: Long,
     override val lang: String,
     private val internalLang: String,
     private val langCode: Language
 ) : HttpSource(), ConfigurableSource {
-
-    // Hardcode the id because the old name wasn't matching the official service name.
-    override val id: Long by lazy {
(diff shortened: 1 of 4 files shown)


src/all/mangaplus/src/eu/kanade/tachiyomi/extension/all/mangaplus/MangaPlus.kt
```diff
@@ -40,7 +40,12 @@ abstract class MangaPlus(
     private val langCode: Language
 ) : HttpSource(), ConfigurableSource {
 
-    override val name = "Manga Plus by Shueisha"
+    // Hardcode the id because the old name wasn't matching the official service name.
```
arkon, line 43: Couldn't you override the in the class instances in the factory?

alessandrojean: I can, will change for that.

SnakeDoc83: Also, if you're just changing capitalization I don't think it's necessary. The name is transformed to lower case when calculating the ID.

alessandrojean: Ooh, you are right, I will try to remove and see if it changes something.

alessandrojean:
Now it should be good to merge.

PR #3544, opened 2020-06-16
[CI SKIP] improvements to CONTRIBUTING.md

AriaMoradi:


AriaMoradi:
#3423 can be closed after this

AriaMoradi:
reading old commits I found this:

> `libVersion` is the version of the [extensions library](https://github.com/inorichi/tachiyomi-extensions-lib)* used. When this value is changed, `extVersionSuffix` should be reset to `1`. With the example used above, the version of the catalogue would be `1.0.1`.

still true?

AriaMoradi:
> You may find yourself needing additional functionality and wanting to add more dependencies to your build.gradle file. Since extensions are run within the main Tachiyomi app, you can make use of its dependencies.

>For example, an extension that needs Gson could add the following:

>dependencies {
    compileOnly 'com.google.code.gson:gson:2.8.2'
}
(Note that Gson, and several other dependencies, are already exposed to all extensions via common.gradle.)

>Notice that we're using compileOnly instead of implementation, since the app already contains it. You could use implementation instead for a new dependency, or you prefer not to rely on whatever the main app has at the expense of app size.

>Note that using compileOnly restricts you to versions that must be compatible with those used in Tachiyomi v0.8.5+ for proper backwards compatibility.

there are some problems with this section.
all dependencies are already added in `common.gradle` so `compileOnly`  has no meaning here.
and about `gson` it's already there, no point in mentioning it here.

I think the only useful part is the firs paragraph.

arkon:
Thanks for all of the edits. I ended up removing the bit about the Preview/debug builds since I'm pretty sure that's not true (correct me if you've seen otherwise).

> still true?

It'd make sense to, but we haven't had a library version bump in a while. We can revisit it whenever that happens.

AriaMoradi:
> I'm pretty sure that's not true (correct me if you've seen otherwise).
@arkon 
did lots of research about this, for a normal release build you have to set a flag in AndridManifest to make adb debug(and then Logcat) work

arkon:
But did you actually try? I'm pretty sure I used to do logs while developing extensions using the stable releases.

AriaMoradi:
https://developer.android.com/studio/publish/preparing#turn-off-logging-and-debugging
https://stackoverflow.com/questions/25610936/enable-logcat-on-release-build-in-android-studio

AriaMoradi:
android:debuggable
is not present in the manifest file, I checked

AriaMoradi:
when an app is run under "user" mode, not even adb can access it

AriaMoradi:
there are some ways to collect logs and stuff, but not through android studio

AriaMoradi:
@arkon

CarlosEsco:
its not true at all.  The extension is ran as debug so longs show fine.  You just have to target the non debug version of tachiyomi in the runner

AriaMoradi:
yeah, confirmed this myself
sorry to bother :/

PR #3569, opened 2020-06-19
MangaDex - fewer broken covers 

SnakeDoc83:
@CarlosEsco Do you think this is okay to do? I don't think the additional requests will be overly burdensome but I could be wrong.


CarlosEsco:
Since it doesn't actually change the cover url. I would just add the thumb.jpg if it's 404. So it just pulls the low quality covers which is for sure there. This could make 2 more additional requests vs 1.  Neko does something similar if when the images themselves load.

SnakeDoc83:
Makes sense. 👍

Issue #3535, opened 2020-06-14
[Source Request] Mangadeep


rro123:
Source Website
- Name: Mangadeep
- Link: www.mangadeep.com
- Language(s): English

E3FxGaming:
I have reason to believe that mangadeep is just a deceiving mirror of www.tenmanga.com, which was requested in #3275.
Whoever is mirroring the website did a poor job though, e.g.
- a link on mangadeep to a Facebook homepage still points to the Facebook page of tenmanga
- the ajax/javascript on mangadeep is pretty broken, while it works still fine on tenmanga.

(This may change now that I've pointed it out, but the similarities between the two websites probably won't change.)

@rro123 Is there a special reason for you wanting this source? Does this website have something other websites don't have?

rro123:
Among all mangareader sites that i follow up so far, i believe that
mangadeep is the one with the fastest update, at least in some manga such
as one piece. They post it at least a few hours (if not days) faster than
other sites. I haven't checked tenmanga yet, so i don't know whether it's
the same or not. Maybe i'll check them later.

The other reason is, in my country, many mangareader sites are blocked by
my ISP. Sites like mangafox, mangakakalot, and some others can't be
accessed. And so far, mangadeep is safe. I don't know about Tenmanga, but
if it's really the same as mangadeep, and it's not yet blocked by my ISP,
then i guess either of them is fine. Thanks for your attention to my
request.

Issue #3275, opened 2020-05-22
[Source Request] Tenmanga


Phenix-Feedback:
- Name: Tenmanga
- Link: http://www.tenmanga.com/
- Language(s): English

It has manga that other don't have. For example 14-sai no Koi up to date.

SnakeDoc83:
>It has manga that other don't have.

I shake my head sometimes when people comment this on requests. A quick google shows Mangakakalot has it, they even posted the most recent chapter 5 months before Tenmanga did. I'm not rejecting your request, but you might want to try some more extensions.

Phenix-Feedback:
> 
> 
> > It has manga that other don't have.
> 
> I shake my head sometimes when people comment this on requests. A quick google shows Mangakakalot has it, they even posted the most recent chapter 5 months before Tenmanga did. I'm not rejecting your request, but you might want to try some more extensions.

My bad, it showed chapter 49 and posted here before reading it, while it's chap 37.5. Not needed, sorry.

PR #3621, opened 2020-06-26
Adding source Tenmanga

E3FxGaming:
Requested in #3275 and brought up again in  #3535.
I'm not a particular fan of the Tenmanga website, but it's better than adding the broken copycat Mangadeep.


SnakeDoc83:
There's also http://www.manga99.com/, and I think all 3 sites are getting content from taadd.com. Taadd's design doesn't seem too far off from them either (at a cursory glance). Also, I think Taadd's content is already available by using the NineManga and NineAnime extensions. 

Taadd should be looked at, there's no open request for it but it's been mentioned before. Also need to see if there are any other sites using this theme (and if so, are they all clones which so far they seem to be). Basically, would like to decide whether this should be a factory extension before merging.

MerradOklom:
https://www.niadd.com/list/Hot-Book/
https://www.taadd.com/list/Hot-Book/
https://www.nineanime.com/list/Hot-Book/
These three sites seems to be sharing the same theme. As mentioned, NineAnime got an extension not long ago, so maybe it would be better (and probably less time consuming?) to tweak it a bit to make it a Factory instead of creating a new one. Also TenManga is using the same theme as NineManga, so maybe you could add it to that one and make it a multi-source extension?

E3FxGaming:
> would like to decide whether this should be a factory extension before merging

I would like to work on this factory, if you decide to do it like that.
I would just need to know what it should be called and on which existing extension or factory extension it should be based on.

Summarizing:

The factory should try to cover the following websites:
- http://www.tenmanga.com
- http://mangadeep.com/
- http://www.manga99.com
- https://www.taadd.com
- https://www.niadd.com
- https://www.nineanime.com
- http://en.ninemanga.com
- http://es.ninemanga.com
- http://br.ninemanga.com
- http://ru.ninemanga.com
- http://de.ninemanga.com
- http://it.ninemanga.com

src/en/tenmanga/build.gradle
```diff
@@ -0,0 +1,12 @@
+apply plugin: 'com.android.application'
+apply plugin: 'kotlin-android'
+
+ext {
+    extName = 'Tachiyomi: Tenmanga'
```
arkon, line 5: Don't need to have the `Tachiyomi: ` part anymore.

E3FxGaming:
I removed the "Tachiyomi:" from the extension name. Could someone take another look at this?

SnakeDoc83:
@E3FxGaming Sorry, tried to reply before but Github wouldn't post the message and I didn't circle back.

I think a factory extension might be the way to go and I would call it MangaYes. From what I can tell they may have been the progenitor of the theme (it originally looked like WieManga, which uses a file named mangayes.css). MangaYes currently forwards to Taadd. 

I would start with the sources below. Although NineManga may fit with them, you'd have to completely rewrite it to make the new sources fit in. I think it makes more sense to start with a cleaner base and not worry about NineManga for now. That said, feel free to look at NineManga, NineAnime, and WieManga for some potentially useful information.

http://www.tenmanga.com
http://mangadeep.com/
http://www.manga99.com
https://www.taadd.com
https://www.niadd.com

E3FxGaming:
Ok, then should I close this pull request and create a new pull request once I've got something to show?

AbdullahM0hamed:
You could do that or make this PR a draft until you have refactored it.

PR #3704, opened 2020-07-06
Update WPComicsFactory.kt

haruki-takeshi:
Add new source: TruyenChon


SnakeDoc83:
You needed to do a bit more. Anyway, I refactored some stuff and made things work.

PR #3733, opened 2020-07-09
Add NetTruyen source

TranHHoang:
Hi,
I found this app yesterday and it is amazing! However, some manga sources (some of my frequently visited sites) are missing, so I created extensions to address these sites.
Hope anyone who need it in the future find it useful.


arkon:
It's already in WP Comics though.

TranHHoang:
Oh I just check it out, it's another host of NetTruyen, but it's missing Filter feature. Don't know about it though, thanks.

SnakeDoc83:
NetTruyen is definitely in it and has filters, not to be confused with NhatTruyen which is the other host and doesn't have filters.

PR #3750, opened 2020-07-10
Add Silent Manga Audition source

alessandrojean:
Their website does not follow a pattern in the pages and a lot of the information are in images. 

Since they only add new entries one or two times a year, I think it's okay for now to store the information locally at the extension, updating only when needed.

If there's any better approach, just let me know. :)


SnakeDoc83:
I think using the "theme" that's in each manga's description as the manga title might be better than titling them "SMA-#". Are you sure you want to go with what you have?

alessandrojean:
Originally I was using the "SMA-# <theme>" format in the title, I will put it back in the title too.

alessandrojean:
@SnakeDoc83 My first attempt was creating a JSON with all the information in a private Gist and parse them in the extension, so when new entries are created, only the Gist has to be updated. Is this approach better?

SnakeDoc83:
I think if there are only going to be new entries once or twice a year it's probably better to do things as is.

Issue #3759, opened 2020-07-11
[Source Request][Spanish] Plot Twist No Fansub


Xtikma:
### Device information
* Tachiyomi version: 0.9.2
* Android version: 6.0
* Device: Samsung J7 (SM-700)

### Source information
* Source name: Madara (Plot Twist no Fansub)
* Source extension version: 1.2.116

## Steps to reproduce
1. Enter any manga
2. Error: No chapters found

## Issue/Request
Any manga, Error: No chapters found

## Other details
The domain they used in the extension is `www.plottwistscan.com`, this fansub also uses the following domain `www.plot-twistnf-scans.com` and apparently it is the one that receives updates and improvements

SnakeDoc83:
Their new site doesn't use Madara's them; so it's going to be removed from the Madara extension. Changing this issue to a source request.

Xtikma:
Sorry for the inconvenience, Thank you very much

PR #3793, opened 2020-07-15
Add Kumanga inorichi/tachiyomi-extensions#921

### LINKED ISSUE #921: [Source Request] [Spanish] Kumanga
URL:⬇️, 🙏

[Kumanga](http://www.kumanga.com)

edgarMejia:
- Closes #921


src/es/kumanga/src/eu/kanade/tachiyomi/extension/es/kumanga/Kumanga.kt
```diff
@@ -0,0 +1,268 @@
...
+            // Calculating total of pages, Kumanga shows 10 chapters per page, total_pages = #chapters / 10
+            val numberOfPages = (numberChapters / 10.toDouble() + 0.4).roundToInt()
+            var nextPage = if (numberOfPages > 1) 2 else 1
+
+            while (nextPage <= numberOfPages) {
+                document.select("div#accordion > div.panel.panel-default.c_panel").map {
+                    add(chapterFromElement(it))
+                }
+
+                nextPage++
```
SnakeDoc83, line 133: You're skipping page 2 when `nextPage` is 2 before the `while` loop starts.

edgarMejia: You're right, I'm uploading the change

src/es/kumanga/src/eu/kanade/tachiyomi/extension/es/kumanga/Kumanga.kt
```diff
@@ -0,0 +1,273 @@
...
+            val numberOfPages = (numberChapters / 10.toDouble() + 0.4).roundToInt()
+            var currentPage = 1
+
+            while (currentPage <= numberOfPages) {
+                document.select("div#accordion > div.panel.panel-default.c_panel").map {
+                    add(chapterFromElement(it))
+                }
+
+                currentPage++
+                document = client.newCall(GET(baseUrl + getMangaUrl(mangaId, mangaSlug, currentPage))).execute().asJsoup()
```
SnakeDoc83, line 140: Won't a second page always be requested now, even when there isn't one, because the `while` loop always runs at least once?

edgarMejia: Fixed

SnakeDoc83: I think this is okay, but I've had a couple cocktails; so, I'll come back to this in a while.

edgarMejia: hahahaha no problem, thanks for your help, I'll be watching for any other observations.

SnakeDoc83: It's still possible to make an extra request. Let's say there are 2 pages, when you go through the `while` loop to process page 2 you end up requesting page 3.  I think what you want to do is 

```
document.select(chapterSelector()).map { add(chapterFromElement(it)) }
page = 2
while (page < = numberOfPages) {
   document = client.newCall(GET(baseUrl + getMangaUrl(mangaId, mangaSlug, currentPage))).execute().asJsoup()
   document.select(chapterSelector()).map { add(chapterFromElement(it)) }
   page++
}
```

This way page 1 is always processed and outside of the loop and the loop

edgarMejia: Fixed

SnakeDoc83:
If you're looking for something to work on after this, there's a request for "Plot Twist No Fansub" (#3759). It used to be part of Madara but they recently changed websites; so it needs a new extension.

edgarMejia:
Sure, I'll work on it

Issue #3802, opened 2020-07-15
[Bug] ehentai: can't browse/search ehentai JP


PrivatePP:
You can't browse or search ehentai (jp), as tachiyomi gives you a blank page with no results found. Ehentai (eng) works just fine, but the japanese one only lets you browse latest.

I'm logged, I've tried to reinstall the extension and clear cookies/cache.

On a side note, I've heard there's a Exhentai extension but I can't seem to find it in the library... ?

---

### Device information
- Tachiyomi version: Stable v0.92
- Android version: 8

### Source information
- Name: ehentai
- Extension version: 1.2.12

SnakeDoc83:
If you want to use ExHentai, you need to use one of the forks (TachiyomiAZ or TachiyomiSY). https://tachiyomi.org/forks/

PR #3812, opened 2020-07-16
EHentai - add language preference

SnakeDoc83:
Closes https://github.com/inorichi/tachiyomi-extensions/issues/3802

Strict language enforcement is optional now, by default it's off. So while it's off, browse and search results should be what they were a version or two ago. It can be toggled in preferences if you want it always on/off or can be used as a filter in search if you want to change it for one-off searches\.

"Misc" tags are a more complete list now. Some tags may be ExHentai only, didn't bother filtering them out.


PrivatePP:
The bug #3802 persists if you enforce language while browsing E-Hentai (JP)

PR #3828, opened 2020-07-18
Comicpunch - fix chapters and pages not loading

pnivlek:
Comicpunch removed the "div#chapterlist" element from their pages, and didn't have a consistent replacement looking across a few comics, so I just took it out in favor of just using li.chapter.

For pages, they are now stored as an array of urls within a script tag instead of being all loaded by default like before.

### DIFF
--- a/src/en/comicpunch/src/eu/kanade/tachiyomi/extension/en/comicpunch/Comicpunch.kt
+++ b/src/en/comicpunch/src/eu/kanade/tachiyomi/extension/en/comicpunch/Comicpunch.kt
@@ -150,7 +150,7 @@ class Comicpunch : ParsedHttpSource() {
         // Then we just split it.
         val pageUrls = document.select("div#code_contain > script")
             .eq(1).first().data()
-            .substringAfter("= [").substringBefore("]").split("'")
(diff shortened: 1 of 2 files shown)


src/en/comicpunch/src/eu/kanade/tachiyomi/extension/en/comicpunch/Comicpunch.kt
```diff
@@ -138,14 +139,17 @@ class Comicpunch : ParsedHttpSource() {
     // Pages
 
     override fun pageListRequest(chapter: SChapter): Request {
```
SnakeDoc83, line 141: You can remove the whole override actually, this is what the function would normally be.

src/en/comicpunch/src/eu/kanade/tachiyomi/extension/en/comicpunch/Comicpunch.kt
```diff
@@ -1,5 +1,6 @@
 package eu.kanade.tachiyomi.extension.en.comicpunch
 
+import android.util.Log
```
SnakeDoc83, line 3: Can remove this import.

PR #3829, opened 2020-07-18
Comic latest

AbdullahM0hamed:
Fixes the search on ComicLatest as per https://github.com/inorichi/tachiyomi-extensions/pull/3790#issuecomment-660398990. Also, added the ability to search by author via a text filter. 

NOTE: author search must match the author's name exactly (case-insensitive).


src/all/wpcomics/src/eu/kanade/tachiyomi/extension/all/wpcomics/WPComicsFactory.kt
```diff
@@ -120,6 +120,25 @@ private class ComicLatest : WPComics("ComicLatest", "https://comiclatest.com", "
         thumbnail_url = element.select("img").attr("data-original")
     }
 
+    override fun searchMangaRequest(page: Int, query: String, filters: FilterList): Request {
+        if (query.startsWith("author:")) {
```
SnakeDoc83, line 124: Why not do a textfield filter instead?

AbdullahM0hamed: Forgot that existed, rip.

PR #3838, opened 2020-07-18
Move MangaZen to seperate extension

### LINKED ISSUE #3788: [Bug] <MangaZen>: <HTTP Error 400>
Please fix Arabic source mangaZen error 400 from Madara extension and thanks

AbdullahM0hamed:
Fixes #3788.

~~Note: The id was calculated (from the string "MangaZen/ar/1"), as I had no easy way to access MangaZen while it was broken. Would like @mekodke or someone else who used MangaZen to test (delete Madara temporarily, then install: https://github.com/AbdullahM0hamed/tachiyomi-extensions/raw/MangaZen-apk/apk/MangaZen-signed.apk) if possible.~~

EDIT: ID corrected to match output of Debug menu in TachiyomiSy. Apparently I should've calculated from "mangazen/ar/1" instead of "MangaZen/ar/1", rip.


SnakeDoc83:
Though the extension name is different, the source `name`, `lang`, and `versionId` are the same; so there's no need to hardcode the ID.

PR #3891, opened 2020-07-25
WP Mangastream - update Sekte Komik, KomikStation, KomikAV

kikuKeii:
Fix NextPageSelector and LastUpdate


SnakeDoc83:
I moved the next page selector to the main class because half the sources are using it.

Added MangaP to close https://github.com/inorichi/tachiyomi-extensions/issues/3874

PR #3919, opened 2020-07-27
Add Accept header to requests at MangasProject

alessandrojean:
Also added `Content-Length` and `Content-Type` to the search request.

### DIFF
--- a/src/pt/mangasproject/src/eu/kanade/tachiyomi/extension/pt/mangasproject/MangasProject.kt
+++ b/src/pt/mangasproject/src/eu/kanade/tachiyomi/extension/pt/mangasproject/MangasProject.kt
@@ -46,11 +46,12 @@ abstract class MangasProject(
         .build()
 
     override fun headersBuilder(): Headers.Builder = Headers.Builder()
-        .add("User-Agent", USER_AGENT)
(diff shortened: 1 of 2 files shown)


src/pt/mangasproject/src/eu/kanade/tachiyomi/extension/pt/mangasproject/MangasProject.kt
```diff
@@ -267,10 +263,9 @@ abstract class MangasProject(
 
         val document = result.asJsoup()
         val token = document.select("script[src*=\"reader.\"]").first()
-            ?.let {
-                HttpUrl.parse(it.attr("abs:src"))!!
-                    .queryParameter("token")
-            } ?: throw Exception("Não foi possível obter o token de leitura.")
+            .attr("abs:src")
```
SnakeDoc83, line 266: Pretty sure this won't work how you expect, either you'll get a string from `.attr()` or an error. You want to do `.firstOrNull()?.attr()`.

alessandrojean: Oh, you are right, I will change it.

Issue #3925, opened 2020-07-27
[Bug] <Manhuagui>: Can't open manga chapter caused by regexp error.


Oldwangtouchtouchdoge:
## Device information
- Tachiyomi version: 0.9.2
- Android version: 9 and 10 (Emulator)

### Source information
- Name: Manhuagui
- Extension version: 1.2.3

## Steps to reproduce
1. Open any manga's chapters page.
2. Open any unread chapter.

### Expected behavior
Chapter load normally.

### Actual behavior
Error pop up and can't open chapter: 
```
java.util.regex.PatternSyntaxException: Syntax error in regexp pattern near index 34
window\[".*?"](\(.*\)\s*\{[\s\S]+}\s*\(.*\))
                                  ^
```
And
```
java.util.regex.PatternSyntaxException: Syntax error in regexp pattern near index 5
\{.*}
     ^
```

![Snipaste_2020-07-27_20-41-49](https://user-images.githubusercontent.com/18094567/88552598-dbb41b00-d056-11ea-99a7-c14d2fd98c5b.png)

![图片](https://user-images.githubusercontent.com/18094567/88560018-24bc9d00-d060-11ea-9311-c79cd49f7c63.png)

![Screenshot_1595853592](https://user-images.githubusercontent.com/18094567/88552496-bf17e300-d056-11ea-832a-adebcce3c91e.png)

### Other details
After some test, I found its caused by removing the backslash to escape '\}', which Android studio suggest to remove them as redundant. Solved by add them back.

Oldwangtouchtouchdoge:
Redundant to #3924, sorry😂

Issue #3924, opened 2020-07-27
[Bug] <Manhuagui>: <regexp syntax error>


29988122:
### Device information
- Tachiyomi version: J2K latest version
- Android version: 8

### Source information
- Name: Manhuagui
- Extension version: 1.2.3

## Steps to reproduce
1. Open any manga
2. This toast message appears:
![2020-07-27 22 11 06](https://user-images.githubusercontent.com/5474709/88552451-b0313080-d056-11ea-8a05-74639a2fe10f.png)

It's the same for every manga.
IP not blocked, working normally in webview.

### Other details
Revert to version 1.2.2 doesn't work. This error only happened recently(<weeks or a month), so I assume that it's not because of this extension update.

Thank you, SnakeDoc83!

SnakeDoc83:
Closed by https://github.com/inorichi/tachiyomi-extensions/pull/3926

PR #3926, opened 2020-07-27
<Manhuagui>: <regexp syntax error> Add the backslashes to escape '\}' back.

Oldwangtouchtouchdoge:
Removing the backslash to escape '\\}', which Android studio suggest to remove them as redundant will cause regexp error. Details in #3925, #3924. The weird thing is I can't reproduce this error in kotlin running on Windows JDK 8 and 11.
Also, fixed manga.genre format in mangaDetailsParse() to display multiple genre separately.

### DIFF
--- a/src/zh/manhuagui/build.gradle
+++ b/src/zh/manhuagui/build.gradle
@@ -5,7 +5,7 @@ ext {
     extName = 'ManHuaGui'
     pkgNameSuffix = 'zh.manhuagui'
     extClass = '.Manhuagui'
-    extVersionCode = 3
+    extVersionCode = 4
     libVersion = '1.2'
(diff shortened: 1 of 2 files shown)


SnakeDoc83:
Weird, think it was fine when I tested on an AVD.

Oldwangtouchtouchdoge:
I found two typo in my commit message, sorry😅.

 \<Manhuagui\>: \<regexp syntax error\> Add the backslashes to escape '\}'…
… back. (#3926)
* Add the backslashes to escape '\}' back.
* Fix manga.genre format in mangaDetailsParse() to display multiple ❗~~multiple~~ genres❗ separately.

Issue #2701, opened 2020-04-18
[Source Request] Scantrad-Union (FR)


o01210o:
### Source Website
Website can be found at https://scantrad-union.com/projets/.

### Other Details
- scanlation group, not a "aggregator"
- to have a alternative to Japscan (because of recent instability)

![GIF2](https://user-images.githubusercontent.com/45663409/79669507-fc342580-8189-11ea-8c0f-beab57c5a160.gif)

SnakeDoc83:
Closed by https://github.com/inorichi/tachiyomi-extensions/pull/3945

PR #3945, opened 2020-07-29
Add Scantrad Union source (#2701)

Mursaat:
#2701


src/fr/scantradunion/src/eu/kanade/tachiyomi/extension/fr/scantradunion/ScantradUnion.kt
```diff
@@ -0,0 +1,214 @@
...
+        val title = document.select(".projet-description h2").text()
+        val url = document.location()
+        val thumbnailUrl = document.select(".projet-image img").attr("src")
+        val description = document.select(".sContent").text()
+        val statusStr = document.select(".label.label-primary")[2].text()
+        val detailElems = document.select(".project-details>a,.project-details>.divider2")
+
+        val status = mapMangaStatusStringToConst(statusStr)
+
+        // This loop build the list of authors.
```
SnakeDoc83, line 93: You can use the selector `div.project-details a[href*=auteur]` to get the author elements instead of doing a loop.

SnakeDoc83:
Also, you don't need to create separate vals for everything before you set them to other vals. It might make sense or be useful in the occasional place but you've done it for everything. So for example when you're setting `manga.thumbnail_url` set it right there; don't a separate val that you then set to `manga.thumbnail_url`.

Mursaat:
Initially I wanted to separate raw data retrieval, from the transformation and object creation from the model, but to go further I would have had to create other objects and mappers, and I thought that was probably too much.

I'll make the correction to simplify this.

Mursaat:
I made some modifications, please tell me if I forgot something :)

SnakeDoc83:
Looked good so I gave it a test run. Popular showed only ~5 manga; should maybe parse https://scantrad-union.com/projets/ for popular? It wouldn't "technically" be what's popular on the site; but it would be a fuller catalog.

You don't usually need to set `chapter_number`. You can leave it if you want, but if you end up doing any more extensions you don't have to spend time on it unless you need to use it for something.

Mursaat:
> Looked good so I gave it a test run. Popular showed only ~5 manga; should maybe parse https://scantrad-union.com/projets/ for popular? It wouldn't "technically" be what's popular on the site; but it would be a fuller catalog.

Yes the section "Séries en avant" is pretty short. OK I will use https://scantrad-union.com/projets/

> You don't usually need to set chapter_number. You can leave it if you want, but if you end up doing any more extensions you don't have to spend time on it unless you need to use it for something.

I remove the assignment then. I thought this property acted as a rank for chapter sorting.

SnakeDoc83:
> I thought this property acted as a rank for chapter sorting.

It can be. By default, Tachiyomi displays chapters in the order that they've been parsed; so if a website lists their chapters from 10 down to 1, Tachiyomi will display them from 10 down to 1.  Sometimes a source will mess up the chapter order completely--in that case a user may opt to sort by chapter number, which normally Tachiyomi will determine on its own. That said, when we're making an extension and if we see that the chapter order is messed up we can fix it right there by setting `chapter_number` and sorting the chapter list we return removing the need for the user to take the action of changing their settings. So you're partly right, but it's something we normally do only when the default order isn't good enough.

PR #3960, opened 2020-08-01
Fix unacessible chapters showing when the manga is licensed at LeitorNet

alessandrojean:
The check if the manga is licensed prevents the chapter list API from being called when it's not supposed, simulating the behavior of the website.

### DIFF
--- a/src/pt/mangasproject/build.gradle
+++ b/src/pt/mangasproject/build.gradle
@@ -5,7 +5,7 @@ ext {
     extName = 'mangásPROJECT'
     pkgNameSuffix = 'pt.mangasproject'
     extClass = '.MangasProjectFactory'
-    extVersionCode = 11
+    extVersionCode = 12
     libVersion = '1.2'
 }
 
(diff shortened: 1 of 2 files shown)


SnakeDoc83:
Tachiyomi isn't supposed to request the chapter list of a manga that's been set to licensed. You're saying they're still being requested and this would prevent that, right?

alessandrojean:
> You're saying they're still being requested and this would prevent that (parsing them, anyway), right?

Exactly. What happens is that apparently the manga details and the chapter list request are called concurrently, and since the status of the manga isn't determined yet, there is no way that the app knows if it is licensed or not to prevent the chapter list request from being called.

This PR adds a request to the manga page before the chapter list request, and if the manga is licensed, it prevents the chapter list requests from being called.

Their API returns the all available chapters even if the manga is licensed (because the LeitorNet and MangaLivre are "unified" -- they share the same database), but in LeitorNet if you try to open a chapter from a licensed manga, it doesn't work.

In the way that is right now, the app loads at the first time the chapters of any licensed manga, and if the user try to read them, it will give an error. With this PR, the chapters of a licensed manga are not shown, so there is no way to the user access them at LeitorNet (in MangaLivre it works fine).

SnakeDoc83:
Should probably bring this up in the main repo; I think I've noticed this before with a different extension.

Issue #4050, opened 2020-08-08
[Bug] ReadManhwa : NSFWs just vanished


leeyan69:
### Device information
- Tachiyomi version: App version: 0.10.3 (standard, 0c07e05a2, 49)
- Android version: 5.0.2 (SDK 21)

### Source information
- Name: ReadManhwa
- Extension version: 1.2.5

## Steps to reproduce
1. First Step
2. Second Step

### Expected behavior
Should show all Manhwas (including NSFWs)

### Actual behavior
Suddenly not showing just the NSFWs manhwas others are working well

### Other details
 unable to see any NSFW manhwa in the ReadManhwa extension. I went through all the troubleshooting steps reinstalled and all to no avail. Show NSFW is checked and when filtering by adult it shows no results found. Everything is working well om webview and NSFWs are also showing on it but not on tachiyomi itself.

SnakeDoc83:
Closed by https://github.com/inorichi/tachiyomi-extensions/pull/4051

PR #4051, opened 2020-08-08
nsfw true added in query param

balagurusurendar:
#4050 
added nsfw=true in query param

### DIFF
--- a/src/en/readmanhwa/build.gradle
+++ b/src/en/readmanhwa/build.gradle
@@ -5,7 +5,7 @@ ext {
     extName = 'ReadManhwa'
     pkgNameSuffix = 'en.readmanhwa'
     extClass = '.ReadManhwa'
-    extVersionCode = 5
+    extVersionCode = 6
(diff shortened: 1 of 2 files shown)


CarlosEsco:
Should probably be an option?  To disable or enable it

balagurusurendar:
yes, it needs to be an option
sorry I am not well versed in android & kotlin
for now, to make it work I added as query param by default.
in case you have a better idea you can go ahead & drop this pull & add your own change.
@CarlosEsco

CarlosEsco:
If you at one of the other extensions with options like mangadex. You can see how the preference is made and used

balagurusurendar:
I will check it out @CarlosEsco 
thank you for pointing it out.

balagurusurendar:
@CarlosEsco 
added preference, please check it out

CarlosEsco:
Looks good you need to update the build.gradle

balagurusurendar:
version changed

PR #4054, opened 2020-08-08
madara: Add manga.detectiveconanar.com

yshalsager:
Added https://www.manga.detectiveconanar.com/ which have all detective Conan different manga chapters in Arabic.


yshalsager:
screenshot-20200808-221007
screenshot-20200808-221015

Issue #4055, opened 2020-08-08
[Meta] Flag NSFW extensions as such


arkon:
Followup to https://github.com/inorichi/tachiyomi/issues/1312

1. Extensions containing NSFW sources should be marked as `containsNsfw = true` in their `build.gradle` files.
2. Individual sources should be annotated with `@Nsfw`

PR #4057, opened 2020-08-08
Mark extensions as nsfw

AbdullahM0hamed:
Attempts to address #4055. Probably missing many more extensions that should be marked as nsfw, but it's a start.

### DIFF
--- a/src/all/ehentai/build.gradle
+++ b/src/all/ehentai/build.gradle
@@ -5,8 +5,9 @@ ext {
     extName = 'E-Hentai'
     pkgNameSuffix = 'all.ehentai'
     extClass = '.EHFactory'
-    extVersionCode = 13
+    extVersionCode = 14
     libVersion = '1.2'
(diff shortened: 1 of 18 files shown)


alessandrojean:
Can you add the flag to the SuperMangás and HiperCool extension too? They are in the Portuguese folder.

joseph619:
Does FMReader have explicit hentai sources? Afaik, most aggregators have some NSFW content so this could potentially affect most extensions. I wonder if we should limit this to hentai majority sources only?

arkon:
You don't need to actually bump the version numbers.

jobobby04:
> Does FMReader have explicit hentai sources? Afaik, most aggregators have some NSFW content so this could potentially affect most extensions. I wonder if we should limit this to hentai majority sources only?

Manhwa18

AbdullahM0hamed:
Updated

PR #4060, opened 2020-08-09
Update NANI? Scans extension to use their API

Dinokin:


SnakeDoc83:
I see that they changed domains, but is there a significant need to switch to parsing from their API vs HTML? Switching is going to break any favorited manga users have as well as their chapter lists.

Dinokin:
Well the API should be more consistent since I don't plan to change anything there anytime soon. It also supports stuff like search by name, tag and id which the site doesn't support yet. As for the lost chapter list and favorites, it would break anyway because I plan to change some routes on that area soon.

SnakeDoc83:
Ah, I see. I'd like to request 2 changes then:
* set `thumbnail_url` on `popularMangaParse()` and `latestUpdatesParse()` (in addition to where you already set it)
* one instance of SimpleDateFormat to use as a date parser

Dinokin:
Done o7

SnakeDoc83:
Thanks for the update.

Note: I bumped `versionId` since this won't be compatible with old favorites.

PR #4081, opened 2020-08-11
Add the rest of the languages per ISO-639

einstein95:


src/all/hitomi/src/eu/kanade/tachiyomi/extension/all/hitomi/HitomiFactory.kt
```diff
@@ -7,49 +7,44 @@ import eu.kanade.tachiyomi.source.SourceFactory
...
+    Pair("no", "norwegian"),
     Pair("pl", "polish"),
-    Pair("pt-BR", "portuguese"),
-    Pair("", "romanian"),
-    Pair("", "albanian"),
-    Pair("", "slovak"),
-    Pair("", "finnish"),
-    Pair("", "swedish"),
-    Pair("", "tagalog"),
+    Pair("pt", "portuguese"),
```
alessandrojean, line 32: The Portuguese language code here should be `pt-BR` as it was before because it's the variation from Brazil.

PR #4110, opened 2020-08-13
Fix Plot Twist no Fansub, Update LectorManga, TMOHentai and TMO

### LINKED ISSUE #4108: [Bug] Kumanga: Error 400
## Device information
- Tachiyomi version: Stable 0.10.4
- Android version: 10

## Source information
- Name: Kumanga
- Extension version: 1.2.5

## Steps to reproduce
1. Browse all mangas

### Expected behavior
Get list of all mangas

### Actual behavior
Error 400

## Other details
N/A

### LINKED ISSUE #4109: [Bug] Plot Twist no Fansub: Latest url return 404
## Device information
- Tachiyomi version: Stable 0.10.4
- Android version: 10

## Source information
- Name: Plot Twist No Fansub
- Extension version: 1.2.1

## Steps to reproduce
1. Browse latest

### Expected behavior
Get list of latest mangas

### Actual behavior
Error 404

## Other details
N/A

edgarMejia:
## Plot Twist no Fansub
- [x] Closes #4109 

## TMO Sources
- [x] LectorManga: Add support to opening links from the website in main app
- [x] TMO: Add support to opening links from the website in main app
- [x] TMOHentai: Add support to opening links from the website in main app

### DIFF
--- a/src/es/kumanga/src/eu/kanade/tachiyomi/extension/es/kumanga/Kumanga.kt
+++ b/src/es/kumanga/src/eu/kanade/tachiyomi/extension/es/kumanga/Kumanga.kt
@@ -61,15 +61,13 @@ class Kumanga : HttpSource() {
         .build()
 
     private var kumangaToken = ""
-
-    // They add various fake tokens but the chosen one always starts with "eyJjdCI6I"
-    private val tokenRegex = Regex(""""(eyJjdCI6I[^"\s]{100,})"""")
+    private val tokenRegex = Regex(""""([^"\s]{100,})"""")
 
     private fun getKumangaToken() {
         kumangaToken = client.newCall(GET("$baseUrl/mangalist?&page=1", headers)).execute().asJsoup()
             .select("div.input-group [type=hidden]")
             .firstOrNull()
-            ?.let { tokenRegex.find(it.outerHtml())!!.value }
+            ?.let { tokenRegex.find(it.outerHtml())?.groupValues?.get(1) }
             ?: throw IOException("No fue posible obtener la lista de mangas")
     }
 
--- a/src/es/kumanga/src/eu/kanade/tachiyomi/extension/es/kumanga/Kumanga.kt
+++ b/src/es/kumanga/src/eu/kanade/tachiyomi/extension/es/kumanga/Kumanga.kt
@@ -142,7 +142,7 @@ class Kumanga : HttpSource() {
             name = it.text()
             date_upload = parseChapterDate(it.attr("title"))
         }
-            scanlator = element.select("span.pull-right.greenSpan")?.text()
+        scanlator = element.select("span.pull-right.greenSpan")?.text()
     }
 
     override fun chapterListParse(response: Response): List<SChapter> = mutableListOf<SChapter>().apply {
--- a/src/es/kumanga/src/eu/kanade/tachiyomi/extension/es/kumanga/Kumanga.kt
+++ b/src/es/kumanga/src/eu/kanade/tachiyomi/extension/es/kumanga/Kumanga.kt
@@ -61,13 +61,15 @@ class Kumanga : HttpSource() {
         .build()
 
     private var kumangaToken = ""
-    private val tokenRegex = Regex(""""([^"\s]{100,})"""")
+
+    // They add various fake tokens but the chosen one always starts with "eyJjdCI6I"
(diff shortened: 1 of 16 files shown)


src/es/kumanga/src/eu/kanade/tachiyomi/extension/es/kumanga/Kumanga.kt
```diff
@@ -61,13 +61,15 @@ class Kumanga : HttpSource() {
...
+
+    // They add various fake tokens but the chosen one always starts with "eyJjdCI6I"
+    private val tokenRegex = Regex(""""(eyJjdCI6I[^"\s]{100,})"""")
 
     private fun getKumangaToken() {
         kumangaToken = client.newCall(GET("$baseUrl/mangalist?&page=1", headers)).execute().asJsoup()
             .select("div.input-group [type=hidden]")
             .firstOrNull()
-            ?.let { tokenRegex.find(it.outerHtml())?.groupValues?.get(1) }
+            ?.let { tokenRegex.find(it.outerHtml())!!.value }
```
SnakeDoc83, line 72: `?` instead of `!!`

SnakeDoc83: And did you test this? I haven't run it, I'm a little worried about changing to `.value`

edgarMejia: Yes I test it with `?.groupValues?.get(1)` and `!!.value` and works with both, but I change it to `!!.value` because I think `?.groupValues?.get(1)` is used when regex returns multiple results and `!!.value` for one like this case.

jobobby04: Value is used to return the whole match, while group values are used to get the matches in the brackets(there can be multiple), it seems the whole regex is in brackets, which is why it could do that before

SnakeDoc83: `.groupValues?.get(1)` is being used to return what the regex captures in the first capture group (a capture group being what's inside `()`), `.value` returns the whole match. With this regex there should be a difference in results, one would return `token_characters` the other returning `"token_characters"` (those quotation marks could be a problem).

edgarMejia: Sorry, this is my first time using regex with Kotlin, so in this case the best option is `.groupValues?.get(1)`?

SnakeDoc83: In this case, yes, and it should be called with `?`, not `!!` because if `find` returns null we'll get an NPE and that could crash the app here.

edgarMejia: Thanks for the help to both

edgarMejia: While I made the correction, Kumanga uploaded another update, so my change is no longer functional.

AbdullahM0hamed: Plot Twist no Fansub, more like Plot Twist: the changes are no longer functional

edgarMejia: hahaha

edgarMejia:
Hi,
is there a way to remove the Kumanga changes from this pull request?

I can't find a solution to the encryption of the images in the pUrl variable.

AbdullahM0hamed:
Yeah, remove the changes locally, commit the removal of said changes, and then push the commit.

edgarMejia:
> Yeah, remove the changes locally, commit the removal of said changes, and then push the commit.

Ty 🙏🏼

edgarMejia:
Ready to merge.

SnakeDoc83:
Think you missed updating a gradle.

edgarMejia:
> Think you missed updating a gradle.

Thanks 😓

PR #4127, opened 2020-08-14
Fix Epikmanga

### LINKED ISSUE #4126: [Bug] <Epik Manga >: <I can't see other manga because of bug>
---

## Device information
- Tachiyomi version: Stable 0.10.4
- Android version: Android 10 Q
## Source information
- Name: FMReader (Epik Manga)
- Extension version: 1.2.20

## Steps to reproduce
1. First Step : Browse the EpikManga catalogue and as you can see there is bug. You will see that there are hundreds of just one manga.
2. Second Step : When you select any of them and press any of the sections you see, you will see that it is not opened.

### Expected behavior
This should have happened show all the manga on the site and we should be able to see the chapters in manga

### Actual behavior
This happen instead,shows a single manga on the site, and most parts of that manga are missing.

## Other details

video

AbdullahM0hamed:
Fixes #4126

### DIFF
--- a/src/all/fmreader/build.gradle
+++ b/src/all/fmreader/build.gradle
@@ -5,7 +5,7 @@ ext {
     extName = 'FMReader (multiple aggregators)'
     pkgNameSuffix = 'all.fmreader'
     extClass = '.FMReaderFactory'
-    extVersionCode = 20
+    extVersionCode = 21
     libVersion = '1.2'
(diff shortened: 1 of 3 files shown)


AbdullahM0hamed:
Duck. I was supposed to make an epik pun based on the name. Somehow that slipped my mind.

PR #4133, opened 2020-08-15
Refined codes and added new features

calion-m:
When I had written my extension of 18comic, I found someone did the same thing, just 12h ahead of me. So I decided to merge my codes, which contained some new features.

### DIFF
--- a/src/zh/jinmantiantang/src/eu/kanade/tachiyomi/extension/zh/jinmantiantang/Jinmantiantang.kt
+++ b/src/zh/jinmantiantang/src/eu/kanade/tachiyomi/extension/zh/jinmantiantang/Jinmantiantang.kt
@@ -261,10 +261,10 @@ class Jinmantiantang : ParsedHttpSource() {
     ))
 
     private class SortFilter : UriPartFilter("排序", arrayOf(
-        Pair("最新", "o=mr&"),
         Pair("最多订阅", "o=mv&"),
-        Pair("最多图片", "o=mp&"),
-        Pair("最多爱心", "o=tf&")
+        Pair("最新", "o=mr&"),
+        Pair("最多爱心", "o=tf&"),
+        Pair("最多图片", "o=mp&")
     ))
 
     private class TimeFilter : UriPartFilter("时间", arrayOf(
--- a/src/zh/jinmantiantang/src/eu/kanade/tachiyomi/extension/zh/jinmantiantang/Jinmantiantang.kt
+++ b/src/zh/jinmantiantang/src/eu/kanade/tachiyomi/extension/zh/jinmantiantang/Jinmantiantang.kt
@@ -193,8 +193,8 @@ class Jinmantiantang : ParsedHttpSource() {
         Pair("百合", "/search/photos?search_query=百合&"),
         Pair("男同", "/search/photos?search_query=YAOI&"),
         Pair("性转换", "/search/photos?search_query=性轉換&"),
-        Pair("暴力", "/search/photos?search_query=NTR&"),
-        Pair("偽娘", "/search/photos?search_query=偽娘&"),
+        Pair("NTR", "/search/photos?search_query=NTR&"),
+        Pair("伪娘", "/search/photos?search_query=偽娘&"),
         Pair("痴女", "/search/photos?search_query=癡女&"),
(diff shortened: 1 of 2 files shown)


src/zh/jinmantiantang/src/eu/kanade/tachiyomi/extension/zh/jinmantiantang/Jinmantiantang.kt
```diff
@@ -71,13 +92,16 @@ class Jinmantiantang : ParsedHttpSource() {
 
     override fun mangaDetailsParse(document: Document): SManga = SManga.create().apply {
         determineChapterInfo(document)
-        title = document.select("div.panel-heading").select("div.pull-left").first().text()
-        thumbnail_url = document.select("img.lazy_img.img-responsive").attr("src").split("\\?")[0]
+        // title = document.select("div.panel-heading").select("div.pull-left").first().text()
+        // thumbnail_url = document.select("img.lazy_img.img-responsive").attr("src").split("\\?")[0]
```
SnakeDoc83, line 96: These shouldn't be commented out.

src/zh/jinmantiantang/src/eu/kanade/tachiyomi/extension/zh/jinmantiantang/Jinmantiantang.kt
```diff
@@ -47,19 +50,37 @@ class Jinmantiantang : ParsedHttpSource() {
...
+            } else {
+                // 在搜索栏的关键词前添加-号来实现对筛选结果的过滤, 像 "-YAOI -扶他 -毛絨絨 -獵奇", 注意此时搜索功能不可用.
+                val removedGenres = query.split(" ").filter { it.startsWith("-") }.map { it.removePrefix("-") }.joinToString("+")
+                HttpUrl.parse("$baseUrl$params&page=$page&screen=$defaultRemovedGenres$removedGenres")?.newBuilder()
+            }
         }
+        return GET(url.toString(), headers)
     }
+    // 默认过滤类型, 仅针对能够自己编译应用的读者
+    private val defaultRemovedGenres: String = "" // like ”YAOI+扶他+毛絨絨+獵奇+“
```
SnakeDoc83, line 83: Since it's always an empty string what's its use?

src/zh/jinmantiantang/src/eu/kanade/tachiyomi/extension/zh/jinmantiantang/Jinmantiantang.kt
AbdullahM0hamed: They are what I like to call: an illusion.

calion-m: > These shouldn't be commented out.

Actually, It still works fine, and manga can directly inherit thumbnail from main page. 
The reason I comment them is that refreshing thumbnail is really annoying(the two thumbnail look same, but their urls are different, which means cache will not take effect).

src/zh/jinmantiantang/src/eu/kanade/tachiyomi/extension/zh/jinmantiantang/Jinmantiantang.kt
calion-m: For now, it is for the people who can compile extension themselves to remove the genres they dislike. For future considerations, users may set it in the extension configuration.

src/zh/jinmantiantang/src/eu/kanade/tachiyomi/extension/zh/jinmantiantang/Jinmantiantang.kt
SnakeDoc83: >Actually, It still works fine

That's not the point. With it commented out users won't be able to refresh their covers if the site changes them and if a user needs to restore a backup they won't get any covers at all. This won't be merged with it commented out.

ZongerZY:
He did a great job, because it was the first time I was engaged in the development of extensions. Many codes were not standardized. I learned a lot from his modifications. It was great. He is a very good programmer.

PR #4138, opened 2020-08-15
Dmzj - update page parsing

SnakeDoc83:
Closes https://github.com/inorichi/tachiyomi-extensions/issues/4132


pzy2222:
@SnakeDoc83 A large number of manga are hidden on the website due to government's bans, which makes reading them impossible on tachiyomi now. Maybe there is a method for us to load images from the PC website when they're accessible on the web, and to use v3api to read those being hidden on the web? (Sorry for my poor English. And I don't know whether you can see this comment under my closed bug issue so I make a copy here.)

PR #4211, opened 2020-08-26
[Komga] filter by genres, tags and publishers

gotson:


gotson:
The build failure doesn't seem to be related to my changes.

jobobby04:
Jitpack is down, so any dependency pulling from it doesn't work currently

PR #4214, opened 2020-08-26
mangakiss.org site no longer exist

shashank-p:


ANETAIII:
mangakiss.org site back online please add this site back to madara

ANETAIII:
What about mangakiss.org? It's not exist on tachiyomi

scb261:
@ANETAIII create new issue so your request won't be lost

PR #4260, opened 2020-08-30
[Ru] New source. Yaoi-chan.me

Pavkazzz:


src/ru/yaoichan/src/eu/kanade/tachiyomi/extension/ru/yaoichan/Yaoichan.kt
```diff
@@ -0,0 +1,311 @@
...
+
+    override fun searchMangaParse(response: Response): MangasPage {
+        val document = response.asJsoup()
+        var hasNextPage = false
+
+        val mangas = document.select(searchMangaSelector()).map { element ->
+            searchMangaFromElement(element)
+        }
+
+        val nextSearchPage = document.select(searchMangaNextPageSelector())
```
SnakeDoc83, line 157: What are you trying to accomplish from here to line 171? It seems like an overly complicated way to get `hasNextPage`, am I not understanding something?

src/ru/yaoichan/src/eu/kanade/tachiyomi/extension/ru/yaoichan/Yaoichan.kt
```diff
@@ -0,0 +1,311 @@
...
+        .addNetworkInterceptor(rateLimitInterceptor).build()
+
+    override fun popularMangaRequest(page: Int): Request =
+        GET("$baseUrl/mostfavorites?offset=${20 * (page - 1)}", headers)
+
+    override fun searchMangaRequest(page: Int, query: String, filters: FilterList): Request {
+        var pageNum = 1
+        when {
+            page < 1 -> pageNum = 1
+            page >= 1 -> pageNum = page
```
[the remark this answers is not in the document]
Pavkazzz: I've mostly copy-paste from other source, so you are right here. Refactor too: 
```
        var pageNum = when {
            page < 1 -> 1
            else -> page
        }
```

src/ru/yaoichan/src/eu/kanade/tachiyomi/extension/ru/yaoichan/Yaoichan.kt
Pavkazzz: Yes, it's looks like we have many way to get next page, for some reason. Pretty sure it's very rearly used part of code, will fix it if getting some real trouble with it

src/ru/yaoichan/src/eu/kanade/tachiyomi/extension/ru/yaoichan/Yaoichan.kt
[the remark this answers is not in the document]
SnakeDoc83: `page` is never less than one though, right?

src/ru/yaoichan/src/eu/kanade/tachiyomi/extension/ru/yaoichan/Yaoichan.kt
SnakeDoc83: I can sort of guess where you've got this from and what the original intent was, but I don't think it's needed. I just took a look at the website and I would think that you don't actually need to override `searchMangaParse()` at all, but I haven't tested.

Pavkazzz: You are absolutely right. Remove this override

src/ru/yaoichan/src/eu/kanade/tachiyomi/extension/ru/yaoichan/Yaoichan.kt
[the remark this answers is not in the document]
Pavkazzz: Yep. Remove

src/ru/yaoichan/src/eu/kanade/tachiyomi/extension/ru/yaoichan/Yaoichan.kt
Pavkazzz: Thx for review, I realy appreciate this

PR #4261, opened 2020-08-30
[Jinmantiantang] Added preference screen to select server location

rawfh:


src/zh/jinmantiantang/src/eu/kanade/tachiyomi/extension/zh/jinmantiantang/Jinmantiantang.kt
```diff
@@ -198,6 +207,46 @@ class Jinmantiantang : ParsedHttpSource() {
...
+            key = "SERVER_LOCATION"
+            title = "服务器位置"
+            entries = arrayOf("大陆", "海外")
+            entryValues = arrayOf("https://18comic2.biz", "https://18comic.org")
+            setDefaultValue(entryValues[0])
+            setOnPreferenceChangeListener { _, newValue ->
+                val selected = newValue as String
+                val index = this.findIndexOfValue(selected)
+                val entry = entryValues[index] as String
+                baseUrl = entry
```
SnakeDoc83, line 225: This is flawed, when a user restarts Tachiyomi `baseUrl` will go back to being the default.

PR #4276, opened 2020-08-31
Add EliMangas

TechnoJo4:


AbdullahM0hamed:
Yeet the original japscan while you're at it

AbdullahM0hamed:
More pain than it's worth

Bamaraj09:
What a goat

PR #4289, opened 2020-09-02
[Ru] Libmanga. Revert changes from 19->20 update

Pavkazzz:
mangalib.me disable download chapters for unauth users. Revert last update


Pavkazzz:
@SnakeDoc83 Merge pls

PR #4290, opened 2020-09-02
Update Newtoki Extension v1.2.14

DitFranXX:
v1.2.14
=====
* Site name/domain changes
  - ManaMoa merged into NewToki's manga side. calls ManaToki.
  - Site wasn't not changed too hard so decided to keep old site ids.
* Fix issue with pages with optional \<p\> tags
* Search Filter on ManaToki side
  - The site doesn't support title search with filter. It runs separately  

----

nah idea's new ui is too complicated to me

The site is very unstable on title name search now.


DitFranXX:
It's ready to merge.

PR #4298, opened 2020-09-02
Fix Golden Manga due to site redesign

### LINKED ISSUE #4297: Golden Mangás extension not working
The following error appears when trying to read a manga from that source:
Attempt to invoke virtual method 'org.jsoup.select.Elements org.jsoup.nodes.Element.select(java.lang.String)' on a null object reference

alessandrojean:
Closes #4297.

### DIFF
--- a/src/pt/goldenmangas/src/eu/kanade/tachiyomi/extension/pt/goldenmangas/GoldenMangas.kt
+++ b/src/pt/goldenmangas/src/eu/kanade/tachiyomi/extension/pt/goldenmangas/GoldenMangas.kt
@@ -6,6 +6,7 @@ import eu.kanade.tachiyomi.source.model.Page
 import eu.kanade.tachiyomi.source.model.SChapter
 import eu.kanade.tachiyomi.source.model.SManga
 import eu.kanade.tachiyomi.source.online.ParsedHttpSource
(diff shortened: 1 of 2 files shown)


src/pt/goldenmangas/src/eu/kanade/tachiyomi/extension/pt/goldenmangas/GoldenMangas.kt
```diff
@@ -45,12 +47,15 @@ class GoldenMangas : ParsedHttpSource() {
...
 
     override fun popularMangaFromElement(element: Element): SManga = SManga.create().apply {
-        title = element.select("h3").first().text().withoutLanguage()
-        thumbnail_url = element.select("img").first()?.attr("abs:src")
-        url = element.attr("href")
+        title = element.select("img").attr("alt")
+            .substringAfter("online ")
+            .withoutLanguage()
+        thumbnail_url = element.select("img").attr("abs:data-src")
+        url = "/" + element.attr("href")
```
SnakeDoc83, line 58: This is fine, but I'll mention that a good alternative to this is `setUrlWithoutDomain("abs:href")` (using `abs` will get you the preceding `/`.

PR #4371, opened 2020-09-14
MangaFast: ignore unparseable dates

### LINKED ISSUE #4346: [Bug] <Mangafast>: <Unparseable date>
#3793  Device information
- Tachiyomi version: j2k 1.0.8
- Android version: 8.1

## Source information
- Name: Mangafast
- Extension version: 1.2.2

## Steps to reproduce
1. First Step. open any manga from magafast
2. Second Step try refresh for new charapter and get msg error. Unparseable date "Scheduled PDF"

### Expected behavior
add new chatapters to the list

### Actual behavior
get error msg and miss new charapters

## Other details

Screenshot_20200910-035618_TachiyomiJ2K

druppelt:
Fixes #4346

MangaFast introduced the concept of teaser chapters and placeholders for upcoming chapters. These do not have a proper date and previously led to a ParseException.

Ideally the extension would mark those chapters as teasers and in case of placeholders just skip them, but I couldn't think of a way to do so.

### DIFF
--- a/src/en/mangafast/build.gradle
+++ b/src/en/mangafast/build.gradle
@@ -5,7 +5,7 @@ ext {
     extName = 'MangaFast'
     pkgNameSuffix = 'en.mangafast'
     extClass = '.MangaFast'
-    extVersionCode = 2
+    extVersionCode = 3
     libVersion = '1.2'
 }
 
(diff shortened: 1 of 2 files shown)


arkon:
I think you'd want to try to ignore those entries by somehow removing them from `chapterListSelector` instead.

druppelt:
As I'm not familiar with the selectors offered by JSoup I thought that was impossible. But it is possible, so thanks for the suggestion, `tr:has(td.tgs:matches(\d{4}-\d{2}-\d{2}))` does the trick. Should I remove the catch block for ParseExceptions?

arkon:
You could keep it on the off chance that the selector doesn't work as intended, but you'll just end up with chapter entries that probably wouldn't work properly (since it wouldn't be able to get actual pages).

Remember to bump the version number in the extension's `build.gradle` file too.

druppelt:
Then I'll keep it. At least for the placeholder chapters I tested I didn't run into any issues getting the pages, it's just that in absence of any teaser pages MangaFast presents you the previous chapter. Useless, but not harmful.

arkon:
Thanks!

PR #4445, opened 2020-09-27
Fix typo for WP MangaStream 

scb261:
Not sure if I should change extension's version, since I didn't actually change anything inside of it. And I didn't build/test it, but this shouldn't break anything, I think.


arkon:
You don't need to bump the version.

PR #4467, opened 2020-09-30
Fix for thumbnails loading

LordUTF8:


arkon:
Update the version number too.

Issue #4388, opened 2020-09-17
[Bug] <WNACG>: <image won't load in main>


kschou95:
![Screenshot_20200917-223307](https://user-images.githubusercontent.com/21261240/93485959-3884d100-f936-11ea-8e46-fa4aa6c69c5a.jpg)

## Device information
- Tachiyomi version:  stable 0.10.5
- Android version: 10

## Source information
- Name: WNACG
- Extension version: 1.2.2

## Steps to reproduce
1. just open the extension

### Expected behavior
manga cover should appear

### Actual behavior
as in screenshot

## Other details
going to web browser mode seems that wnacg site is working finr, so its probably the extension's problem

PR #4489, opened 2020-10-01
Fix wnacg thumbnails loading

youyeyuki:
fix issues #4388


src/zh/wnacg/src/eu/kanade/tachiyomi/extension/zh/wnacg/wnacg.kt
```diff
@@ -89,7 +103,16 @@ class wnacg : ParsedHttpSource() {
         }
         return pages
     }
+
     override fun chapterFromElement(element: Element) = throw Exception("Not used")
     override fun imageUrlRequest(page: Page) = throw Exception("Not used")
     override fun imageUrlParse(document: Document) = throw Exception("Not used")
+    private fun debug(msg: String?, pre_fix: String = "") {
```
CarlosEsco, line 110: remove debug

PR #4491, opened 2020-10-02
Update MangaDoom.kt

LeMyst:
Change manga order (Top is recent)
Use chapter number from URL


arkon:
Update the version in build.gradle too

LeMyst:
> Update the version in build.gradle too

Do I need to update the "extVersionCode" or "libVersion"?

arkon:
Yup, thanks!

PR #4505, opened 2020-10-03
Update issue_closer.yml

sainAk:
Limit the issue_closer workflow to run only when the issue is opened, edited or reopened


arkon:
Thanks!

PR #4571, opened 2020-10-10
LANraragi: Update to API 0.7.2

### LINKED ISSUE #4278: [Bug] LANraragi: Broken
## Device information
- Tachiyomi version: 0.10.4
- Android version: 8.0.0

## Source information
- Name: LANraragi
- Extension version: 1.2.1

## Steps to reproduce
1. Point at server hostname. Give API key.
2. Open source.

### Expected behavior
I see my library.

### Actual behavior
Screenshot_20200831-165244

RePod:
Uses new search, extract, and thumbnail endpoints.
extract endpoint now uses POST instead of GET.
If an API key is provided it is base64 encoded and added to the headers.

Fixes #4278

### DIFF
--- a/src/all/lanraragi/src/eu/kanade/tachiyomi/extension/all/lanraragi/LANraragi.kt
+++ b/src/all/lanraragi/src/eu/kanade/tachiyomi/extension/all/lanraragi/LANraragi.kt
@@ -5,11 +5,13 @@ import android.content.SharedPreferences
 import android.net.Uri
 import android.support.v7.preference.EditTextPreference
 import android.support.v7.preference.PreferenceScreen
+import android.util.Base64
 import com.github.salomonbrys.kotson.fromJson
 import com.google.gson.Gson
 import eu.kanade.tachiyomi.extension.all.lanraragi.model.ArchivePage
 import eu.kanade.tachiyomi.extension.all.lanraragi.model.ArchiveSearchResult
 import eu.kanade.tachiyomi.network.GET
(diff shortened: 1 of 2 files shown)


src/all/lanraragi/src/eu/kanade/tachiyomi/extension/all/lanraragi/LANraragi.kt
```diff
@@ -49,20 +52,23 @@ open class LANraragi : ConfigurableSource, HttpSource() {
...
-        val uri = getApiUriBuilder("/api/extract")
-        uri.appendQueryParameter("id", id)
+        val uri = getApiUriBuilder("/api/archives/$id/extract")
 
         return listOf(
             SChapter.create().apply {
                 val uriBuild = uri.build()
 
-                url = "${uriBuild.encodedPath}?${uriBuild.encodedQuery}"
+                url = "${uriBuild.encodedPath}"
```
arkon, line 61: Can remove the string interpolation here.

RePod: Would `uriBuild.encodedPath.toString()` suffice?

arkon: It should, if it doesn't already give you a string.

arkon:
Remember to bump the version number in the `build.gradle` file as well.

PR #4596, opened 2020-10-12
Go through the sources and change fallback chapter numbers to -1

jobobby04:
Went through the sources and changed the chapter number fallback to -1 where I could find any


src/ru/mintmanga/src/eu/kanade/tachiyomi/extension/ru/mintmanga/Mintmanga.kt
```diff
@@ -187,7 +187,7 @@ class Mintmanga : ParsedHttpSource() {
                 }
             }
             extra.containsMatchIn(chapter.name) -> // Extra chapters doesn't contain chapter number
-                chapter.chapter_number = -2f
+                chapter.chapter_number = -1f
```
arkon, line 190: Probably fine to keep these `-2f` ones. Probably done to help order them, but the app generally just checks if it's `>= 0`.

jobobby04: Done

Issue #4732, opened 2020-10-27
[Bug] <Jinmantiantang>: <New chapters are being cropped and it's unreadable>


XeroxV:
Device information
- Tachiyomi version: SY Stable 1.3.1
- Android version: 8

 Source information
- Name: Jinmantiantang (禁漫天堂）
- Extension version: 1.2.3

 Steps to reproduce
1. Read the new chapters

Expected behavior
The images should load normally

 Actual behavior
All the images are being cropped and it's unreadable

![1](https://user-images.githubusercontent.com/67626218/97345738-652cf080-18c5-11eb-9188-e946ea8323a9.jpg)

CarlosEsco:
#4767

PR #4767, opened 2020-11-02
Solve the problem that the pictures are arranged in reverse order after being segmented

ZingerZY:
Solve the problem that the pictures are arranged in reverse order after being segmented


Riztard:
will Closes #4732

PR #4828, opened 2020-11-08
[CI SKIP] Improve wording, manga details new info

AriaMoradi:
I'm not 100% sure about the caching behavior. please confirm and correct.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -172,9 +172,10 @@ The class which is refrenced and defined by `extClass` in `build.gradle`. This c
 
 a.k.a. the Browse source entry point in the app (invoked by tapping on the source name).
 
-- The app calls `fetchPopularManga` with `page=1`, and it returns a `MangasPage` and will continue to call it for next pages, when the user scrolls the manga list and more results must be fetched (until you pass `MangasPage.hasNextPage` as `false` which marks the end of the found manga list).
-- While passing magnas here you should at least set `url`, `title` and `thumbnail_url`.
-    - If `thumbnail_url` is not set, `fetchMangaDetails` will be **immediately** called.
+- The app calls `fetchPopularManga` which should return a `MangasPage` containing the first batch of found `SManga` entries.
+    - This method supports pagination. When user scrolls the manga list and more results must be fetched, the app calls it again with increasing `page` values(starting with `page=1`). This continues until you pass `MangasPage.hasNextPage` as `false` which marks the end of the found manga list.
+- To show the list properly, the app needs `url`, `title` and `thumbnail_url`. You must set them here. The rest of the fields could be filled later.(refer to Manga Details below)
+    - If `thumbnail_url` is not avialable, you could not set it, in this case `fetchMangaDetails` will be **immediately** called. To fetch the info.
 
 #### Latest Manga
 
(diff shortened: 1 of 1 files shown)


AriaMoradi:
this one is gathering dust! is there any problems with it?

CONTRIBUTING.md
```diff
@@ -192,9 +193,12 @@ a.k.a. the Latest source entry point in the app (invoked by tapping on the "Late
 #### Manga Details
 
 - When user taps on a manga, `fetchMangaDetails` and `fetchChapterList` will be called and the results will be cached.
+    - A `SManga` entry is identified by it's `url`.
```
CarlosEsco, line 196: yes and no.  Its identified by its db id.  to be more precise but I don't know why that needs to be said here as it seems unrelated

CONTRIBUTING.md
```diff
@@ -192,9 +193,12 @@ a.k.a. the Latest source entry point in the app (invoked by tapping on the "Late
...
 - `fetchMangaDetails` is called to update a manga's details from when it was initialized earlier.
     - During a backup, only `url` and `title` are stored. To restore the rest of the manga data, the app calls `fetchMangaDetails`, so all fields should be (re)filled in if possible.
     - `SManga.initialized` tells the app if it should call `fetchMangaDetails`. If you are overriding `fetchMangaDetails`, make sure to pass it as `true`.
+    - If a `SManga` is cached and you update the details in `fetchPopularManga`(or similar methods), details it won't change in the database(and thumbnails won't update). updates to database are only done after `fetchMangaDetails` is called.
```
CarlosEsco, line 200: This is incorrect.  There is no cache.  A manga when clicked into the first time is saved to the db.  Any clicks into that manga are just reading from the db unless a manual refresh is done (or if its a library manga and the library update is ran).  No objects are cached, the only time a cache would play a result is the HTTP response is cached for ~15mins by default so if you manually refreshed a manga multiple times in 15mins  there would be no network call out it for the fetch manga details

CONTRIBUTING.md
```diff
@@ -192,9 +193,12 @@ a.k.a. the Latest source entry point in the app (invoked by tapping on the "Late
...
+    - If a `SManga` is cached `fetchMangaDetails` will be only called when the user does a manual update(Swipe-to-Refresh).
 - `fetchChapterList` is called to display the chapter list.
     - The list should be sorted descending by date/chapter number.
```
CarlosEsco, line 203: It should be sorted by the source order on the website descending

CONTRIBUTING.md
```diff
@@ -172,9 +172,10 @@ The class which is refrenced and defined by `extClass` in `build.gradle`. This c
...
-- While passing magnas here you should at least set `url`, `title` and `thumbnail_url`.
-    - If `thumbnail_url` is not set, `fetchMangaDetails` will be **immediately** called.
+- The app calls `fetchPopularManga` which should return a `MangasPage` containing the first batch of found `SManga` entries.
+    - This method supports pagination. When user scrolls the manga list and more results must be fetched, the app calls it again with increasing `page` values(starting with `page=1`). This continues until you pass `MangasPage.hasNextPage` as `false` which marks the end of the found manga list.
+- To show the list properly, the app needs `url`, `title` and `thumbnail_url`. You must set them here. The rest of the fields could be filled later.(refer to Manga Details below)
+    - If `thumbnail_url` is not avialable, you could not set it, in this case `fetchMangaDetails` will be **immediately** called. To fetch the info.
```
CarlosEsco, line 178: add a caveat that if possible try to set it.  This makes the source super slow as you are making 2N number of additional network requests and not a lot of sources appreciate that.

CarlosEsco (changes requested):

CONTRIBUTING.md
AriaMoradi: this explains when the same manga with the same name can exist in the context of Source

CONTRIBUTING.md
AriaMoradi: ok, removing this line.

CONTRIBUTING.md
AriaMoradi: done

CONTRIBUTING.md
AriaMoradi: done

PR #4886, opened 2020-11-16
Mangapanda: Switch protocol from HTTPS to HTTP

scb261:
Their certificate expired a month ago (28 days ago to be exact) and they still didn't renew it. I guess it's time to switch protocols


CarlosEsco:
Maybe someone should just email them and tell them to fix their cert lol.

Issue #3795, opened 2020-07-15
[Source Request] <copymanga>


maoyi2016:
### Source Website
- Name: copymanga
- Link: [copymanga.com](https://www.copymanga.com)
- Language(s): Chinese

### Other Details
There are mangas that has been deleted in other site.

KiriKira:
Waiting for the new source extension!

arkon:
Already exists

PR #4933, opened 2020-11-23
Add copymanga source

sunbeams001:
#3795


src/zh/copymanga/src/eu/kanade/tachiyomi/extension/zh/copymanga/CopyManga.kt
```diff
@@ -0,0 +1,323 @@
...
+        throw UnsupportedOperationException("This method should not be called!")
+
+    // NOT WORK!!! As for now, copymanga has different logic in polular and search page
+    override fun getFilterList() = FilterList(
+        ThemeFilter(
+            "题材",
+            "theme",
+            arrayOf(
+                Pair("全部", "")
+//                Pair("愛情", "aiqing"),
```
CarlosEsco, line 188: why are these all commented out?

sunbeams001: // NOT WORK!!! As for now, copymanga has different logic in polular and search page
As I commented in the code, for now the framework flow will call "searchManga" progress in polupar page when consider filters, but in this site, it's two different implementation, the popular page by parse html, the search page by api, and only the popular page can filter comics.
I don't know if the site will change the search logic in the future, so I remain these filter items for now.

CarlosEsco: I would remove them then if they are not used at this time.

sunbeams001: OK, got it.

PR #4949, opened 2020-11-25
ComicKiba: fix pageListParse for some manga

AbdullahM0hamed:
Example: https://comickiba.com/manga/the-strongest-cheat-player/


FaTe-code:
Still some manga getting the same error even tho webview works like the "King of Gold".

PR #5023, opened 2020-12-05
Fixed imageloading again

Blatzar:
Reworked image loading functions to simply use the normal urls instead of deobfuscating a string. Now using same system as HentaiFox. Fixes https://github.com/inorichi/tachiyomi-extensions/issues/5022

Should probably be more maintanable now.


kevin01523:
Stil broken tho

CarlosEsco:
Well it changed agsjn and don't comment on closed prs

PR #5041, opened 2020-12-07
copymanga: Add ordering filter

rawfh:


src/zh/copymanga/src/eu/kanade/tachiyomi/extension/zh/copymanga/CopyManga.kt
```diff
@@ -43,7 +43,7 @@ class CopyManga : HttpSource() {
         var requestUrlString: String
 
         val params = filters.map {
-            if (it is ThemeFilter) {
+            if (it is MangaFilter) {
                 it.toUriPart()
             } else ""
         }.filter { it != "" }.joinToString("&")
```
sunbeams001, line 49: Is there any conflicts when involving ordering filter?
I mean, the `htmlUrlString` already has ordering parameter by default

rawfh: Oh, sorry, I didn't notice the default parameter. But it still works, the server will only use the last ordering parameter.
Like this: [https://www.copymanga.com/comics?ordering=-popular&ordering=popular]()

sunbeams001: Got it.

PR #5089, opened 2020-12-11
Add Zahard and remove some dead sources.

Vesnyx:
Zahard appears to have change their main site to this CMS.

Other sources had slight domain changes or completely removed their readers.


Hasan-01:
Add Zahard.

PR #5109, opened 2020-12-12
Readmanhwa.com enhanced search and filters - fix #4972

### LINKED ISSUE #4972: [Bug] readmanhwa.com: enhance search
## Device information
- Tachiyomi version: 0.10.5
- Android version: 10

## Source information
- Name: readmanhwa
- Extension version: 1.2.6

1- there are 54 current tags but the site has 85
2- the site can sort by a field but extension not
3- the site can sort ascending or descending 
4- ~no duration field in the current api~ duration could be added as a filter
5- site can filter by `Artist`, `Author`, `Attributes`?, and `Statuses`
6- per_page could be a larger value than 18...

antonycaporossi:
Closes #4972

### DIFF
--- a/src/en/readmanhwa/build.gradle
+++ b/src/en/readmanhwa/build.gradle
@@ -5,7 +5,7 @@ ext {
     extName = 'ReadManhwa'
     pkgNameSuffix = 'en.readmanhwa'
     extClass = '.ReadManhwa'
-    extVersionCode = 6
+    extVersionCode = 7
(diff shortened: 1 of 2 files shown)


AriaMoradi:
What's changed?

antonycaporossi:
Added more genres
All sort fields now are availables
Same for the ordering (with desc and asc)
Added all "Duration" fields
Doubled per_page value (= less requests)

PR #5112, opened 2020-12-12
[Ru]Mangalib. Fix images with double slash

Pavkazzz:
Fix default server

Fix images not loading. Cause server require double slash after hostname


Pavkazzz:
mangalib.me goes crasy and shake cloudflare settings of image cdn

PR #5135, opened 2020-12-15
New Source: Keenspot

### LINKED ISSUE #457: [English] TwoKinds
http://twokinds.keenspot.com/archive/ select `a[href^="/comic"]` for chapter links
http://twokinds.keenspot.com/comic/1/ select `img[alt="Comic Page"]` for image
If no one picks it up, I'll probably get to finishing it sometime.

AriaMoradi:
closes #457

This is a comic hosting website and each comic has their own structure, and each requires a different source to be made.
Thus, this is a multi-source extension but only TwoKinds is implemented for now.


AriaMoradi:
The icons @as280093 made work for towkinds but this should have icons for keenspot

src/en/keenspot/src/eu/kanade/tachiyomi/extension/en/keenspot/KeenspotFactory.kt
```diff
@@ -0,0 +1,12 @@
+package eu.kanade.tachiyomi.extension.en.keenspot
+
+import eu.kanade.tachiyomi.source.Source
+import eu.kanade.tachiyomi.source.SourceFactory
+
+/**
+ *  @author Aria Moradi <aria.moradi007@gmail.com>
+ */
+
```
CarlosEsco, line 9: remove author section

src/en/keenspot/src/eu/kanade/tachiyomi/extension/en/keenspot/TwoKinds.kt
```diff
@@ -0,0 +1,161 @@
...
+import eu.kanade.tachiyomi.source.online.HttpSource
+import eu.kanade.tachiyomi.util.asJsoup
+import okhttp3.Request
+import okhttp3.Response
+import rx.Observable
+
+/**
+ *  @author Aria Moradi <aria.moradi007@gmail.com>
+ */
+
```
CarlosEsco, line 19: same as above

CarlosEsco:
@arkon  thoughts?  I feel like this shouldn't be a multisource extension if you have to create a source and override the httpsource anyways.  The whole point of multisource is they are using the same code, but this just seems that they are keenspot but have no relation since the comic structure is different per a comic

AriaMoradi:
> The whole point of multisource is they are using the same code, but this just seems that they are keenspot but have no relation since the comic structure is different per a comic

The only reason I did this is to make finding each source easier and future proofing if we add more keenspot comics.

src/en/keenspot/src/eu/kanade/tachiyomi/extension/en/keenspot/KeenspotFactory.kt
AriaMoradi: May I ask why?

AbdullahM0hamed:
> make  finding each source easier
Maybe I've understood you incorrectly, but this would actually make finding each source harder as we only have a search for extensions and not individual sources.

AriaMoradi:
> > make  finding each source easier
> 
> Maybe I've understood you incorrectly, but this would actually make finding each source harder as we only have a search for extensions and not individual sources.

I assume when someone wants to read TwoKind doesn't search for TwoKind, they search for keenspot. tell me if I'm wrong.

arkon:
I do prefer if extension modules are actually related by their code, but I'm aware that TwoKind and Keenspot do have a strong coupling in their website's relationships.

I'm fine with how it is for now, we can split it if it causes actually maintenance issues.

src/en/keenspot/src/eu/kanade/tachiyomi/extension/en/keenspot/KeenspotFactory.kt
arkon: We just generally don't do this since there's a lot of contributors, and it gets confusing if others start modifying it too.

If you really feel strongly about "owning" this extension and being the sole person responsible for it, I'd be open to you keeping it.

src/en/keenspot/src/eu/kanade/tachiyomi/extension/en/keenspot/TwoKinds.kt
```diff
@@ -0,0 +1,161 @@
...
+        val lastPage = document.select(".navprev").first().attr("href").split("/")[2].toInt() + 1
+
+        val chapters = mutableListOf<SChapter>()
+
+        if (manga.url == "1") {
+            for (i in 1..lastPage) {
+                chapters.add(
+                    SChapter.create().apply() {
+//                        url = baseUrl + "/comic/${i}/"
```
arkon, line 104: Remove the commented out lines

src/en/keenspot/src/eu/kanade/tachiyomi/extension/en/keenspot/KeenspotFactory.kt
AriaMoradi: I just like to keep track of which extensions I originally made...

src/en/keenspot/src/eu/kanade/tachiyomi/extension/en/keenspot/TwoKinds.kt
AriaMoradi: done

src/en/keenspot/src/eu/kanade/tachiyomi/extension/en/keenspot/KeenspotFactory.kt
CarlosEsco: write it in a spreadsheet or use https://github.com/tachiyomiorg/tachiyomi-extensions/commits?author=AriaMoradi.   I don't feel anyone should be personally "owning" a file in an opensource project.

AriaMoradi: Alright

PR #5136, opened 2020-12-15
Timelessleaf add manga description

Riztard:


AriaMoradi:
You should test any new feature for this source on every single entry they have...

Riztard:
i did

PR #5144, opened 2020-12-15
Update MangaLink.kt

baytout:


arkon:
Please update the `extVersionCode` value in the extensions' `build.gradle` files as well.

PR #4288, opened 2020-09-02
Genkan - convert to themesources implementation

SnakeDoc83:
Introduces `themesources` library and replaces the multi-source Genkan extension with individual extensions.

See https://github.com/inorichi/tachiyomi-extensions/issues/4287

Issue #4287, opened 2020-09-02
[Meta] An end to multi-source extensions


SnakeDoc83:
Extensions like Madara that support multiple websites have had their time but that time is coming to an end. Going forward, extensions that contain multiple sources will be split in to single-source extensions that inherit shared code from a library. For end users this means that if they want to stay up to date they'll eventually need to uninstall their multi-source extension and install individual extensions for their sources as replacements. On the upside, eventually there won't be a need to search for extensions in multiple places or update every 3-4 days because someone found a new Madara site.

For developers: don't PR any new multi-source factory extensions.  Multi-lang factory extensions (one website with multiple languages) are fine and updates to _current_ multi-source extensions are also fine.  The general process for creating these new extensions is a bit more complicated than making a factory extension.  I've commented some of my code but as I convert more extensions I'm sure I'll have to make changes to the process; so for now it'd be pointless to go in to great detail about it.  The general concept is that there's a library (`themesources`) that will contain all the common code for what used to be multi-source factory extensions; individual extensions are generated from an executable script that creates boilerplate code and default icons for new sources which can then be modified.

Extension icons: As I convert extensions the new single-source extensions will continue to use the icons from their parent multi-source extension. It's possible to give each of them their own icons but I won't be spending time on that.  If you want to submit icons keep in mind they should have rounded corners.

arkon:
See #5672

PR #5154, opened 2020-12-16
lib-themesources, split Genkan into single-source extensions

AriaMoradi:
This is my attempt at continuing what @SnakeDoc83 did in [his library branch](https://github.com/SnakeDoc83/tachiyomi-extensions/tree/library) and relates to #4287 

related to tachiyomiorg/tachiyomi#3338
closes #4288
can close tachiyomiorg/tachiyomi#3872


AriaMoradi:
Here are few problems and ideas with this solution:
1- each source must use the exact same code as the general implementation, This current implementation does not support source overrides, they must be added after sources are generated.
2- since sources are generated they should be gitignored and generated at build time.

AriaMoradi:
can be merged now.

AriaMoradi:
One idea that I explored(and works) is putting generated extensions in a directory other than `src` like `srcgen`.
the only thing that needs to be changed for it to work might be github actions.

It might make things cleaner considering how big madara(and other multisource extensions) are.
opinions?

arkon:
Yes, having a separate `generated-src` folder or something would be preferable, then we can gitignore it as well.

I'll review this at some point after that's changed so the PR doesn't include the generated contents' diffs too.

AriaMoradi:
> Yes, having a separate generated-src folder or something would be preferable, then we can gitignore it as well.

gitignoring it is not a good idea, if we want to generate on every build two things need to be done, 1- somekind of tool to get you autogen 2- a framework to handle overrides 3- somewhere to put icons

also a big bad point even after doing it is that it makes the library heavy and centralized for no good reason.
If we need to change `themesource`'s API it's going to need more changes(sometimes less)

AriaMoradi:
overrides done.

but for automatic builds I couldn't figure it out completely:

3 steps are needed:(I got stuck on step 3)
1- build the library: `./gradlew :lib-themesources:generateDebugSources :lib-themesources:generateDebugAndroidTestSources :lib-themesources:createMockableJar :lib-themesources:compileDebugAndroidTestSources :lib-themesources:compileDebugUnitTestSources :lib-themesources:compileDebugSources`
2- collect all Generator files: is not too hard
3- run Generator files: to do this 3 things are needed: 1- jre path 2- java classpath 3- Generator className

Also my research suggests that it should be possible to do with gradle, but I couldn't fine any resources describing a working solution.

AriaMoradi:
Can I get a github project/milestone for spiting multi-source extensions?

arkon:
> Can I get a github project/milestone for spiting multi-source extensions?

I think you should be able to have your own projects in your fork.

AriaMoradi:
> 2- collect all Generator files: is not too hard

`GeneratorMain` does this. Only running this file remains.

AriaMoradi:
ready for review

AriaMoradi:
so far trying to re-create the class path I've come with this gradle task:
```gradle
tasks.register("findjars") {
    doLast {
        "$androidSDK/platforms/android-29/android.jar"
        "$androidSDK/platforms/android-29/data/res"
        "$projectRoot/lib/themesources/build/intermediates/javac/debug/classes"
        "$projectRoot/lib/themesources/build/intermediates/compile_r_class_jar/debug/R.jar"
        "$projectRoot/lib/thzemesources/build/tmp/kotlin-classes/debug"
        "$projectRoot/lib/themesources/build/generated/res/resValues/debug"

        var classPath = ""

        configurations.debugCompileOnly.asFileTree.forEach { classPath = "$classPath$it:" }
```

to complete this I need to 1- find the path for android sdk(or where android platform we are building for is) 2- get the project root path(or this module path)

AriaMoradi:
I tried the installed jre(openjdk 8) on my system and it worked!

so this is the new gradle task:

```gradle
tasks.register("classpath") {
    doLast {
        // android sdk dir is not documented but this hidden api gets it
        // ref: https://stackoverflow.com/questions/20203787/access-sdk-dir-value-in-build-gradle-after-project-evaluation
        val androidSDK = "${android.sdkDirectory.absolutePath}"

        val projectRoot = rootProject.projectDir

        var classPath = ""
        classPath += "$androidSDK/platforms/android-29/android.jar:"
        classPath += "$androidSDK/platforms/android-29/data/res:"
        classPath += "$projectRoot/lib/themesources/build/intermediates/javac/debug/classes:"
        classPath += "$projectRoot/lib/themesources/build/intermediates/compile_r_class_jar/debug/R.jar:"
        classPath += "$projectRoot/lib/themesources/build/tmp/kotlin-classes/debug:"
        classPath += "$projectRoot/lib/themesources/build/generated/res/resValues/debug"

        configurations.debugCompileOnly.asFileTree.forEach { classPath = "$classPath:$it" }

        val javaPath = System.getProperty("java.home") + "/bin/java"

        val mainClass = "eu.kanade.tachiyomi.lib.themesources.GeneratorMain"

        println("$javaPath -classpath $classPath $mainClass")
}
```

AriaMoradi:
> now just running
> 
> ```shell
> $ $(./gradlew -q :lib-themesources:classpath)
> ```
> 
> works fine.

this is the final form

```gradle
tasks.register("runAllGenerators") {
    doLast {
        // android sdk dir is not documented but this hidden api gets it
        // ref: https://stackoverflow.com/questions/20203787/access-sdk-dir-value-in-build-gradle-after-project-evaluation
        val androidSDK = "${android.sdkDirectory.absolutePath}"

        val projectRoot = rootProject.projectDir

        var classPath = ""
        classPath += "$androidSDK/platforms/android-29/android.jar:"
        classPath += "$androidSDK/platforms/android-29/data/res:"
        classPath += "$projectRoot/lib/themesources/build/intermediates/javac/debug/classes:"
        classPath += "$projectRoot/lib/themesources/build/intermediates/compile_r_class_jar/debug/R.jar:"
        classPath += "$projectRoot/lib/themesources/build/tmp/kotlin-classes/debug:"
        classPath += "$projectRoot/lib/themesources/build/generated/res/resValues/debug"

        configurations.debugCompileOnly.asFileTree.forEach { classPath = "$classPath:$it" }

        val javaPath = System.getProperty("java.home") + "/bin/java"

        val mainClass = "eu.kanade.tachiyomi.lib.themesources.GeneratorMain"

        Runtime.getRuntime().exec("$javaPath -classpath $classPath $mainClass")
    }
}
```

AriaMoradi:
Now changing githib workflows to run `:lib-themesources:runAllGenerators`
@arkon can you do this part please?

AbdullahM0hamed:
> Now changing github workflows to run :lib-themesources:runAllGenerators the only thing left to be done.
arkon can you do this part please?

Why not just change the work flow as part of your PR?

AriaMoradi:
> Why not just change the work flow as part of your PR?

I don't have access to github secrets but I'll try

AbdullahM0hamed:
> I don't have access to github secrets and work flows are not transferable.

You can create secrets for your fork though, right? I think you should be able to do it that way

AriaMoradi:
> You can create secrets for your fork though, right? I think you should be able to do it that way

and I dont have access to travis.

AbdullahM0hamed:
> I don't have access to travis

This shouldn't be a problem either, since we migrated from building via Travis to building via github actions (should be automatic, your fork is doing some right now: https://github.com/AriaMoradi/tachiyomi-extensions/actions)

AriaMoradi:
> > I don't have access to travis
> 
> This shouldn't be a problem either, since we migrated from building via Travis to building via github actions (should be automatic, your fork is doing some right now: https://github.com/AriaMoradi/tachiyomi-extensions/actions)

Yes. My fork was old. I try to do it myself.

AriaMoradi:
github workflows done too.

AriaMoradi:
this is ready for merge

.github/workflows/build.yml
```diff
@@ -47,6 +47,26 @@ jobs:
...
+      - name: Run theme sources generator
+        uses: eskatos/gradle-command-action@v1
+        with:
+          build-root-directory: master
+          wrapper-directory: master
+          arguments: :lib-themesources:runAllGenerators
+          wrapper-cache-enabled: true
+          dependencies-cache-enabled: true
+          configuration-cache-enabled: true
+
```
arkon, line 69: The default assemble tasks should invoke this automatically, i.e. we should be able to just run the existing `assembleDebug` or `assembleRelease` tasks like before and have everything be built.

lib/themesources/build.gradle.kts
```diff
@@ -0,0 +1,63 @@
...
+    compileOnly("org.jetbrains.kotlin:kotlin-stdlib-jdk8:${Deps.kotlin.version}")
+
+    compileOnly("com.github.inorichi.injekt:injekt-core:65b0440")
+    compileOnly("com.squareup.okhttp3:okhttp:3.10.0")
+    compileOnly("io.reactivex:rxjava:1.3.6")
+    compileOnly("org.jsoup:jsoup:1.10.2")
+    compileOnly("com.google.code.gson:gson:2.8.2")
+    compileOnly("com.github.salomonbrys.kotson:kotson:2.5.0")
+    compileOnly(project(":duktape-stub"))
+}
```
arkon, line 36: We shouldn't be duplicating all of these configs across "normal" and the theme source module.

Maybe it can be moved into the core module from #5214.

lib/themesources/build.gradle.kts
```diff
@@ -0,0 +1,63 @@
...
+
+
+tasks.register("runAllGenerators") {
```
arkon, line 39: Is this complicated because of the need for the Android SDK? I would've expected some normal `JavaExec` task or something, with maybe some classpath config.

lib/themesources/src/main/java/eu/kanade/tachiyomi/lib/themesources/genkan/GenkanGenerator.kt
```diff
@@ -0,0 +1,32 @@
...
+        SingleLangThemeSourceData("Method Scans", "https://methodscans.com", "en"),
+        SingleLangThemeSourceData("Sleeping Knight Scans", "https://skscans.com", "en")
+    )
+
+    companion object {
+        @JvmStatic
+        fun main(args: Array<String>) {
+            GenkanGenerator().createAll()
+        }
+    }
```
arkon, line 31: Rather than having this for every generator, maybe you can just make a base Launcher class that accept the generator class as an argument?

e.g. https://stackoverflow.com/questions/2988334/inheriting-the-main-method

settings.gradle.kts
```diff
@@ -10,6 +10,9 @@ project(":duktape-stub").projectDir = File("lib/duktape-stub")
 include(":lib-dataimage")
 project(":lib-dataimage").projectDir = File("lib/dataimage")
 
+include(":lib-themesources")
+project(":lib-themesources").projectDir = File("lib/themesources")
```
arkon, line 14: I'd put this in a separate module since it isn't really a library used by the extensions, per se.

Something like `/compiler/themesources`?

arkon (changes requested):
This is my first pass at comments. I'll need to look at the actual generator part later when I have more time.

lib/themesources/build.gradle.kts
AriaMoradi: everything is compile only mind you.

.github/workflows/build.yml
AriaMoradi: Well I copied this from android studio logs, but yes `assembleDebug` worked. guess simpler config is better.

lib/themesources/build.gradle.kts
AriaMoradi: Nope it wont work. I tried `JavaExec` and it can't see the libraries(classpath)
If you can make a equivalent working  `JavaExec` I'll change it

lib/themesources/src/main/java/eu/kanade/tachiyomi/lib/themesources/genkan/GenkanGenerator.kt
AriaMoradi: Doing it this way makes debugging and focusing on one Extension much much eaiser.

AriaMoradi: It acts as a scaffolding and building tool at the same time

settings.gradle.kts
AriaMoradi: You mean move the module directory?

lib/themesources/build.gradle.kts
AriaMoradi: 9 common dependencies. I'll change it after #5214

lib/themesources/src/main/java/eu/kanade/tachiyomi/lib/themesources/genkan/GenkanGenerator.kt
AriaMoradi: sometimes I run something like this do bypass having to sync every time:
`/opt/android-studio/jre/bin/java -javaagent:/opt/android-studio/lib/idea_rt.jar=45393:/opt/android-studio/bin -Dfile.encoding=UTF-8 -classpath /opt/Android-SDK/platforms/android-29/android.jar:/opt/Android-SDK/platforms/android-29/data/res:/home/armor/programming/android/tachiyomi-extensions/src/all/boommanga/build/intermediates/javac/debug/classes:/home/armor/programming/android/tachiyomi-extensions/src/all/boommanga//build/tmp/kotlin-classes/debug:/home/armor/programming/android/tachiyomi-extensions/src/all/boommanga/build/intermediates/compile_and_runtime_not_namespaced_r_class_jar/debug/R.jar:/home/armor/programming/android/tachiyomi-extensions/src/all/boommanga/build/generated/res/resValues/debug:/home/armor/.gradle/caches/modules-2/files-2.1/com.github.tachiyomiorg/extensions-lib/a596412/eaa76a7d8891fc26a3839d591d067e9d71a4aa2/extensions-lib-a596412.jar:/home/armor/.gradle/caches/modules-2/files-2.1/org.jetbrains.kotlin/kotlin-stdlib-jdk8/1.4.10/998caa30623f73223194a8b657abd2baec4880ea/kotlin-stdlib-jdk8-1.4.10.jar:/home/armor/.gradle/caches/modules-2/files-2.1/com.github.inorichi.injekt/injekt-core/65b0440/63b7078668b1d86c7675e65082c9f56c6f5edfa/injekt-core-65b0440.jar:/home/armor/.gradle/caches/modules-2/files-2.1/com.squareup.okhttp3/okhttp/3.10.0/7ef0f1d95bf4c0b3ba30bbae25e0e562b05cf75e/okhttp-3.10.0.jar:/home/armor/.gradle/caches/modules-2/files-2.1/io.reactivex/rxjava/1.3.6/f8596dcb885d65c86f96cf96fe12772c56d51d1d/rxjava-1.3.6.jar:/home/armor/.gradle/caches/modules-2/files-2.1/org.jsoup/jsoup/1.10.2/33ee82e324f4b1e40167f3dc5e01234a1c5cab61/jsoup-1.10.2.jar:/home/armor/.gradle/caches/modules-2/files-2.1/com.github.salomonbrys.kotson/kotson/2.5.0/40daf22108aa4436d46257e14cc48447cca542c3/kotson-2.5.0.jar:/home/armor/.gradle/caches/modules-2/files-2.1/com.google.code.gson/gson/2.8.2/3edcfe49d2c6053a70a2a47e4e1c2f94998a49cf/gson-2.8.2.jar:/home/armor/.gradle/caches/modules-2/files-2.1/org.jetbrains.kotlin/kotlin-stdlib-jdk7/1.4.10/30e46450b0bb3dbf43898d2f461be4a942784780/kotlin-stdlib-jdk7-1.4.10.jar:/home/armor/.gradle/caches/modules-2/files-2.1/org.jetbrains.kotlin/kotlin-stdlib/1.4.10/ea29e063d2bbe695be13e9d044dcfb0c7add398e/kotlin-stdlib-1.4.10.jar:/home/armor/.gradle/caches/modules-2/files-2.1/com.github.inorichi.injekt/injekt-api/65b0440/ff489b6dd4c80b8ace3c703f1d443aea08749f23/injekt-api-65b0440.jar:/home/armor/.gradle/caches/modules-2/files-2.1/com.squareup.okio/okio/1.14.0/102d7be47241d781ef95f1581d414b0943053130/okio-1.14.0.jar:/home/armor/.gradle/caches/modules-2/files-2.1/org.jetbrains.kotlin/kotlin-stdlib-common/1.4.10/6229be3465805c99db1142ad75e6c6ddeac0b04c/kotlin-stdlib-common-1.4.10.jar:/home/armor/.gradle/caches/modules-2/files-2.1/org.jetbrains/annotations/13.0/919f0dfe192fb4e063e7dacadee7f8bb9a2672a9/annotations-13.0.jar:/home/armor/programming/android/tachiyomi-extensions/defaultRes/build/intermediates/compile_r_class_jar/debug/R.jar:/home/armor/programming/android/tachiyomi-extensions/defaultRes/build/generated/res/resValues/debug:/home/armor/programming/android/tachiyomi-extensions/lib/themesources/build/intermediates/javac/debug/classes:/home/armor/programming/android/tachiyomi-extensions/lib/themesources/build/intermediates/compile_r_class_jar/debug/R.jar:/home/armor/programming/android/tachiyomi-extensions/lib/themesources/build/tmp/kotlin-classes/debug:/home/armor/programming/android/tachiyomi-extensions/lib/themesources/build/generated/res/resValues/debug:/home/armor/programming/android/tachiyomi-extensions/annotations/build/classes/kotlin/main:/home/armor/programming/android/tachiyomi-extensions/annotations/build/libs/annotations.jar:/home/armor/programming/android/tachiyomi-extensions/lib/duktape-stub/build/classes/java/main:/home/armor/programming/android/tachiyomi-extensions/lib/duktape-stub/build/libs/duktape-stub.jar:/home/armor/.gradle/caches/modules-2/files-2.1/org.jetbrains.kotlin/kotlin-reflect/1.4.21/748f681f4e3edbe9285ff46710c79049c70f4dfa/kotlin-reflect-1.4.21.jar eu.kanade.tachiyomi.extension.all.boommanga.Inspector`

settings.gradle.kts
arkon: > You mean move the module directory?

Only for this module, since it's not a library (all of the existing ones are dependencies/functionality that extensions actually use, rather than a build tool).

lib/themesources/src/main/java/eu/kanade/tachiyomi/lib/themesources/genkan/GenkanGenerator.kt
arkon: Assuming each generator module has a respective build task, can't you just invoke that from the list of tasks in the Gradle pane?

I'm fine with keeping these if there's no clean way though, the run icon argument is fair.

AriaMoradi: Also when the library is frozen in the future and a extension needs something special, i.e. gradle extra dependencies, they can customize the behavior completely this way.

settings.gradle.kts
AriaMoradi: How about moving it to project root?

and rename it to to `multisource` or `multisrc` than `themesource`?
`multisrc` looks the best.

arkon: `multisrc` sounds fine.

AriaMoradi: everything moved to `multisrc`

lib/themesources/build.gradle.kts
AriaMoradi: the problem is finding out if there's a way to calculate the classpath automatically
I found this https://stackoverflow.com/questions/51810254/execute-javaexec-task-using-gradle-kotlin-dsl

but sourceAets are moved to `project.android.sourceSets` and no longer include the classpath

AriaMoradi: also this, but again doesn't work: https://stackoverflow.com/questions/54082700/run-gradle-task-javaexec-in-android-gradle-plagin

AriaMoradi: I cleaned it up as much as possible

.github/workflows/build.yml
```diff
@@ -47,6 +47,26 @@ jobs:
...
+      - name: Run multi-source generators
+        uses: eskatos/gradle-command-action@v1
+        with:
+          build-root-directory: master
+          wrapper-directory: master
+          arguments: :multisrc:runAllGenerators
+          wrapper-cache-enabled: true
+          dependencies-cache-enabled: true
+          configuration-cache-enabled: true
```
arkon, line 68: I commented on this last time too, but I mentioned that all of the new tasks should be run as part of the main `assemble*` (e.g. `assembleRelease`) tasks. The end result is that this workflow shouldn't change.

multisrc/build.gradle.kts
```diff
@@ -0,0 +1,44 @@
...
+
+repositories {
+    mavenCentral()
+}
+
+dependencies {
+    compileOnly(project(":annotations"))
+
+    compileOnly("com.github.salomonbrys.kotson:kotson:2.5.0")
+    compileOnly(project(":duktape-stub"))
```
arkon, line 24: Why aren't these just in `common-dependencies.gradle`?

multisrc/build.gradle.kts
```diff
@@ -0,0 +1,44 @@
...
+
+
+
+tasks.register("runAllGenerators") {
+    doLast {
+        var classPath = ""
+        classPath += configurations.androidApis.get().asFileTree.first().absolutePath + ":" // android.jar path
+        classPath += "$projectDir/build/intermediates/aar_main_jar/debug/classes.jar:" // jar made from this module
+
+        configurations.debugCompileOnly.get().asFileTree.forEach { classPath = "$classPath:$it" } // every dependency we have
```
arkon, line 36: You can make a list of the various things and join them with `:` rather than this string concatenation.

multisrc/src/main/java/eu/kanade/tachiyomi/multisrc/GeneratorMain.kt
```diff
@@ -0,0 +1,37 @@
+package eu.kanade.tachiyomi.multisrc
+
+import java.io.File
+
+/**
+ * Finds and calls all `ThemeSourceGenerator`s
+ */
+class GeneratorMain {
```
arkon, line 8: Does this work if the file just has the main function? I vaguely recall not needing the class or object at all in Kotlin, although I'm not sure if that's just because it "just works" in IntelliJ, or if invoking it via `java` actually works normally too.

multisrc/src/main/java/eu/kanade/tachiyomi/multisrc/GeneratorMain.kt
```diff
@@ -0,0 +1,37 @@
...
+
+                // invoke main methods
+                generatorClasses.forEach {
+                    val generatorClassPath = "eu/kanade/tachiyomi/multisrc/$themeSource/$it".replace("/", ".").substringBefore(".kt")
+                    Class.forName(generatorClassPath).methods.forEach {
+                        if (it.name == "main")
+                            it.invoke(null, emptyArray<String>())
+                    }
+                }
+            }
```
arkon, line 34: You could probably clean this up a lot, e.g. something like

```kotlin
File(sourcesDirPath).list()!!
  .filter { File(sourcesDir, it).isDirectory }
  .flatMap { 
    File("$sourcesDirPath/$themeSource").list()!!.filter {
      it.endsWith("Generator.kt")   
    }
  }
  .map { 
    val generatorClassPath =  "eu/kanade/tachiyomi/multisrc/$themeSource/$it".replace("/", ".").substringBefore(".kt") 
    Class.forName(generatorClassPath).methods
  }
  .filter { it.name == "main" }
  .forEach { it.invoke(null, emptyArray<String>()) }
```

multisrc/src/main/java/eu/kanade/tachiyomi/multisrc/ThemeSourceGenerator.kt
```diff
@@ -0,0 +1,202 @@
...
+    companion object {
+        private fun pkgNameSuffix(source: ThemeSourceData, separator: String): String {
+            return if (source is SingleLangThemeSourceData)
+                listOf(source.lang.substringBefore("-"), source.pkgName).joinToString(separator)
+            else
+                listOf("all", source.pkgName).joinToString(separator)
+        }
+
+        private fun writeGradle(gradle: File, source: ThemeSourceData, baseVersionCode: Int) {
+            var text = "apply plugin: 'com.android.application'\n" +
```
arkon, line 51: You can do multiline strings in Kotlin. Same comment goes for other parts of this generator too.

common.gradle
```diff
@@ -60,20 +60,11 @@ dependencies {
...
-
-    compileOnly 'com.github.inorichi.injekt:injekt-core:65b0440'
-    compileOnly 'com.squareup.okhttp3:okhttp:3.10.0'
-    compileOnly 'io.reactivex:rxjava:1.3.6'
-    compileOnly 'org.jsoup:jsoup:1.10.2'
-    compileOnly 'com.google.code.gson:gson:2.8.2'
-    compileOnly 'com.github.salomonbrys.kotson:kotson:2.5.0'
-    compileOnly project(':duktape-stub')
+    // multi-source library
+    implementation project(':multisrc')
```
arkon, line 64: What's the impact of this on the APK outputs?

multisrc/src/main/java/eu/kanade/tachiyomi/multisrc/ThemeSourceGenerator.kt
```diff
@@ -0,0 +1,202 @@
...
+        abstract class ThemeSourceData {
+            abstract val name: String
+            abstract val baseUrl: String
+            abstract val isNsfw: Boolean
+            abstract val className: String
+            abstract val pkgName: String
+
+            /**
+             * overrideVersionCode defaults to 0, if a source changes their override code or
+             * a previous existing source suddenly needs overrides, overrideVersionCode should be increased.
```
arkon, line 171: Overrides for what? This JavaDoc isn't clear.

multisrc/src/main/java/eu/kanade/tachiyomi/multisrc/ThemeSourceGenerator.kt
```diff
@@ -0,0 +1,202 @@
...
+                    classText += "        $themeClass(\"${source.name}\", \"${source.baseUrl}\", \"$lang\"),\n"
+                classText +=
+                    "    )\n" +
+                        "}"
+            }
+
+            classFile.writeText(classText)
+        }
+
+        abstract class ThemeSourceData {
```
arkon, line 162: You should use a sealed class for these.

settings.gradle.kts
```diff
@@ -18,6 +21,14 @@ File(rootDir, "src").eachDir { dir ->
         project(name).projectDir = File("src/${dir.name}/${subdir.name}")
     }
 }
+    // Loads generated extensions from multisrc
```
arkon, line 24: Indentation is off.

arkon (changes requested):

.github/workflows/build.yml
AriaMoradi: what about `runAllGenerators`?

arkon: The name doesn't matter too much, but the original assemble tasks should depend on your new ones.

The idea is that the same old "simple" build will still build everything in an idempotent fashion.

multisrc/build.gradle.kts
AriaMoradi: `implementation` vs `compileOnly`?

multisrc/build.gradle.kts
AriaMoradi: it's a special kind of collection, I'm not sure how to convert it to a list?

multisrc/build.gradle.kts
arkon: 2 of them _are_ in the other file though. The annotations bytecode is required for them to work too.

common.gradle
AriaMoradi: All multitheme libraries will be included as they are in the package, i suggested running proguard  to fix this with tree shaking .

it's probably possible to split them inside different gradle modules and include the proper theme class in each respective extension; though i suggest doing this later. it's not big enough yet to account to much.

multisrc/src/main/java/eu/kanade/tachiyomi/multisrc/ThemeSourceGenerator.kt
AriaMoradi: each extension either uses the default generated code or a code override, look at `multisrc/overrides/src`

multisrc/src/main/java/eu/kanade/tachiyomi/multisrc/GeneratorMain.kt
AriaMoradi: it worked.

multisrc/build.gradle.kts
AriaMoradi: sorry, i missed them.

AriaMoradi: fixed

.github/workflows/build.yml
AriaMoradi: this probably can be fixed by a `dependsOn`, but I'm not sure where and on what it should be applied

multisrc/src/main/java/eu/kanade/tachiyomi/multisrc/ThemeSourceGenerator.kt
AriaMoradi: done.

multisrc/src/main/java/eu/kanade/tachiyomi/multisrc/GeneratorMain.kt
AriaMoradi: done

multisrc/src/main/java/eu/kanade/tachiyomi/multisrc/ThemeSourceGenerator.kt
AriaMoradi: done

multisrc/src/main/java/eu/kanade/tachiyomi/multisrc/ThemeSourceGenerator.kt
AriaMoradi: I made it clearer?

common.gradle
AriaMoradi: It's possible to copy the theme class (i.e. `Genkan` and `GenkanOriginal`) to the generated module and bundle it this way.

what do you think?

AriaMoradi: fixed in 0a11e72

multisrc/build.gradle.kts
```diff
@@ -0,0 +1,39 @@
...
+    id("com.android.library")
+    kotlin("android")
+}
+
+android {
+    compileSdkVersion(Config.compileSdk)
+    buildToolsVersion(Config.buildTools)
+
+    defaultConfig {
+        minSdkVersion(29)
```
arkon, line 11: Why 29?

.github/workflows/build.yml
arkon: I think having this in the root `build.gradle` file works:

```
task assemble {
    dependsOn(":multisrc:assemble")
}
```

multisrc/build.gradle.kts
arkon: Something like this should work:

```
val classPath =
    // every dependency we have 
    configurations.debugCompileOnly.get().asFileTree + 
    listOf(
        // android.jar path
        configurations.androidApis.get().asFileTree.first().absolutePath,
        // jar made from this module
        "$projectDir/build/intermediates/aar_main_jar/debug/classes.jar"
    )
    .joinToString(":")
```

multisrc/build.gradle.kts
AriaMoradi: because `java.nio.file.Files.copy` and this is essentially only a command line app only, so It's not important which api level it uses.

.github/workflows/build.yml
AriaMoradi: This kind of solution won't work because even if somehow we call `:multisrc:assemble` then `:multisrc:runAllGenerators`, we must restart gradle because it has to read `settings.gradle.kts` again and find the generated modules, looks like there is no way to include projects in the middle of tasks running.

multisrc/build.gradle.kts
AriaMoradi: done

arkon (changes requested):
You'll need to fix the conflicts before I can merge this.

AriaMoradi:
It should be fine now.

Issue #5290, opened 2021-01-01
[Source Request] <LoveHug>


lasthm:
## Source Website
- Name: LoveHug
- Link: [https://lovehug.net/](https://lovehug.net/)
- Language(s): #JP

## Other Details
LoveHeaven(RawLH [https://loveheaven.net](https://loveheaven.net)) is closed.They transfers to LoveHug.
```
Transfer Website to LoveHug.

LoveHeaven is closed. We transfer to a new domain.

You can access the website "LoveHug" to continue reading the new manga.

We sincerely apologize for the trouble.

This is new theme of the website.
```

PR #5292, opened 2021-01-01
baseUrl: loveheaven.net -> lovehug.net (#5290)

OncePunchedMan:
changed baseUrl for RawLH 
issue #5290


Riztard:
they use slightly different theme :/

PR #5327, opened 2021-01-06
fix scantrad

### LINKED ISSUE #5322: [Bug] <Scantrad>: <isn't working.. again lol>
## Device information
- Tachiyomi version: r2959
- Android version: 10

## Source information
- Name: Scantrad [FR]
- Extension version: 1.2.8

### Actual behavior

Extension isn't working (again), it is the same problem as #5243 

## Other details

Scantrad team thinks you are a manga stealing app
(yup they're smooth brains)

OncePunchedMan:
fixed scantrad blocking tachis user agent by randomizing it for now. guess they'll keep blocking so maybe we need to remove them

Closes #5322

### DIFF
--- a/src/fr/scantrad/build.gradle
+++ b/src/fr/scantrad/build.gradle
@@ -5,7 +5,7 @@ ext {
     extName = 'Scantrad'
     pkgNameSuffix = 'fr.scantrad'
     extClass = '.Scantrad'
-    extVersionCode = 8
+    extVersionCode = 9
     libVersion = '1.2'
 }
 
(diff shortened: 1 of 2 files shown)


Zealest:
thanks for the fix :)) however it looks like the extension is blocked now, and i guess it will work after some times (i deduce it is due to the randomizing?)

OncePunchedMan:
@Zealest it's best to delete cache and cookies in settings -> advanced. it should work after that

Zealest:
it's weird because the "home screen" of the ext. still doesn't work (atleast for me)
Screenshot_20210106-030639
but chapters are loading correctly (even though it's a bit longer than usual)
Screenshot_20210106-030654

Zealest:
**it's weird because the "home screen" isn't working (atleast for me)**


Screenshot_20210106-030621



**but chapters are loading correctly (even though it is a bit longer than usual)**


Screenshot_20210106-030639

Zealest:
**or for example like rn**

Screenshot_20210106-031618

i can't load any chapters, like i said earlier it's weird because the extension is working, then it isn't, then it is bla bla bla

OncePunchedMan:
Can't seem to figure out how to solve this tbh. Maybe someone elseknows how, but i guess we will put it to rest since they'll keep blocking tachi over and over again

Zealest:
yup it isnt working anymore for me :(

PR #5430, opened 2021-01-15
New extension for IMHentai.com

vaginadesolator:
New extension for [imhentai.com](https://imhentai.com) (NSFW)

Supported features:
- Popular mangas
- Latest mangas
- Search mangas by keyword
- Sort desc by popularity, date of upload, downloads and rating
- Filter by language (source language is excluded from filter view as it is locked as enabled)
- Filter by category


src/all/imhentai/src/eu/kanade/tachiyomi/extension/all/imhentai/IMHentai.kt
```diff
@@ -0,0 +1,242 @@
...
+
+    private fun getLanguageURIs() = listOf(
+        SearchFlagFilter("English", "en"),
+        SearchFlagFilter("Japanese", "jp"),
+        SearchFlagFilter("Spanish", "es"),
+        SearchFlagFilter("French", "fr"),
+        SearchFlagFilter("Korean", "kr"),
+        SearchFlagFilter("German", "de"),
+        SearchFlagFilter("Russian", "ru")
+    )
```
arkon, line 241: Is it possible to make this extension a factory for the different languages? It's misleading to have it labelled as "all" in the extensions list, but only English in the actual source, then have filters for more languages.

vaginadesolator: Reworked as extension factory. Retained the language filter but re-labelled as "Other Languages" so user has option to view results for current source language (not listed as filter) + other languages (disabled by default).

PR #5452, opened 2021-01-16
Add Nsfw flag to Bato.to

jopejoe1:
Bato.to has NSFW content


arkon:
Thanks! Can you add the annotation to the source factory as well?

jopejoe1:
Sure No problem.

PR #5497, opened 2021-01-20
Update Gmanga.kt

ellpatron:
fixing when hitting latest ---> getting jsonNull

### DIFF
--- a/src/ar/gmanga/build.gradle
+++ b/src/ar/gmanga/build.gradle
@@ -5,7 +5,7 @@ ext {
     extName = 'GMANGA'
     pkgNameSuffix = 'ar.gmanga'
     extClass = '.Gmanga'
-    extVersionCode = 21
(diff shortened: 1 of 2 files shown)


CarlosEsco:
you need to update the build.gradle version also

ellpatron:
> you need to update the build.gradle version also

Sorry i forget thinks

src/ar/gmanga/build.gradle
```diff
@@ -5,7 +5,7 @@ ext {
     extName = 'GMANGA'
     pkgNameSuffix = 'ar.gmanga'
     extClass = '.Gmanga'
-    extVersionCode = 2
+    extVersionCode = 21
```
CarlosEsco, line 8: only bump it by one.  So this should be 3 now

CarlosEsco (changes requested):

PR #5502, opened 2021-01-21
<dmzj>: fix url

OncePunchedMan:
Closes #5501 

i fixed the urls for getting all info, but it seems like they didn't change the chapter urls on their site:

_should be images.dmzj1._

you can't read on the website either. so we have to wait until they change it. extension is ready tho

PR #5512, opened 2021-01-22
<zh-dmzj> Fix url replace rule and parse HD images in v3api.

Oldwangtouchtouchdoge:
1. Delete redundant url replace rule: https://github.com/tachiyomiorg/tachiyomi-extensions/pull/5502#issuecomment-764735616
2. Get HD comic images instead of compressed ones in v3api.


ZerOri:
请问有办法修复dmzj插件无法跳转漫画原网页、分享漫画原网址的bug么？

Issue #5511, opened 2021-01-22
[Feature Request] <Desu, Remanga...>: <redirect from api>


e-shl:
This is not really a suggestion, but I would like to know how to redirect from api to a page with manga?
For these sources 
https://github.com/tachiyomiorg/tachiyomi-extensions/blob/master/src/ru/desu/src/eu/kanade/tachiyomi/extension/ru/desu/Desu.kt
https://github.com/tachiyomiorg/tachiyomi-extensions/blob/master/src/ru/remanga/src/eu/kanade/tachiyomi/extension/ru/remanga/Remanga.kt

mlneko:
![image](https://user-images.githubusercontent.com/70600003/105464774-b8cb9200-5c9a-11eb-88ac-6bd1929e102c.png)
![image](https://user-images.githubusercontent.com/70600003/105464784-bcf7af80-5c9a-11eb-8f9a-214286247ce1.png)
сам задал - сам ответил))
по котлину не шарю, так бы кинул готовое
получаешь из json-а выделенный dir и подставляешь в формате https://remanga.org/manga/{dir}

mlneko:
с desu.me ещё проще:
![image](https://user-images.githubusercontent.com/70600003/105465783-0bf21480-5c9c-11eb-9f93-3c00cb1ae21a.png)

OncePunchedMan:
you want the source not to use the apis?

you'd have to change the urls and some other functions. basically you could just copy from another source and set the html tags and urls and so on correctly

e-shl:
@OncePunchedMan, no, I want the api to be used, but open the manga page via webview or share

OncePunchedMan:
i see

arkon:
You can refer to how MangaPlus does it: https://github.com/tachiyomiorg/tachiyomi-extensions/blob/b32647e7023b8bc13063f68bacf58edb01c1b5ea/src/all/mangaplus/src/eu/kanade/tachiyomi/extension/all/mangaplus/MangaPlus.kt#L188-L195

Issue #5524, opened 2021-01-23
[Bug] <动漫之家，漫画人>: <无法在webview中打开原网页>


ZerOri:
## Device information
- Tachiyomi version: 0.10.17
- Android version: 10

## Source information
- Name: 动漫之家，漫画人
- Extension version: 1.2.14，1.2.7

### Expected behavior
无法在网页中打开漫画源网页（显示一堆脚本代码而非跳转漫画页面），同样也无法分享漫画正确网址，见图

### Actual behavior
请求修复bug，让分享漫画网址功能恢复正常，另外，动漫之家被错误标记了R18标志，请求移除。

![QQ图片20210123211152](https://user-images.githubusercontent.com/46775615/105579151-b151d900-5dbf-11eb-8a8c-5844a59da8eb.jpg)

Oldwangtouchtouchdoge:
动漫之家的webview和分享我试着修了, 然后r18内容仔细想想也夹的差不多了，我之前加nsfw标志是有点草率😂

ZerOri:
> 动漫之家的webview和分享我试着修了, 然后r18内容仔细想想也夹的差不多了，我之前加nsfw标志是有点草率😂

嗯，在大陆运营的漫画网站，R15都不可能有，更别说R18了，就算漫画中有和谐内容远比不上真正R18的成年漫画。

ringosham:
You should have rewritten this issue in English. This is probably why this issue went unnoticed.
About the bug:
There needs to be a change in the underlying framework. The WebView URL (The one that opens up the browser) and the API URL (That is used to look up manga) is completely different.
For example (One Punch Man):
WebView: https://manhuaren.com/manhua-yiquanchaoren/
API: http://mangaapi.manhuaren.com/v1/manga/getDetail?mangaId=10684

The WebView API can only be obtained during fetching manga details, but by that point the manga URL is fixed by the Tachiyomi framework. If we try to get the WebView API earlier (During listing latest/popular/search manga), we would be spamming the API for every single entry. That's a no go too.

ZerOri:
[...]

PR #5537, opened 2021-01-24
<zh-dmzj>Add ratelimit to fix HTTP 429 and return webpage url to "Open in browser" and "Share manga".

Oldwangtouchtouchdoge:
1. Add ratelimit to fix HTTP 429 Too Many Requests when updating library.
2. Add workaround to allow "Open in browser" and "Share manga" use webpage url. Like #5511.
<del>3. Rewrite the rate limit interceptor that only work for manhuagui, now it can work for any url and handles multiple url host's rate limiting separately. Since extensions can't override fetchImage function, this interceptor is a workaround to allow web page urls and imageCDN urls can have different rate limit setting.</del>


Oldwangtouchtouchdoge:
About removing nsfw mark, #5524 request to remove it, and strictly speaking, the R18+ contents in dmzj are mostly unvisitable, banned or censored now, so I think I was a bit hasty to add nsfw mark before.

ZerOri:
感谢修复~😀

Oldwangtouchtouchdoge:
I still have something need to change🤣

Oldwangtouchtouchdoge:
An interceptor that only handle one specific url host and add multiple these interceptors to an okhttp client should be more elegant than an interceptor accept an array of url hosts and handle them one by one.🤣

Oldwangtouchtouchdoge:
Tuned up v3api's default rate limit permits from 2 to 5 per second.

PR #5541, opened 2021-01-24
[Ru]Remanga. Fix chapter number

Pavkazzz:
Fix chapter number for remanga

### DIFF
--- a/src/ru/remanga/build.gradle
+++ b/src/ru/remanga/build.gradle
@@ -5,7 +5,7 @@ ext {
     extName = 'Remanga'
     pkgNameSuffix = 'ru.remanga'
     extClass = '.Remanga'
-    extVersionCode = 15
+    extVersionCode = 16
     libVersion = '1.2'
 }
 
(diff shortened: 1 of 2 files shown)


arkon:
You could also just not setting it and try to let the app infer it.

Pavkazzz:
> You could also just not setting it and try to let the app infer it.
Remanga api response with string of chapter number. Currently we have a problem with numbers like "1.2.3". Any suggestion to how truncate it to float?

arkon:
What exactly does that mean? Is it including a volume number?

Pavkazzz:
> What exactly does that mean? Is it including a volume number?

I doesn't realy know for sure. Maybe it's a mistake, maybe some of chapters devided by 2 parts.

e-shl:
https://github.com/tachiyomiorg/tachiyomi-extensions/pull/5536

arkon:
I'm fine with this change if it seems to function as expected for you.

Pavkazzz:
@arkon thx! Can u merge?)

PR #5569, opened 2021-01-27
scantrad useragent2

Riztard:
try hexa random that will/should be close to/similar if use pc/normal


CarlosEsco:
Can you also a severe rate limiter

Riztard:
like that?

CarlosEsco:
yep that works

PR #5577, opened 2021-01-28
scantrad userAgent3

Riztard:
look like the random int only called once and being used again causing repeated number


CarlosEsco:
Can you also include a referrer

Riztard:
like this?

raphaelboulch:
It seems that they found another way to block Tachiyomi users.

Issue #5399, opened 2021-01-13
[Bug] Bato.to: Filter not working fully and missing stuff


jopejoe1:
## Device information
Tachiyomi fork: TachiyomiSY
App version: 1.4.1 (standard, fca7dad7b, 12)
Preview build: 324
Android version: 10 (SDK 29) 
Android build ID: QP1A.190711.020.G975FXXS9DTI8
Device brand: samsung
Device manufacturer: samsung
Device name: beyond2
Device model: SM-G975F
Device product name: beyond2ltexx

## Source information
- Name: Bato.to
- Extension version: 1.21

## Steps to reproduce
1. Open Bato.to Extension
2. Click on Filter button
3. Try any of the filter option
4. try another one or another few

### Expected behavior
Different results each time or most of the time

### Actual behavior
Same results most of the time (more info in other details)

## Other details
Some seem to work but most don't seem to work.
Working: 
- Chapters

Not Woking: 
- Stars (Can't find this filter on website and does nothing in the extension)
- Author (is the same search field as normal search but the one in the filter does nothing)
- Styles & Demographic (Seem to be under genre and are not Woking in current extension)

Missing: 
- Original
- Status

Partly working: 
- Genres(some missing, and no 3 level select, would recommend splitting into four sub categories of Styles, Demographic, Content and Genres)
- Sort By (A-Z is in reverse and totally is displaying the sort of Create Date instead of the total sort)

Other:
- Ther is a Completed Tick box no idea what that for when I tick it, it shows nothing in my result

PR #5618, opened 2021-02-01
Tri-state genre filtering for Batoto

pappleby:
* include or exclude genres
* Fixes typo for Harlequin genre
* Update build.gradle

Adds additional feature mentioned in #5399


src/all/batoto/src/eu/kanade/tachiyomi/extension/all/batoto/BatoTo.kt
```diff
@@ -94,15 +94,24 @@ open class BatoTo(
                     is GenreFilter -> {
                         val genreToInclude = mutableListOf<String>()
                         filter.state.forEach { content ->
-                            if (content.state) {
+                            if (content.isIncluded()) {
                                 genreToInclude.add(content.name)
                             }
                         }
```
arkon, line 100:
```kt
val genreToInclude = filter.state
  .filter { it.isIncluded() }
  .map { it.name }
```

And likewise for below.

pappleby: Change made, thanks for the tip!

PR #5662, opened 2021-02-07
Madara split: part 1

AriaMoradi:
This covers about 50 of madara sources, can be merged now.

I'll do more of it when I have time.


arkon:
Can you remove the stuff from the Madara source rather than commenting them out?

AriaMoradi:
I need the list for cross referencing, but i can remove the class implementations

Soitora:
How would one know their source in Madara has been off-sourced when the extension itself isn't gone?

AriaMoradi:
> How would one know their source in Madara has been off-sourced when the extension itself isn't gone?

migration tab shows id instead of name

AriaMoradi:
removed moved classes.

arkon:
You might want to try splitting the other smaller ones first and doing Madara last.

arkon:
Read #5672.

PR #5667, opened 2021-02-07
MangAdventure split

ObserverOfTime:
Do not merge until the source generator supports URL intents.

They can be implemented as a list of `Triple(host, pathPattern, scheme)` in `ThemeSourceData` so the generator can add them to the manifest. `ThemeSourceData` will also need another field that declares the name of the activity class.

Also, it would be nice if the generator could replace hyphens and other invalid identifier characters in the name.


arkon:
Were you planning on making the change to the generator, or hoping that @AriaMoradi does it?

ObserverOfTime:
I can do it next weekend if no one else does it by then.

AriaMoradi:
> They can be implemented as a list of `Triple(host, pathPattern, scheme)` in `ThemeSourceData` so the generator can add them to the manifest. `ThemeSourceData` will also need another field that declares the name of the activity class.

This multi-source is maybe the only one that needs this kind of feature so I'm against adding it to the library. 

> Also, it would be nice if the generator could replace hyphens and other invalid identifier characters in the name.

The Source name can be any string of Unicode characters,  and adding this feature doesn't seem wise to me, you can already provide pakage name and file(class) name overrides manually. 
Also src override gives you all the power you need to pull off this split. 
The `ThemeSource` just acts as a definition tool for sources which essentially replaces the factory class and it does a minimal source generation for sources that don't need overrides.

arkon:
> This multi-source is maybe the only one that needs this kind of feature so I'm against adding it to the library.

Are you saying this can be achieved with the overrides rather than having to modify the generator code, or are you against multi-source modules having the ability to handle intents?

AriaMoradi:
> > This multi-source is maybe the only one that needs this kind of feature so I'm against adding it to the library.
> 
> Are you saying this can be achieved with the overrides rather than having to modify the generator code, or are you against multi-source modules having the ability to handle intents?

Any shared source file should be put in the `multisrc.multisrcname` package and any source-specific code should be in the src override.
If you take a look at the logic, it copies any class file that doesn't end with `Generator.kt` (`ThemeSourceGenerator.kt` line 114)

arkon:
Sure, but it doesn't handle overriding the manifest, does it?

AriaMoradi:
> Sure, but it doesn't handle overriding the manifest, does it?

Ah, I completely forgot that manifest needs overrides to handle intents, oops.
we can add a `manifest` directory to `overrides` and handle it that way or put it in the src override directory and have the generator look for a manifest file?

ObserverOfTime:
That should work but you also need to handle the activity class.

AriaMoradi:
> That should work but you also need to handle the activity class.

Put inside `multisource.themename`? what's special about he activity class?

ObserverOfTime:
It just accepts the intent and sends it to the main app.

AriaMoradi:
> It just accepts the intent and sends it to the main app but it needs to be declared in `build.gradle`.

Noting is delared in https://github.com/tachiyomiorg/tachiyomi-extensions/blob/81f9eede50758261f46b888a518c126221860d74/src/all/mangadventure/build.gradle ?

ObserverOfTime:
Oh, right, it's in the manifest.

ObserverOfTime:
It should be ready to merge now.

multisrc/src/main/java/eu/kanade/tachiyomi/multisrc/mangadventure/MangAdventureGenerator.kt
```diff
@@ -0,0 +1,21 @@
...
+/** [MangAdventure] source generator. */
+class MangAdventureGenerator : ThemeSourceGenerator {
+    override val themePkg = "mangadventure"
+
+    override val themeClass = "MangAdventure"
+
+    override val baseVersionCode = 1
+
+    override val sources = listOf(
+        SingleLang("Arc-Relight", "https://arc-relight.com", "en", className = "ArcRelight"),
```
AriaMoradi, line 15: Only one source?!
Why are you doing multisrc then?!

AriaMoradi (changes requested):

multisrc/src/main/java/eu/kanade/tachiyomi/multisrc/mangadventure/MangAdventureGenerator.kt
ObserverOfTime: Only one source *now*. There was another one that changed to Madara and there may be more in the future.

PR #5695, opened 2021-02-08
improve tasks, support AndroidManifest override

AriaMoradi:
- simplifies running the generators
- `AndroidManifest.xml` override support for #5667

### DIFF
--- a/multisrc/src/main/java/eu/kanade/tachiyomi/multisrc/ThemeSourceGenerator.kt
+++ b/multisrc/src/main/java/eu/kanade/tachiyomi/multisrc/ThemeSourceGenerator.kt
@@ -68,12 +68,16 @@ interface ThemeSourceGenerator {
             """.trimIndent())
         }
 
-        private fun writeAndroidManifest(androidManifestFile: File) {
-            androidManifestFile.writeText("""
+        private fun writeAndroidManifest(androidManifestFile: File, manifestOverridesPath: String, source: ThemeSourceData, projectRootPath: String) {
+            val androidManifestOverride = File("$manifestOverridesPath/${source.pkgName}")
+            if (androidManifestOverride.exists())
+                androidManifestOverride.copyRecursively(File(projectRootPath))
+            else
+                androidManifestFile.writeText("""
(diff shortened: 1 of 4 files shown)


multisrc/src/main/java/eu/kanade/tachiyomi/multisrc/ThemeSourceGenerator.kt
```diff
@@ -82,6 +86,7 @@ interface ThemeSourceGenerator {
             val overridesPath = "$userDir/multisrc/overrides" // userDir = tachiyomi-extensions project root path
             val resOverridesPath = "$overridesPath/res/$themePkg"
             val srcOverridesPath = "$overridesPath/src/$themePkg"
+            val manifestOverridesPath = "$overridesPath/manifest/$themePkg"
```
arkon, line 89: This is fine for now, but I feel like overrides are distributed across too many disparate folders. It'd probably make more sense to have something like:

```
overrides
|- [themeName]
  |- [sourceName]
    |- src
    |- res
    |- AndroidManifest.xml
  |- 
```

AriaMoradi: that looks better indeed, but rearranging the files is something I'm not wiling to do!
If I could automate it with a script or sth, I'll pr it.

ObserverOfTime: Here's a quick bash script that will rearrange the files when ran from the overrides folder.
```bash
for d in {src,res}/*/*; do
    mkdir -p "${d#*/}"
    mv "$d" "$_/${d%%/*}"
done

rm -r src res
```

AriaMoradi: You are a bit late, it's already done.

ObserverOfTime: ¯\\\_(ツ)\_/¯

PR #5696, opened 2021-02-08
No need to scope the task name

AriaMoradi:
something that was missed from the last PR.


arkon:
Wouldn't this cause Gradle to unnecessarily check for the task in every other module?

AriaMoradi:
I will, but I don't think it'll affect the build time?

AriaMoradi:
close the PR if you think it's not good.

PR #5711, opened 2021-02-08
Remove paprika

AriaMoradi:
All domains are down.
confirmed by 3 people.


arkon:
These sites still load for me:

- http://mangastream.xyz
- http://readmangafox.xyz
- http://manganelos.com
- http://mangadogs.fun
- http://mangareader.cc/

AriaMoradi:
Spreadsheet links are https, thats why none where loading, they don't have https.

arkon:
The others in the factory _are_ seemingly dead though.

Soitora:
Updated the spreadsheet to fix https and tell which of these were down

PR #5713, opened 2021-02-08
Missing nsfw directives

AriaMoradi:


arkon:
You don't actually need to bump the versions since it only affects the extensions list, and there's no need for people to update.

PR #5783, opened 2021-02-11
fixed missing css selector Rain of snow

jopejoe1:


CarlosEsco:
you need to update the gradle ext version

PR #5800, opened 2021-02-12
Fmreader split

AriaMoradi:
**Note:** `Manhwa18.net Raw` should be migrated to `Manhwa18.net` (the Korean source).


multisrc/src/main/java/eu/kanade/tachiyomi/multisrc/fmreader/FMReaderGenerator.kt
```diff
@@ -23,8 +24,7 @@ class FMReaderGenerator : ThemeSourceGenerator {
             SingleLang("LHTranslation", "https://lhtranslation.net", "en"),
             SingleLang("Manga-TR", "https://manga-tr.com", "tr", className = "MangaTR"),
             SingleLang("ManhuaScan", "https://manhuascan.com", "en"),
-            SingleLang("Manhwa18", "https://manhwa18.com", "en"),
-            SingleLang("Manhwa18.net Raw", "https://manhwa18.net", "ko", className = "Manhwa18NetRaw"),
+            MultiLang("Manhwa18.net", "https://manhwa18.net", listOf("en", "ko"), className = "Manhwa18NetFactory"),
             SingleLang("Manhwa18.net", "https://manhwa18.net", "en", className = "Manhwa18Net"),
```
Riztard, line 28: removing the wrong one?

AriaMoradi: It's the same site, just the original dev added a  default `raw` only filter to the second one and a Not `raw` filter to the second one.

AriaMoradi: we could merge them, but its' just annoyance to the correct user base.

Riztard: i mean the bottom one

Riztard: uh add back the .com ?

AriaMoradi: done

PR #5833, opened 2021-02-13
fix manhwa.club's name

AriaMoradi:
Requested by Ichigo Kurosaki on Discord.

### DIFF
--- a/multisrc/src/main/java/eu/kanade/tachiyomi/multisrc/madara/MadaraGenerator.kt
+++ b/multisrc/src/main/java/eu/kanade/tachiyomi/multisrc/madara/MadaraGenerator.kt
@@ -146,7 +146,7 @@ class MadaraGenerator : ThemeSourceGenerator {
             SingleLang("Manhwa Raw", "https://manhwaraw.com", "ko"),
             SingleLang("Manhwatop", "https://manhwatop.com", "en"),
             SingleLang("Manwahentai.me", "https://manhwahentai.me", "en", className = "ManwahentaiMe", isNsfw = true),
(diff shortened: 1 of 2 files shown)


arkon:
Backwards compatibility with the version pre-split? This will change the ID.

AriaMoradi:
This time I tested it thoughtfully and made sure it works backwards compatible properly.

multisrc/src/main/java/eu/kanade/tachiyomi/multisrc/madara/MadaraGenerator.kt
```diff
@@ -146,7 +146,7 @@ class MadaraGenerator : ThemeSourceGenerator {
             SingleLang("Manhwa Raw", "https://manhwaraw.com", "ko"),
             SingleLang("Manhwatop", "https://manhwatop.com", "en"),
             SingleLang("Manwahentai.me", "https://manhwahentai.me", "en", className = "ManwahentaiMe", isNsfw = true),
-            SingleLang("Manwha Club", "https://manhwa.club", "en"),
+            SingleLang("Manhwa.club", "https://manhwa.club", "en", className="ManwhaClub", overrideVersionCode = 1), // wrong class name for backward compatibility
```
arkon, line 149: I'm still confused as to why the class name needs to be wrong?

AriaMoradi: The gradle module is named after the pkgName which is named after className by default.
look at uses of `pkgNameSuffix`

PR #5840, opened 2021-02-13
Split FoolSlide Extension

Riztard:


multisrc/overrides/foolslide/hentaicafe/src/HentaiCafe.kt
```diff
@@ -1,22 +1,21 @@
...
 import org.jsoup.nodes.Document
 import org.jsoup.nodes.Element
+import eu.kanade.tachiyomi.network.asObservable
+import eu.kanade.tachiyomi.source.model.Filter
+import eu.kanade.tachiyomi.source.model.MangasPage
+import eu.kanade.tachiyomi.util.asJsoup
 import rx.Observable
 import java.net.URLEncoder
 
-@Nsfw
```
arkon, line 19: Why removing this?

Riztard: already add nsfw in generator?

arkon: That only affects the gradle file and generated class. If you're overriding the class, you still need to have it explicitely.

Riztard: naruhodo

PR #5857, opened 2021-02-15
Change mangadex api url

az4521:
i havent even tested this, they were just making fun of us in the mangadex discord for using the wrong url


arkon:
Sorry if this wasn't a shitpost and MD actually has a plan.

az4521:
it was a shitpost that wouldve actually fixed the extension

arkon:
"Fix" is sort of a stretch if their main webserver(s) die again from the load. Did they actually say something about it?

az4521:
nah this is on the level of the shared-account fix for searching

Issue #5850, opened 2021-02-14
[Bug] Scantrad: 403 Error


vic-blt:
## Device information
- Tachiyomi version: 0.10.9
- Android version: 10

## Source information
- Name: Scantrad [FR]
- Extension version: 1.2.12

## Steps to reproduce
1. Install Scantrad extension
2. Go to **Browse** tab
3. Open Scantrad source

### Expected behavior
It should list mangas available on Scantrad website

### Actual behavior
There's a 403 HTTP error

### Related issues
#5243, #5560, #5322

OncePunchedMan:
they are actively blocking tachi and we tried fixing scantrad by adding this useragent

vic-blt:
I know, I've read the issues regarding the situation with Scantrad's admins.
But currently, the headers have to be removed in order to get rid of the 403 HTTP error and be able to fetch Scantrad URLs.
Maybe it's the setting of the referer which causes the 403 HTTP error.
Could you test with the setting of the user-agent header only ?

vic-blt:
Well that was quick, tachiyomi is blocked again.

PR #5885, opened 2021-02-16
Generate a totally random user-agent

vic-blt:
Related to #5850


ObserverOfTime:
Just so you know, this user agent can easily be blocked with a regex: `[a-zA-Z0-9]{10,30}`.

vic-blt:
For sure, but it would also match valid user agents and block users reading from browsers.

ObserverOfTime:
Browser user agents aren't usually between 10 and 30 characters.

vic-blt:
User agents aren't only about browsers.
But I agree with you that it's not effective at all.
For proof it's been blocked easily.

PR #5900, opened 2021-02-17
MangaMutiny: WebView + pagination fix

E3FxGaming:
- WebView now works (main page + pages of individual manga)
- changed baseUrl to https://mangamutiny.org and moved https://api.mangamutiny.org to different variable
(the extension still relies on the lightweight json responses from the API for everything that isn't WebView related - the baseUrl only had to be changed for WebView to work)
- fixed pagination bug (would previously skip 21 manga after the first 21 manga, now fixed)
- Internally reworked filters to make them compatible with the WebView changes (does not affect users in any way)

### DIFF
--- a/src/en/mangamutiny/build.gradle
+++ b/src/en/mangamutiny/build.gradle
@@ -5,7 +5,7 @@ ext {
     extName = 'Manga Mutiny'
     pkgNameSuffix = "en.mangamutiny"
     extClass = '.MangaMutiny'
-    extVersionCode = 4
+    extVersionCode = 5
     libVersion = '1.2'
     containsNsfw = true
 }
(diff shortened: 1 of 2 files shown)


arkon:
I'm not sure I understand what the WebView changes are meant to be doing. Is the concern just being able to open the site from the manga info view?

E3FxGaming:
>  Is the concern just being able to open the site from the manga info view?

No. WebView doesn't work at all with the currently released version of the extension. Due to the API url being assigned to the baseUrl, users were only ever shown json responses that had nothing in common with what the website actually looked like.

This left users confused about how they can get it to work.

WebView opened from browse + latest, as well as from the manga info view would be addressed with the proposed change.

PR #6051, opened 2021-03-03
MangaMainac: filter out latest chapter if it's not released yet, update URLs

scb261:
- filter unreleased chapters (the same as in #6050)
- update two URLs (old manga in library will still work with http or redirect)

No opened issues

Side note for future: `The Promised Neverland` and `The Quintessential Quintuplets` sites are down, I couldn't find new links to them and if they will remain like this for long time, they should be removed


CarlosEsco:
feel free to break this out of multi source :)

PR #6052, opened 2021-03-03
Remove TCBScans since the site is present in MangaMainac

### LINKED ISSUE #5896: [Meta] TCB Scans duplicated
##  Why/Project's Benefit/Existing Problem
Explain why this should be added

TCB Scans has its own extension but is also part of the sites that are included in the MangaMainac extension.
## Other Details

scb261:
I looked at MangaMainac and it's actually implemented as a single source with 15 manga, each one has its own site. So I don't think it needs to be split and therefore my concern in #5896 isn't valid.

Users will need to [migrate](https://tachiyomi.org/help/guides/source-migration/#source-migration) from TCBScans to MangaMainac.

Closes #5896

Feel free to reject if it actually needs a split as one source per one manga (I'm not going to do anything with new multisrc for now).


CarlosEsco:
Yea i feel this should be split, no reason to make things hard to find again

PR #6066, opened 2021-03-04
fix Mangalife chapter numbers

curche:
This PR changes how chapter number is shown in MangaLife
chapterRegex was taken from MangaSee extension code (which share most of their code already)
Have been using the extension for some time and its working fine

| before | after |
|----|---|
| "Chapter 0142" | "Chapter 142" |
| Screenshot_1614798631 | Screenshot_1614803771 |

### DIFF
--- a/src/en/mangalife/build.gradle
+++ b/src/en/mangalife/build.gradle
@@ -5,7 +5,7 @@ ext {
     extName = 'MangaLife'
     pkgNameSuffix = 'en.mangalife'
     extClass = '.MangaLife'
-    extVersionCode = 10
+    extVersionCode = 11
     libVersion = '1.2'
 }
 
--- a/src/en/mangalife/src/eu/kanade/tachiyomi/extension/en/mangalife/MangaLife.kt
+++ b/src/en/mangalife/src/eu/kanade/tachiyomi/extension/en/mangalife/MangaLife.kt
@@ -220,8 +220,10 @@ class MangaLife : HttpSource() {
         return "-chapter-$n$index$suffix.html"
     }
 
-    private fun chapterImage(e: String): String {
-        val a = e.substring(1, e.length - 1)
+    private val chapterImageRegex = Regex("""^0+""")
+
+    private fun chapterImage(e: String, cleanString: Boolean = false): String {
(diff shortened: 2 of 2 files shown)


CarlosEsco:
Problem with changing this is all downloads will be orphaned

curche:
Strangely enough, I had installed the apk and was using it during the day and it worked fine
none of my previously downloaded files were affected
thus, I sent this PR

I could be wrong here but in case a third person can check and confirm, do try the apk from the [zip](https://github.com/tachiyomiorg/tachiyomi-extensions/files/6086373/tachiyomi-en.mangalife-v1.2.11-debug.zip)

I'm okay with closing the PR (since its anyways a minor thing) in case the downloads are getting orphaned

curche:
Got to test it on Stable in a different device
The downloads were orphaned. However, download badges were still the same
Have to move the downloads to local if I want to use them

As for my device, it was my wrong observation. Actually, I was using Sy (which seems to be renaming my downloads). I'll have to report that, I guess
Sorry for that

Closing this PR for now

curche:
*Seems like I have been proven wrong (again).*

The downloads wouldn't be orphaned since this PR doesn't change the source url, just the chapter name in the chapter list (as the pic above shows)

This has been confirmed by the following - 
Downloads get renamed when chapter list changes
ref here:
https://github.com/tachiyomiorg/tachiyomi/blob/54dc01253d419079eaec61333d4c6f4deb5a58f8/app/src/main/java/eu/kanade/tachiyomi/util/chapter/ChapterSourceSync.kt#L70-L72

Chapters are fetched by url and changing chapter name wouldn't cause an issue
ref here:
https://github.com/tachiyomiorg/tachiyomi/blob/54dc01253d419079eaec61333d4c6f4deb5a58f8/app/src/main/java/eu/kanade/tachiyomi/util/chapter/ChapterSourceSync.kt#L56

Thanks to @scb261 for the explanation and @Seishirou101 for testing it to check that it works smoothly

arkon:
I think they'll get orphaned in J2K since I'm pretty sure it doesn't have the renaming logic, but at least other versions aren't affected.

Seishirou101:
it gets renamed on j2k too

PR #6077, opened 2021-03-06
MangaAction  / FIX URL

zeredouni:


multisrc/src/main/java/eu/kanade/tachiyomi/multisrc/madara/MadaraGenerator.kt
```diff
@@ -87,7 +87,7 @@ class MadaraGenerator : ThemeSourceGenerator {
             SingleLang("مانجا العاشق", "https://3asq.org", "ar", className = "Manga3asq"),
             SingleLang("Manga3S", "https://manga3s.com", "en"),
             SingleLang("Manga68", "https://manga68.com", "en"),
-            SingleLang("Manga Action", "https://manga-action.com", "ar"),
+            SingleLang("Manga Action", "https://manga-action.com", "ar", overrideVersionCode = 1),
```
jobobby04, line 90: Update url here?

Riztard: it getting overrides so its not being used, but ¯\_(ツ)_/¯

CR1M50:
Guys you need to change Imhentai.com to Imhentai.xxx now...They suddenly changed their website url for some reason...

PR #6089, opened 2021-03-07
moved gecenin lordu from madara to mangareader

### LINKED ISSUE #6081: [Bug] < Gecenin Lordu>: < source does not open>
## Device information
- Tachiyomi version: 10.9
- Android version: 9

## Source information
- Name: Gecenin Lordu
- Extension version: 1.2.1

## Steps to reproduce
1. First Step
2. Second Step

### Expected behavior
Describe what should have happened

### Actual behavior
Source does not open

OncePunchedMan:
Closes #6081

### DIFF
--- a/multisrc/src/main/java/eu/kanade/tachiyomi/multisrc/madara/MadaraGenerator.kt
+++ b/multisrc/src/main/java/eu/kanade/tachiyomi/multisrc/madara/MadaraGenerator.kt
@@ -10,7 +10,7 @@ class MadaraGenerator : ThemeSourceGenerator {
 
     override val themeClass = "Madara"
 
-    override val baseVersionCode: Int = 1
+    override val baseVersionCode: Int = 2
 
     override val sources = listOf(
             SingleLang("Adonis Fansub", "https://manga.adonisfansub.com", "tr"),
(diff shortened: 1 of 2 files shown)


multisrc/src/main/java/eu/kanade/tachiyomi/multisrc/madara/MadaraGenerator.kt
```diff
@@ -10,7 +10,7 @@ class MadaraGenerator : ThemeSourceGenerator {
 
     override val themeClass = "Madara"
 
-    override val baseVersionCode: Int = 1
+    override val baseVersionCode: Int = 2
```
scb261, line 13: I think this bumps version for all Madara/WPMangaReader sources, you should use `overrideVersionCode` instead

Riztard: yeah

OncePunchedMan: i should leave baseversion of madara as is, cause i removed a source. and for mangareader use overrideversioncode, right?

PR #6091, opened 2021-03-08
LANraragi: Random item, clear new status

RePod:
- Adds a "Random" entry to popular's MangasPage, assuming nothing else prevents it.
 - Clears an archive's new status on the server similar to two versions ago before moving to API endpoints.
   - The web reader happened to be taking care of it

### DIFF
--- a/src/all/lanraragi/src/eu/kanade/tachiyomi/extension/all/lanraragi/LANraragi.kt
+++ b/src/all/lanraragi/src/eu/kanade/tachiyomi/extension/all/lanraragi/LANraragi.kt
@@ -15,6 +15,7 @@ import eu.kanade.tachiyomi.extension.all.lanraragi.model.ArchiveSearchResult
 import eu.kanade.tachiyomi.extension.all.lanraragi.model.Category
 import eu.kanade.tachiyomi.network.GET
 import eu.kanade.tachiyomi.network.POST
+import eu.kanade.tachiyomi.network.asObservable
 import eu.kanade.tachiyomi.source.ConfigurableSource
 import eu.kanade.tachiyomi.source.model.Filter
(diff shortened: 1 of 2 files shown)


src/all/lanraragi/src/eu/kanade/tachiyomi/extension/all/lanraragi/LANraragi.kt
```diff
@@ -53,19 +55,47 @@ open class LANraragi : ConfigurableSource, HttpSource() {
...
+        if (tachiAZ == 0) tachiAZ = 1
+
+        return archiveToSManga(archive)
     }
 
     override fun chapterListRequest(manga: SManga): Request {
-        // Upgrade the LRR reader URL to the API metadata endpoint
-        // without breaking WebView (i.e. for management).
+        // In every other situation this beats mangaDetailsParse and can't be AZ
+        if (tachiAZ == 0) tachiAZ = 2
```
CarlosEsco, line 94: find another way to do this, or PR az.  Extensions aren't having fork specific stuff

CarlosEsco (changes requested):

src/all/lanraragi/src/eu/kanade/tachiyomi/extension/all/lanraragi/LANraragi.kt
RePod: I'd be fine with reverting it. The PR and AZ will function without it just the AZ experience will be slightly off compared to the rest.

arkon: Should revert that, not worth the technical debt of dealing specifically with AZ.

Issue #6100, opened 2021-03-08
[Bug]  <NeoxScanlator>: <URL changed>


matheuswr89:
## Device information
- Tachiyomi version: 10.9
- Android version: 10

## Source information
- Name: NeoxScanlator
- Extension version: 1.2.1

## Steps to reproduce
1. Open The tachiyomi app
2. Enter the NeoxScanlator extension

### Expected behavior
It is expected that the manga of this extension will load

### Actual behavior
![Screenshot_20210308-122804](https://user-images.githubusercontent.com/44068234/110342866-a6b77000-800a-11eb-9c35-b8eb45a79641.png)

## Other details
In discord with the scan from which the extension comes, it says that the URL for accessing the site has changed:
![Screenshot_20210308-123615~2](https://user-images.githubusercontent.com/44068234/110343153-f433dd00-800a-11eb-9b66-f0e82169753d.png)

Poiares:
é verdade eles mudaram a URL.

Poiares:
precisamos de ajuda urgente tem muito capítulo novo que precisamos ler.

Poiares:
Help me Leon

Poiares:
Please i need a help

Poiares:
HEEELLPPPP

PR #6106, opened 2021-03-09
Update URL NeoxScanlator

matheuswr89:
Extension URL change as mentioned in #6100


scb261:
You also need to add `overrideVersionCode = 1`, see example here https://github.com/tachiyomiorg/tachiyomi-extensions/pull/6005/files. And while you did change URL in the right place, you should also update it here to avoid confusion for future contributors
https://github.com/tachiyomiorg/tachiyomi-extensions/blob/7d4e7437bd13314c6c8f7aef00f78bad1f86a666/multisrc/src/main/java/eu/kanade/tachiyomi/multisrc/madara/MadaraGenerator.kt#L171

matheuswr89:
Obsolet

Issue #5938, opened 2021-02-20
[Meta] Mangairo url


runboy93:
Link to Mangairo is not "https://m.mangairo.com", it should be "https://h.mangairo.com/" instead.

Extension v 1.2.1 currently.

Unlike Mangabat, Mangairo does redirect correctly even with older URL, but still there seem to be slight delay, compared to correct URL (tested web browser version)

scb261:
Was changed in #6115

PR #6115, opened 2021-03-09
Change mangabat and mangairo domains to correct website

### LINKED ISSUE #5937: [Meta] Mangabat url
Link to Mangabat is not "https://mangabat.com", like it is on extension (v 1.2.1) currently, it should be "https://m.mangabat.com/h", like it is really now.

On browser this causes extra redirect to correct website (slower start), until user can start check manga updates.

loocool2:
Closes #5937 and #5938 
Couldn't find the build.gradle file, don't remember if multisrc exts need it anymore


Seishirou101:
I believe you have to do an override like this example https://github.com/tachiyomiorg/tachiyomi-extensions/pull/6005/files

PR #6130, opened 2021-03-10
Fixed MangaBat URL

loocool2:


JustAchie:
Can you fix the mangabat Extension

PR #6150, opened 2021-03-11
Mangasee: use regex to find the page host

az4521:
as usual i havent even tested if this will compile. should be a more robust fix for mangasee.


az4521:
tested, it works

CarlosEsco:
Closing for now as see has said they will potentially unblock for now.

PR #6172, opened 2021-03-13
Mangadex: Add status parsing for completed titles

starg09:
Status 2 is of manga that is finished, think this should work fine? Can't currently build to test it though, if someone could confirm this being the proper solution and building properly it'd be appreciated.

Status 3 (Cancelled) and 4 (Hiatus) aren't supported in the extension itself, so those will keep saying Unknown Status for the time being.

### DIFF
--- a/src/all/mangadex/build.gradle
+++ b/src/all/mangadex/build.gradle
@@ -5,7 +5,7 @@ ext {
     extName = 'MangaDex'
     pkgNameSuffix = 'all.mangadex'
     extClass = '.MangaDexFactory'
-    extVersionCode = 105
+    extVersionCode = 106
     libVersion = '1.2'
     containsNsfw = true
 }
--- a/src/all/mangadex/src/eu/kanade/tachiyomi/extension/all/mangadex/MangaDex.kt
+++ b/src/all/mangadex/src/eu/kanade/tachiyomi/extension/all/mangadex/MangaDex.kt
@@ -700,6 +700,8 @@ abstract class MangaDex(
 
     private fun parseStatus(status: Int) = when (status) {
         1 -> SManga.ONGOING
+        2 -> SManga.COMPLETED
+        // TODO: Missing SManga values for Mangadex's API status values 3 (Cancelled) and 4 (Hiatus)
         else -> SManga.UNKNOWN
     }
 


scb261:
It already parses completed status to count in translation status for chosen language, canceled status and oneshots. Is there some particular case when it doesn't work as intended?
https://github.com/tachiyomiorg/tachiyomi-extensions/blob/849e4456fd50f9619fad81778c08219442a1486f/src/all/mangadex/src/eu/kanade/tachiyomi/extension/all/mangadex/MangaDex.kt#L443-L457

starg09:
> It already parses completed status to count in translation status for chosen language, canceled status and oneshots. Is there some particular case when it doesn't work as intended?
>

Missed that section of code, thanks for bringing it up. Not sure if I fully follow the logic (and in that case this PR is not needed), but as an example the title "Real Account" (https://mangadex.org/title/14219/real-account) is marked as complete inside mangadex itself, though still doesn't have all chapters translated in english. And in the app it shows as "Unknown status".

Same case with "Melt Away! Mizore-chan" (https://mangadex.org/title/31899/tokedase-mizore-chan). Is that intended until fully translated/available in the site?

> 
> As for building and testing - you can read https://github.com/tachiyomiorg/tachiyomi-extensions/blob/master/CONTRIBUTING.md

Thanks for the link as well. Meant I'm currently unable to setup the IDE in my computer atm, will try to do it in a couple days 😅

scb261:
Hmm, looks like this case isn't considered by the extension. Probably should change this section like this:
- `(status == 2 || status == 3) && chapterJson != null` is true, but there are no chapters in chosen language -> set `Unknown` (or skip, since it's covered by the last `else`
- `(status == 2 || status == 3) && chapterJson != null` is true, there are chapters in chosen language, `isMangaCompleted()` returns true -> set as `Completed`
- `(status == 2 || status == 3) && chapterJson != null` is true, there are chapters in chosen language, `isMangaCompleted()` returns false -> set as `Ongoing`

https://github.com/tachiyomiorg/tachiyomi-extensions/blob/849e4456fd50f9619fad81778c08219442a1486f/src/all/mangadex/src/eu/kanade/tachiyomi/extension/all/mangadex/MangaDex.kt#L443-L448

CarlosEsco:
Setting as ongoing when publication complete and still scanning would be weird.  Which is why it was unknown.  

As for OP. The reason publication complete doesn't map to complete is because then if you have don't update complete manga enabled in app, you won't get new  scans of the manga as it will skip being updated.

starg09:
Ahh, I see! Makes sense, thanks for the explanation. Will close this PR then 😅

PR #6186, opened 2021-03-16
CCC webcomic: support multi-image chapters

az4521:
in the shower rn so I can't test this but I think it'll work w/e


az4521:
just tested, works

PR #4567, opened 2020-10-09
Use scan status instead of published status

AbdullahM0hamed:
Closes #4548

PR #6267, opened 2021-03-24
Move MangaSee/Life to multisrc, improve search

### LINKED ISSUE #6259: [Bug] <MangaLife>: <Cannot search with Alternate Names>
## Device information
- Tachiyomi version: 0.10.9
- Android version: 8

## Source information
- Name: MangaLife
- Extension version: 1.2.16

## Steps to reproduce
1. Search for a manga using one of its alternate names.

### Expected behavior
Manga searched using one of its alternate names should appear in the search results.

### Actual behavior

Manga searched using one of its alternate names do not appear in the search results.

For instance, searching "Shounen no Abyss" would yield search results, whereas searching "Boy's Abyss" would not.

## Other details

Searching for a manga using one of its alternate names works in Tachiyomi using other extensions, such as Mangakakalot and MangaFreak.

Searching for a manga using one of its alternate names works using the MangaLife website itself.

scb261:
MultiSrc notes:
- MangaSee had fixed status, while MangaLife didn't (#4567).
- MangaSee had fixed some chapters, while MangaLife didn't (#5647).
- I think the only way to add rate limit is for each source separately, because of dependencies, correct me if I'm wrong.
- Not sure if the theme name is good enough, can't think of anything else.
- Tested with manga in library with downloaded chapters from old extensions - works fine.

Search (original PR #6265):
- Consider alternative names. Closes #6259.
- Trim search. It's trimmed in browser, but in tachi it would show no results with extra spaces. No opened issue.


scb261:
Why setting up jdk fails? It did fail on multisrc en, but why jdk?

CarlosEsco:
Maybe name it NepNep since who knows if he decides to add a third site :smile:

Riztard:
> I think the only way to add rate limit is for each source separately, because of dependencies, correct me if I'm wrong.

correct

but for the additional gradle you can put it on default 
multisrc/overrides/mangaseelife/`default`/additional.gradle.kts

scb261:
Some black magic fixed PR check. 

So should I change the theme name or nah?

CarlosEsco:
i vote yes, and the black magic was i manually clicked rerun :laughing:

Riztard:
> Rename theme name from MangaSeeLife to NepNep

😂😂

PR #6269, opened 2021-03-24
Updated Bato.to supported languages list

jopejoe1:
added en_us and unfiltered to language list
removed xh, hi and ku from language list


CarlosEsco:
need to update the build gradle

jopejoe1:
done

PR #6314, opened 2021-03-27
Duktape: removed unused method

AriaMoradi:


arkon:
Why? I don't really see a good reason to remove it since it _could_ be used.

AriaMoradi:
It's not used anywhere, we can bring it back later.

AriaMoradi:
closing as arkon is not keen on it...

PR #6319, opened 2021-03-28
change name and language code

Syrth:


scb261:
- You shouldn't move the file to a different directory
- When changing name and/or language, you should override id of the source with the old value, so users wouldn't need to migrate
- You should use `overrideVersionCode = 1` to bump the extension version

PR #6321, opened 2021-03-28
ScanManga: fix searching with special characters

papjul:


scb261:
You should also bump the extension version in the gradle file

papjul:
Thank you for pointing it to me, fixed x)

PR #5793, opened 2021-02-11
add inspector

AriaMoradi:
Generates a source-id mapping database.

example output: https://github.com/AriaMoradi/tachiyomi-extensions/blob/repo/inspector.txt

PR #6410, opened 2021-04-03
[RFC] Sources Inspector

AriaMoradi:
Continues #5793

Successful run: https://github.com/AriaMoradi/tachiyomi-extensions/actions/runs/714708497
result: https://github.com/AriaMoradi/tachiyomi-extensions/blob/repo/inspector.txt

-----
I can re-produce index.json and put sources inside it or it can become a separate csv file.

Sample output with sources inside index.json
```json
[
  {
    "name": "Tachiyomi: Bato.to",
    "pkg": "eu.kanade.tachiyomi.extension.all.batoto",
    "apk": "tachiyomi-all.batoto-v1.2.5.apk",
    "lang": "all",
    "code": 5,
    "version": "1.2.5",
    "nsfw": 1,
    "sources": [
      {
        "name": "Bato.to",
        "lang": "all",
        "id": 4531444389842992129
      },
      {
        "name": "Bato.to",
        "lang": "en",
        "id": 7890050626002177109
      },
      ...
    ]
  },
  ...
]
```

### DIFF
--- a/.github/scripts/create-repo.sh
+++ b/.github/scripts/create-repo.sh
@@ -50,6 +50,8 @@ gzip -c index.json > index.json.gz
 
 cat index.json
 
+set +e
+
 echo
 echo "executing inspector"
 echo
--- a/.github/scripts/create-repo.sh
+++ b/.github/scripts/create-repo.sh
@@ -61,5 +61,3 @@ java -jar Inspector-latest.jar ../apk inspector.txt
 cat inspector.txt
 
 rm Inspector-latest.jar
-rm *.jar
-
--- a/.github/scripts/create-repo.sh
+++ b/.github/scripts/create-repo.sh
@@ -49,3 +49,18 @@ jq -c '.' < index.json > index.min.json
 gzip -c index.json > index.json.gz
 
 cat index.json
+
+echo
+echo "executing inspector"
+echo
+
+curl "https://raw.githubusercontent.com/AriaMoradi/tachiyomi-extensions-inspector/repo/Inspector-latest.jar" -o "Inspector-latest.jar"
+
+java -jar Inspector-latest.jar ../apk inspector.txt
+
+cat inspector.txt
+
+rm Inspector-latest.jar
+rm *.zip
+rm *.jar
+
--- a/.github/scripts/create-repo.sh
+++ b/.github/scripts/create-repo.sh
@@ -61,6 +61,5 @@ java -jar Inspector-latest.jar ../apk inspector.txt
 cat inspector.txt
 
 rm Inspector-latest.jar
-rm *.zip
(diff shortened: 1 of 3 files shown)


AriaMoradi:
Any ideas, problems, suggestions for this feature?

CarlosEsco (commented):
While overall this is nice, not a fan that relies on a headless fork of tachidesk to actually work.

.github/scripts/create-repo.sh
```diff
@@ -49,3 +49,15 @@ jq -c '.' < index.json > index.min.json
 gzip -c index.json > index.json.gz
 
 cat index.json
+
+echo
+echo "executing inspector"
+echo
+
+curl "https://raw.githubusercontent.com/AriaMoradi/tachiyomi-extensions-inspector/repo/Inspector-latest.jar" -o "Inspector-latest.jar"
```
arkon, line 57: Downloading an external JAR and executing in the context of something that can change the extension themselves isn't really something I'm willing to accept since it's an inherent security risk, even if it's from someone that's a contributor to the project.

At the very least, it should be a separate action.

.github/workflows/build_push.yml
```diff
@@ -69,7 +69,6 @@ jobs:
           configuration-cache-enabled: true
 
       - name: Sign APKs
-        if: github.repository == 'tachiyomiorg/tachiyomi-extensions'
```
arkon, line 72: This stuff obviously shouldn't be part of the PR.

arkon (commented):
The PR description doesn't really state it explicitly, but this is to get the full list of sources from the extension APKs, right?

.github/workflows/build_push.yml
AriaMoradi: Yeah, I'll remove It after we decide to merge, It's necessary for now or else actions won't work on my fork.

.github/scripts/create-repo.sh
AriaMoradi: how about running it after apk files have been commited and then generate the list after that and add it in a separate commit?

arkon: Maybe if you strip it down to the bare necessities, we could transfer it over to the main organization?

AriaMoradi: Applying the same patches to both projects will be hard, unless someone decides to manually watch Tachidesk and apply relevant patches manually...

arkon:
I don't actually expect to use this data in the app anytime soon, so maybe we can just have a separate repo that runs an action on a schedule that pulls the extensions (maybe once a day) and runs the inspector to generate the data as a reference for support or something?

arkon:
Superseded by #8250

Issue #6417, opened 2021-04-04
[Bug] Webtoons: Cannot fetch new chapters 


mlilyd:
Refreshing the library does not pull new chapters. I knew that some series I follow updates on Sunday, but I didn't get any update notification from Tachiyomi. I tried manually refreshing my library for new chapters, as well as manually refreshing just that one series, but no new chapters. I checked Webtoons's website, and indeed a new chapter was uploaded, so the problem must be with Tachiyomi.

This applies for both English and Indonesian source, but I haven't checked for other languages.

mlilyd:
Ah, it seems that the entire extension is broken? I can't browse Webtoons through Tachiyomo either

curche:
possible duplicate of #6394

EstebanOD:
Same problem with English and French extension, the chapters can't be reached and the images of each WEBTOON disappeared and can't be downloaded back.

keinezeit8:
there was an update to the extension.
does it still happen?

EstebanOD:
Hello, thank you for the answer.
Fortunately no, it doesn't, I updated my library, the covers came back and
it updated the WebToons' chapters.

Issue #6394, opened 2021-04-02
[Bug] webtoons.com: nothing being loaded from extension


tiktak24:
## Device information
- Tachiyomi version: j2k fork 1.0.11
- Android version: 10 miui 12.0.3

## Source information
- Name: webtoons.com
- Extension version: 1.2.25

## Steps to reproduce
1. Browsing the source doesn't load anything

CrepeTF:
On my end at least, using a VPN works. However webview still works with or without vpn

tiktak24:
Vpning to the states worked but in UK either by VPN, data or normal WiFi it doesn't load

CrepeTF:
VPN from UK to UK works on my end

keinezeit8:
I tried it as well. Doesn't work without VPN and VPN to US. 
Tried clearing cookies.
Opening in webview allows to enter birthdate, but extension still doesn't work.

PR #6424, opened 2021-04-04
Webtoons Fix for EU users

E3FxGaming:
Should fix Webtoons not working for EU users
Reported in #6417 and #6394


E3FxGaming:
Users outside of the EU should have the `needGDPR` cookie with the value `false` anyways, so this should make no difference for them.

PR #6426, opened 2021-04-04
Fixed VMP thumbnails and chapters not loading

Neoterux:
Fixed IllegalArgumentException when saving thumbnail and when loading chapter images.


arkon:
Please update the `extVersionCode` value in the extensions' `build.gradle` files as well.

Neoterux:
updated version codes

E3FxGaming:
You're not supposed to increment the `libVersion`, that one should remain as `1.2`.

Only increment the `extVersionCode` to `1.3`

Neoterux:
oh, ok sorry for that

PR #6430, opened 2021-04-05
Fix for the multi-sources not being generated properly

E3FxGaming:
As described here https://www.infoworld.com/article/2071275/when-runtime-exec---won-t.html under "Listing 4.2 BadExecJavac2.java" the limited buffer size of the standard input associated with the sub-process can cause problems with the Runtime exec method.

Consuming what the sub-process outputs allows for the process to finish (+ if you enable log level info you get some nice debug information).

I've tested this on my local machine (Linux), on the tachiyomi-extensions GitHub runner, improved upon it and tested it again on the tachiyomi-extensions GitHub runner and it worked fine.

What I can't test is how it behaves on Windows, it should be fine but if someone could check out my https://github.com/E3FxGaming/tachiyomi-extensions/tree/FixMultisourceGen branch on Windows and run `./gradlew multisrc:generateExtensions` to see if it works, that would be great.

The `errorStream` of the sub-process gets redirected to the normal input that we receive from the sub-process, therefore we don't have to concern ourselves with the separate `errorStream`.

### DIFF
--- a/multisrc/build.gradle.kts
+++ b/multisrc/build.gradle.kts
@@ -1,6 +1,5 @@
 import java.io.BufferedReader
 import java.io.InputStreamReader
-import java.util.concurrent.TimeUnit
 
 plugins {
     id("com.android.library")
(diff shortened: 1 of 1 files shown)


arkon:
Thanks!

PR #6444, opened 2021-04-06
Webtoons split

jopejoe1:
Splits webtoons into 3 extensions
- Webtoons
- Webtoons Translate
- Dongman Manhua


jopejoe1:
Fixed Backwards compatibility

PR #6446, opened 2021-04-06
removed void scans and helvetica scans

OncePunchedMan:


multisrc/src/main/java/eu/kanade/tachiyomi/multisrc/foolslide/FoolSlideGenerator.kt
```diff
@@ -34,7 +34,6 @@ class FoolSlideGenerator : ThemeSourceGenerator {
         SingleLang("Storm in Heaven", "https://www.storm-in-heaven.net", "it", className = "StormInHeaven"),
         SingleLang("LupiTeam", "https://lupiteam.net", "it"),
         SingleLang("Zandy no Fansub", "https://zandynofansub.aishiteru.org", "en"),
-        SingleLang("Helvetica Scans", "https://helveticascans.com", "en"),
```
arkon, line 37: Remove the override files too.

arkon (changes requested):

jopejoe1:
Did i remove wrong void scan in #6421 ?

arkon:
> Did i remove wrong void scan in #6421 ?

@CrepeTF would need to confirm that since the removal issue didn't specify. Looks like this site is dead anyway though.

OncePunchedMan:
hm both voidscans and ntsvoidscans are non-reachable

CrepeTF:
> > Did i remove wrong void scan in #6421 ?
> 
> @CrepeTF would need to confirm that since the removal issue didn't specify. Looks like this site is dead anyway though.

Yep it's correct @arkon

PR #6455, opened 2021-04-07
MangasProject split

jopejoe1:
Splits MangasProject into:
- Leitor.net
- Mangá Livre
- Toonei


multisrc/src/main/java/eu/kanade/tachiyomi/multisrc/mangasproject/MangasProject.kt
```diff
@@ -1,11 +1,12 @@
...
+package eu.kanade.tachiyomi.multisrc.mangasproject
 
 import com.github.salomonbrys.kotson.array
 import com.github.salomonbrys.kotson.obj
 import com.github.salomonbrys.kotson.string
 import com.google.gson.JsonObject
 import com.google.gson.JsonParser
-import eu.kanade.tachiyomi.lib.ratelimit.RateLimitInterceptor
+//For some reason i can't import it into a multisrc
+//import eu.kanade.tachiyomi.lib.ratelimit.RateLimitInterceptor
```
jopejoe1, line 9: Some knows why I can't import this into multisrc?

OncePunchedMan: do it in the override

jopejoe1: Thanks Didn't think about that.

alessandrojean:
Just a note: all the sites from the MangasProject extension are mirrors of each other and share the same content, API, and database, and that's the reason they are in a single extension (and in general the users know about this). Is it really necessary to split them?

jopejoe1:
> Just a note: all the sites from the MangasProject extension are mirrors of each other and share the same content, API, and database, and that's the reason they are in a single extension (and in general the users know about this). Is it really necessary to split them?

Didn't know that. Gone wait for the opinion of some that's better informed than me.

CarlosEsco:
I don't see why the can't be split.  Split or together they still share the same code base.  It just makes it easier for the user to find the source

alessandrojean:
I thought it wasn't really necessary because they are mirrors of each other and the extension only has 3 sources, but if it will make it easier to find, then it's fine.

CarlosEsco:
yea the main thing is when you go into the app and search Toonei.  It will now show up.  Behind the scenes its still a multisource so should effect the actual coding portion

PR #6456, opened 2021-04-07
JapanRead: fix wrong chapter loaded

papjul:
There was a mismatch between chapter number and chapter id, resulting in wrong chapter loaded.

It's a recently added source but I'm surprised no one noticed.


CarlosEsco:
need to update the version number

papjul:
Thank you for the reminder, it is fixed

PR #6462, opened 2021-04-07
Fix issue of MangasProject not building

jopejoe1:
For some reason it refuse to import RateLimitInterceptor


alessandrojean:
Check my solution in #6463 :)

PR #6463, opened 2021-04-07
Fix wrong language in multisrc MangasProject

alessandrojean:
I fixed the wrong language code in the classes, it should be `pt-BR`.

I am just having some trouble to make the `RateLimitInterceptor` to be imported correctly, and the build is failing (locally and in the GitHub actions, commit 36af40d and the ones after it).

```
/tachiyomi-extensions/generated-src/pt/mangalivre/src/eu/kanade/tachiyomi/extension/pt/mangalivre/MangaLivre.kt: (3, 28): Unresolved reference: lib
/tachiyomi-extensions/generated-src/pt/mangalivre/src/eu/kanade/tachiyomi/extension/pt/mangalivre/MangaLivre.kt: (20, 25): Unresolved reference: RateLimitInterceptor
```

I have checked the WPMangaStream with [some sources that uses the interceptor](https://github.com/tachiyomiorg/tachiyomi-extensions/blob/defa03d68d9b7b7a25140951e73df878d7fe46dc/multisrc/overrides/wpmangastream/boosei/src/Boosei.kt#L4) in the overrides, but it's exactly the same way, so I don't know if I am missing something...


alessandrojean:
I managed to get it working, the file `additional.gradle.kts` with the import of the `lib-ratelimit` was missing.

I will just add the custom icons for each source and I will push the commit.

PR #6520, opened 2021-04-13
Update MH url (again)

alessandrojean:
I also updated the URL of the titles to avoid (for a while) the HTTP 403.

Entries in the library should still work, but users can migrate from MH to MH to update the URLs if they want too.

### DIFF
--- a/src/pt/mangahost/build.gradle
+++ b/src/pt/mangahost/build.gradle
@@ -5,7 +5,7 @@ ext {
     extName = 'Mangá Host'
     pkgNameSuffix = 'pt.mangahost'
     extClass = '.MangaHost'
-    extVersionCode = 20
+    extVersionCode = 21
     libVersion = '1.2'
 }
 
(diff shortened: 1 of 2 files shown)


jasperalbarn:
Dear,

I admire Tachiyomi's work and quality. I avoided contacting you, but as the technician responsible for the site (mangahosted.com / mangahostz.com) I have to say and explain that **Tachi simply consumes 50% of the bandwidth and 40% of the site's CPU**. Every time the APP starts to work, everything is slow. It is very complicated to stay in this Lock x Unlock war. Some users seem to want to download the site at once... Who are users of the site who use Tachi about, because eventually the Fixed IPs of these visitors are permanently blocked.

**I formally request the removal as a reading source PT from mangahost**. If it is not possible, we will continue this until someone gets tired... PS. We are the most disadvantaged side...

CarlosEsco:
We normally don't honor requests from aggregators, I will however add more severe rate limiting so there will be some less load

PR #6566, opened 2021-04-15
Attempt to fix japscan extension

### LINKED ISSUE #6234: [Bug] <Japscan>: Some chapter won't download, or open with online reading
## Device information
- Tachiyomi version: preview r3159 and latest stable release
- Android version: 11

## Source information
- Name: Japscan
- Extension version: 1.2.25

## Steps to reproduce
1. Try to download a chapter (Ex: solo leveling 144, Versatile Mage 515)
2. Try to open the same chapter with online reading

### Expected behavior
Chapter is downloaded or opened from online reading

### Actual behavior
With online reading, endless loading appears, without error
For the download, no page number is displayed and the download does not start

## Other details
For the same source, other series can be downloaded without problem, but for a few days some chapter of some series no longer works

BayLee4:
Should close #6234
It's possible to test on Solo Leveling for instance, most of 50+ chapters won't load without the patch

### DIFF
--- a/src/fr/japscan/build.gradle
+++ b/src/fr/japscan/build.gradle
@@ -5,7 +5,7 @@ ext {
     extName = 'Japscan'
     pkgNameSuffix = 'fr.japscan'
     extClass = '.Japscan'
-    extVersionCode = 25
+    extVersionCode = 26
     libVersion = '1.2'
 }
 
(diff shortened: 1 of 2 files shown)


az4521:
does this handle scrambled images? you commented out webview screenshotting but I don't see a descrambler.

BayLee4:
I guess so ? From what I can tell images are always loaded in the right order, so no need to descramble them

az4521:
scrambled like this. the webview screenshotting wasnt just thrown in for fun, it had a real purpose.  
Screenshot_20210415-171320_TachiyomiAZ

BayLee4:
Holy sh*t I wasn't aware of that issue at all, so that's what you meant by scrambled... I'll put more work in the patch, if I'm not able to bypass the scrambling of the picture I'll re-add the original fallback to the webview screenshotting. I guess that it also isn't so easy to detect if the image is scrambled or not, as checking if there's two "ujs" scripts does not seem reliable.

Thanks for the feedback, do you have any other informations I could be missing ?

BayLee4:
Do you have other titles known for having scrambled images ?

PR #6569, opened 2021-04-15
mmrcms split

### LINKED ISSUE #6466: [Meta] Split mmrcms to multisrc

jopejoe1:
To-Do:

- [x] Implement Old generator / or some other way to make it easier to add sources
- [x] add icons
- [x] test all sources
- [x] change stuff to work with #6742
- [x] Fix bug
- [-] ~~Improve Hentai Shark language selection~~ Not doing it for now maybe later

Closes #6466


jopejoe1:
Removed:
- **Mangas Yuri** (Sells Manga now)
- **frscan** (Is offline from what I see)

multisrc/src/main/java/eu/kanade/tachiyomi/multisrc/mmrcms/MMRCMSGenerator.kt
```diff
@@ -0,0 +1,122 @@
...
+class MMRCMSGenerator : ThemeSourceGenerator {
+
+    override val themePkg = "mmrcms"
+
+    override val themeClass = "MMRCMS"
+
+    override val baseVersionCode: Int = 1
+
+    override val sources = listOf(
+        SingleLang("مانجا اون لاين", "https://onma.me", "ar", className = "onma"),
```
AriaMoradi, line 15: Capitalize class name?

AriaMoradi (changes requested):

multisrc/src/main/java/eu/kanade/tachiyomi/multisrc/mmrcms/MMRCMSGenerator.kt
jopejoe1: Yeah, seems right didn't think about that.

papjul:
> Removed:
> 
>     * **[frscan](https://www.frscan.me/)** (Is offline from what I see)

I don't know what the old website looked like, but I believe it's frscan.cc now?

jopejoe1:
> > Removed:
> > ```
> > * **[frscan](https://www.frscan.me/)** (Is offline from what I see)
> > ```
> 
> I don't know what the old website looked like, but I believe it's frscan.cc now?

Added frscan.cc

jopejoe1:
Should be ready for merge

CarlosEsco:
Unmark it draft?  How does this source work now?  Run MMRCMSJsonGen and thats it?

jopejoe1:
Found a major bug that i need to fix first thats why i put it back into draft.

jopejoe1:
Also need to change stuff to work with #6742

jopejoe1:
> Unmark it draft?

See comments above

> How does this source work now? Run MMRCMSJsonGen and thats it?

That was the intended way, not sure if its still gone be like that after i fixe the bug.

PR #6748, opened 2021-04-28
Remove JsonParser deprecation warning in some sources

alessandrojean:
It shouldn't change anything to the users, so I don't bumped the `extVersionCode` of the extensions.

If it's better to bump them, just let me know.


arkon:
I'd actually like to eventually remove usages of Gson in favour of kotlinx-serialization, which the app uses. That's out of scope of this PR though, obviously.

alessandrojean:
> I'd actually like to eventually remove usages of Gson in favour of kotlinx-serialization, which the app uses. That's out of scope of this PR though, obviously.

That's nice. When the `kotlinx-serialization-json` gets added to this repo I can help to update some extensions :thumbsup:.

PR #6764, opened 2021-04-29
Fix Tapas and Japanread

nzoba:
I fixed a few things in `Japanread` (missing chapters when multiple pages, wrong `popularManga` url).

I also added a preference in settings for `Tapas` to show the lock for locked chapters even after login (I think it's quite useful to know whether it's already unlock or not). And I changed the default header to display the mobile version in WebView instead of the desktop one.


papjul:
Thank you for the missing chapters fix, however regarding the popular manga, you replaced "most viewed" (Les + vus) with "Liste des mangas" (sorted by most recent update), which is redundant with the existing "Most recent" filter. I asked to revert this change ;)

PR #6766, opened 2021-04-30
Japanread: fix popular

papjul:
In b13ccc27ecb301416e3b54949b4033da00297c1c "popular" filter was replaced by "most recent".

Reverting to original working version.

### DIFF
--- a/src/fr/japanread/build.gradle
+++ b/src/fr/japanread/build.gradle
@@ -5,7 +5,7 @@ ext {
     extName = 'Japanread'
     pkgNameSuffix = 'fr.japanread'
     extClass = '.Japanread'
-    extVersionCode = 5
+    extVersionCode = 6
     libVersion = '1.2'
     containsNsfw = true
 }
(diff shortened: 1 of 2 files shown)


nzoba:
Yeah I forgot to add a parameter in the url. It's better to use the manga-list where you can add `sortType=9` for "Les + vus" or `sortType=7` for "Les mieux notés" and you will have the full list, not only the first ones. Example : https://www.japanread.cc/manga-list?sortType=9

papjul:
Wow, I didn't know about this parameter, where did you find this? :o

You're right, it's better to use this page since it can be paginated ;)
I will use add the parameter on top of your fix instead then ;)

nzoba:
This filter appeared in the url when I changed the released date (Année de sortie) in the filters :)
If you change the number you can have access to even more sortType so it could be possible to implement a filterlist to allow the user to change the sorting order himself

papjul:
Ready for review: I also made some refactoring so all calls use the same logic behind ;)

@nzoba I don't know exactly how filters works and I don't have much time ahead, but I can see quite some nice filters to make indeed x)

nzoba:
No worries, I'll look into it when I find the time :)

Issue #6790, opened 2021-05-02
[Feature Request] <Bato.to>: <arbitrary chapter count filtering >


Jaykoerner:
Currently the extension has options matching what is selectable on the site, but all the selections do is pass hardcoded parameters via a query in the URL, as such fi!get options can be made to allow arbitrary input.

---

** What feature should be added to this source?**
The current chapter filter options should be replaced with two text boxes, one for min, and one for max chapter count.

**Explain why this should be added**
Allows better filtering of the source

**Design**

- One header to the following, that has the text, Chapters
- Two text boxes preferable limited to integer only input, given the labels Min and Max(or some variant of that) 

**Description of api**

- The query has the general Format of chapters=#-#
- First # sets minimum chapter count 
- Second # sets maximum chapter Count
- Minimum chapter count is not required
- Maximum chapter count is not required
- "-" is only required when setting maximum chapter count 

 Examples

Request titles with chapter count above 99

- https://bato.to/latest?chapters=99 

Request titles with chapter count below 99

- https://bato.to/latest?chapters=-99 
or
- https://bato.to/latest?chapters=0-99 (you may pass a 0 if desired but it is not required)

Request titles with a chapter count between 10 and 99

- https://bato.to/latest?chapters=10-99

  

**Function**
Use of the Text input boxes will call the url query Chapters=(output of min)(-(output of max))
If no input is given the corresponding output of the query should be blank,

Jaykoerner:
That does not close this request, that was a reference to show that arbitrary filtering works, i can not do the coding and just added skme no exisen options

Jaykoerner:
Sorry. Thought that was the one i did, my bad

PR #6792, opened 2021-05-02
Added some chapter filtering options to Bato.to

Jaykoerner:
added a few more options then what is available on the site


Jaykoerner:
This is a simple edit as a alternative to implementing feature request #6790 ,I'm trying my hand at it but without luck because of lack of ability, i added this for the time being, feel free to change or add to the options submitted

Issue #3475, opened 2020-06-10
[Meta] Source requests and known issues


SnakeDoc83:
#### This list is not exhaustive and there is no timetable on when or if a source request may be filled (also applies to bug reports and enhancement requests).

**General**
- In general heavily paywalled sites
- source's that require cracking some app's source code (realistically, no one's going to put in the effort to do it)

**Too Difficult to maintain**
- fanfox.net (MangaFox) #988 
- mangago.me #988
- SuperMangás and SuperHentais
- Union Mangás https://github.com/tachiyomiorg/tachiyomi-extensions/pull/7065
- Mangá Host https://github.com/tachiyomiorg/tachiyomi-extensions/pull/7065
- HQ Dragon https://github.com/tachiyomiorg/tachiyomi-extensions/pull/7065
- YES Mangás https://github.com/tachiyomiorg/tachiyomi-extensions/pull/7065

**Requested removal by Scanlator team**
- TeamX/tqneplus  #4877
- JuinJutsuReader
- Manga disk
- KomikTap.in
- Gourment Scans
- Manga Crimson/MangaWOW: #591
- voidscans #6418 
- Hikari Scans #6611
- Mangawow
- Mangagecesi
- PiedPiperfb #6927
- KomikTap #6133

-------------

### Known issues

**Cloudflare**:  Some sources have trouble bypassing Cloudflare; it can be an intermittent or ongoing problem. Often when Cloudflare isn't bypassed automatically it can be bypassed by simply tapping "retry." In other cases it must be bypassed by opening the source in webview and letting Cloudflare fully load. As you wait you'll either see the webview forward to your source's website or you'll be presented with a captcha to solve. If you get a captcha, solve it and then you should be forwarded to the source's website. At that point you can close the webview and try to use the source as you would normally.

**MangaRaw Handshake Failed** (manga1000, manga1001): Should be fixed in the newest releases of Tachiyomi-stable (#3233). 

**Can't browse Korean sources (JMana, NewToki)**: Some Korean sources use numbered domains (e.g. newtoki55.com) that change every 7-10 days. Use extension settings to change that source's domain to the current number.

**MangaHasu 400 Bad Request**: Clear your cookies.

**NineManga won't load / gardening site in webview**: Clear your cookies.

[...]

PR #6827, opened 2021-05-05
Update MangaHost.kt

gbtx:
Update base_url


alessandrojean:
The site owner doesn't want the extension working anymore, as described in #6520, and that's the reason why the extension doesn't get updates and included a blocking message, because it's a cat and mouse game (if the extension is fixed, they will change something in the site to break it again, it's a endless loop and not worthy to maintain). It's also included in #3475 as "Too difficult to mantain".

It's better to drop the support instead of insisting in something that will always be endless for now on. I am waiting some time to the users update and read the blocking message and migrate their entries in the library to another sources to submit a PR to remove the extension.

PR #6837, opened 2021-05-05
remove duplicate slash

### LINKED ISSUE #6803: [Bug] Manhuadui: some new chapter load failed
## Device information
- Tachiyomi version: 0.10.12
- Android version: 11

## Source information
- Name: Manhuadui
- Extension version: 1.2.14

## Steps to reproduce
1. open https://www.ykmh.com/manhua/cuimianmaikefengDivisionRapBattlesideFPM/ in tachiyomi
2. open "Ⅱ-4 hook-4" chapter

### Expected behavior
chapter load successfully

### Actual behavior
failed to load chapter

## Other details
not only this manga, a lot new chapter in this site(Manhuadui) have this problem, manga cover picture also affected. if i open this chapter in inbuilt webview, all of the picture will failed to load, but if i open this page with other browser (i tried firefox and via), all of pictures will load successfully.

rhjdvsgsgks:
close #6803

### DIFF
--- a/core/build.gradle.kts
+++ b/core/build.gradle.kts
@@ -4,6 +4,7 @@ plugins {
 
 android {
     compileSdkVersion(Config.compileSdk)
+    buildToolsVersion(Config.buildTools)
 
     defaultConfig {
         minSdkVersion(Config.minSdk)
--- a/src/zh/manhuadui/build.gradle
+++ b/src/zh/manhuadui/build.gradle
@@ -5,7 +5,7 @@ ext {
     extName = 'Manhuadui'
     pkgNameSuffix = 'zh.manhuadui'
     extClass = '.Manhuadui'
-    extVersionCode = 14
+    extVersionCode = 15
     libVersion = '1.2'
 }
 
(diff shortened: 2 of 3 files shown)


core/build.gradle.kts
```diff
@@ -4,6 +4,7 @@ plugins {
 
 android {
     compileSdkVersion(Config.compileSdk)
+    buildToolsVersion(Config.buildTools)
```
CarlosEsco, line 7: was this change needed?

rhjdvsgsgks: without this change, gradle will try to use default build tool version (29.0.3), and will fail to compile if only other version of build tool installed.

CarlosEsco: @arkon   thoughts?
It doesn't even match main app
https://github.com/tachiyomiorg/tachiyomi-extensions/blob/50bf2a56e5a72e70d57ba622055a59a29aab680c/buildSrc/src/main/kotlin/Config.kt#L5

https://github.com/tachiyomiorg/tachiyomi/blob/master/buildSrc/src/main/kotlin/AndroidConfig.kt#L5

arkon: We can merge this and update it. It doesn't really matter if they differ either.

PR #6841, opened 2021-05-05
Add host preference madara

### LINKED ISSUE #6835: [Feature Request] Madara: Server/Host selecter
The Madara theme has a built-in server/host switcher.
Would be nice if madara based extensions are able to chose which host  images are going to be provided from. Because sometimes the default host is providing broken or missing images.
They can be accessed by the URL parameter `?host=<hostname>` on the page list of a chapter.

Supported Hosts that I'm aware of:

- local
- amazon
- imgur
- flickr
- picasa (Blogspot)
- gphotos (Google Photos)

nzoba:
I added a preference to choose the host server for Madara sources. If the preference is set to *Auto*, it will try the *default* then each server until it finds some pages (if possible).

There are probably some things to improve so don't hesitate to suggest if you have any idea.

Closes #6835

I also moved `Method Scans` to `GenkanOriginal`

### DIFF
--- a/multisrc/src/main/java/eu/kanade/tachiyomi/multisrc/genkan/GenkanGenerator.kt
+++ b/multisrc/src/main/java/eu/kanade/tachiyomi/multisrc/genkan/GenkanGenerator.kt
@@ -17,7 +17,6 @@ class GenkanGenerator : ThemeSourceGenerator {
         SingleLang("ZeroScans", "https://zeroscans.com", "en"),
         SingleLang("The Nonames Scans", "https://the-nonames.com", "en"),
         SingleLang("Edelgarde Scans", "https://edelgardescans.com", "en"),
-        SingleLang("Method Scans", "https://methodscans.com", "en"),
         SingleLang("LynxScans", "https://lynxscans.com", "en", overrideVersionCode = 1),
     )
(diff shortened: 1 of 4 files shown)


jopejoe1:
Probably would be best if it tries the other hosts to after failing the selected host, as some chapters maybe have different host servers than others.

nzoba:
Well I thought the user would choose *Auto* if there are many servers for his source. But it doesn't cost anything to allow to try others after failing when you choose a particular host so I'll add it :)

nzoba:
@jopejoe1 Do you have examples of sources/chapters that use multiple hosts to verify if everything works with a real case please ?

I decided to implement something much simpler : no preference, it will always automatically try to find a working host. When found, it will become the default one.

jopejoe1:
> I decided to implement something much simpler : no preference, it will always automatically try to find a working host. When found, it will become the default one.

It could cause problems in cases where one of the host has partly broken images like here, or different amount of images:
https://disasterscans.com/manga/martial-peak/chapter-614 (Amazon, Imgur)(default Amazon)
https://disasterscans.com/manga/martial-peak/chapter-635 (Imgur, Local)(default Imgur)

Issue #6862, opened 2021-05-06
[Bug] Mangadex: Not a JSON Object: null


Vverg:
---

## Device information
- Tachiyomi version: Preview r3346 
- Android version: 11

## Source information
- Name: Mangadex
- Extension version: 107

## Steps to reproduce
1. Pick a random manga, pick something with enough chapters.
2. Download all chapters.
3. Notice that eventually the download stops, sometimes with the error 'Not a JSON Object: null'.
4. Pause and continue the download queue doesnt solve the issue, I have the remove everything and download again.

### Expected behavior
That it downloads chapters without issues.

### Actual behavior
Stops downloading, make the queue get stuck where I can only fix the issue by removing everything in the queue.
Also getting a error sometimes as mentioned above.

CarlosEsco:
Can you provide a sample manga?

Vverg:
One example would be: Gokukoku no Brynhildr

Though the chapter skipping (the red explanation mark) and getting stuck does happen with every title I download. If I download a title my group does "Bijutsubu Girl", that one has mostly 1 page chapters you will notice the chapter skipping faster.

Riztard:
does it happen to specific chapter or random?

Vverg:
It seems randomly, since I can download the chapter eventually after I delete the queue and readd it. 
However the error 'Not a JSON Object: Null' i only noticed so far with this title.

However, the queue getting stuck seems to happen with every issue that causes a chapter not to be downloaded. Since I also get this issue with random 502/504 error codes. So that might be a bug of Tachiyomi itself instead of the extension.

shadowvalentine:
I've had this problem as well, with "Mushoku no Eiyuu" chapter 14, and "12 Beasts" chapter 20 and 23. Removing those chapters and re-adding it fixed the problem for the "12 Beasts" chapters, but not the "Mushoku no Eiyuu" chapter; it now appears to be completely skipping it (I've tried moving it to top, and pausing and re-starting the download).  When my downloads all complete, I'll attempt it again alone. 

Trying to open the chapter directly shows that it has 25 pages (about right for this manga) but each page cannot load; could it be malformed somehow on Mangadex's end?

shadowvalentine:
Hi, still having this problem on Mangadex exxtension ver 108 with the same manga and chapter, Mushoku no Eiyuu chapter 14. Same error, "Not a JSON Object: null".

CarlosEsco:
[...]

Issue #6860, opened 2021-05-06
[Enhancement] Mangadex: Hide Content Types in Preferences


Jetspectre:
##  Why/Project's Benefit/Existing Problem
Currently, there's no option to hide 18+ content and they are on the front page. Having the ability to hide this content under the extension settings would make it easier for the people to search without having to look at R18 manga.

PR #6863, opened 2021-05-06
MangaDex: remove empty tag, fix genre typo, revert pt-br changes

Riztard:
- remove empty tag caused by null/empty publicationDemographic/other from nonGenres if its null/empty
- fix typo genre uuid
- revert pt-br locale changes


CarlosEsco:
Can you add this 
https://github.com/tachiyomiorg/tachiyomi-extensions/issues/6860
i was thinking 4 checkbox preferences 
then default checking them on getFilterList

CarlosEsco:
also look into this https://github.com/tachiyomiorg/tachiyomi-extensions/issues/6862

Riztard:
> Can you add this

my brain is too smoll for that ;P :tehe:

CarlosEsco:
> 
> 
> > Can you add this
> 
> my brain is too smoll for that ;P :tehe:

probably something like this  (delete https://github.com/tachiyomiorg/tachiyomi-extensions/blob/master/src/all/mangadex/src/eu/kanade/tachiyomi/extension/all/mangadex/MangaDexFilters.kt#L48)
```
internal fun getMDFilterList(preference: PreferenceHelper, lang : String) : FilterList {
val contentRatings = listOf(
 ContentRating("Safe").apply{ state = preference.get("${MdConstants.safe}_$lang")},
        ContentRating("Suggestive").apply{ state = preference.get("${MdConstants.suggestive}_$lang")},
        ContentRating("Erotica").apply{ state = preference.get("${MdConstants.erotica}_$lang")},
        ContentRating("Pornographic").apply{ state = preference.get("${MdConstants.pornographic}_$lang")}, 
#please name these prefs better then just the rating lol

return FilterList(
        OriginalLanguageList(getOriginalLanguage()),
        ContentRatingList(contentRatings ),
        DemographicList(getDemographics()),
        StatusList(getStatus()),
        SortFilter(sortableList.map { it.first }.toTypedArray()),
        TagList(getTags()),
        TagInclusionMode(),
        TagExclusionMode(),
    )
```

src/all/mangadex/src/eu/kanade/tachiyomi/extension/all/mangadex/MangaDexHelper.kt
```diff
@@ -187,7 +187,7 @@ class MangaDexHelper() {
                 author = authors.joinToString(", ")
                 status = getPublicationStatus(attr["publicationDemographic"].nullString)
                 thumbnail_url = MDConstants.tempCover
-                genre = genreList.joinToString(", ")
+                genre = genreList.joinToString(", ").replace(", , ", ", ")
```
CarlosEsco, line 190: it makes more sense to edit this
https://github.com/tachiyomiorg/tachiyomi-extensions/blob/a725b320dd37675fc6aeffadb8525f75b97f6925/src/all/mangadex/src/eu/kanade/tachiyomi/extension/all/mangadex/MangaDexHelper.kt
and instead of filterNotNull  .filter{it !=null && it.isNotEmpty()}

Riztard: ah right we can use blank thinggy
it has to be "&&" (and)? not "||" (or)? not sure how exactly filter work

CarlosEsco: filter keeps the true result
in this case we want the string to be not blank and not null

that last works one works!

Riztard: > that last works one works!

this? `.filter {it.isNullOrBlank().not()}`
im confused with double works lol

CarlosEsco: yes that works.

I forgot there was isNullOrBlank method

CarlosEsco:
I'll finish out the rest of the changes later today

PR #6954, opened 2021-05-10
[Mangadex]fix crossdressing tag id

INSANEMODE:
the crossdressing id was using the ninja tag id "489dd859-9b61-4c37-af75-5b18e88daafc" when it should be "9ab53f92-3eed-4e9b-903a-917c86035ee3", as seen in https://api.mangadex.org/manga/tag


CarlosEsco:
@INSANEMODE  Did you verify the other tags also?

CarlosEsco:
If you don't mind double checking that would be great.  Also you will need to bump the build.gradle version

paulo27ms:
I checked the differences: https://www.diffchecker.com/IgiTI6gS
Mangadex vs Tachiyomi
Note Tachiyomi has Ecchi which MD doesn't and it's using Gyaru's code.
The others are spellings differences.

CarlosEsco:
Ok if the spelling changes can be matched and ecchi removed.  (this is not a content rating)

INSANEMODE:
alright, everything should be correct now.

PR #7003, opened 2021-05-13
Home Hero Scans: Add support for deeplinking + fix bug where webview wouldn't launch

h-hyuuga:
This pull request:

* Fixes a bug where it was impossible launch webview for manga
* Enables deeplinking to manga

### DIFF
--- a/src/en/homeheroscans/src/eu/kanade/tachiyomi/extension/en/homeheroscans/HomeHeroScans.kt
+++ b/src/en/homeheroscans/src/eu/kanade/tachiyomi/extension/en/homeheroscans/HomeHeroScans.kt
@@ -1,6 +1,5 @@
 package eu.kanade.tachiyomi.extension.en.homeheroscans
 
-import android.util.Log
 import com.github.salomonbrys.kotson.forEach
 import com.google.gson.JsonParser
 import eu.kanade.tachiyomi.network.GET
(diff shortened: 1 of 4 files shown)


src/en/homeheroscans/src/eu/kanade/tachiyomi/extension/en/homeheroscans/HomeHeroScans.kt
```diff
@@ -28,6 +27,23 @@ open class HomeHeroScans : HttpSource() {
...
+        fun decorated(arg: P) = cache[arg]?.let { Observable.from(it) } ?: f(arg).map {
+            cache.getOrPut(arg, ::mutableListOf).add(it)
+            it
+        }
+        return ::decorated
+    }
+
+    val memoizedFetchPopularManga = memoizeObservable { page: Int -> super.fetchPopularManga(page) }
+    // reduce number of times we call their api, user can force a call to api by relaunching the app
+    override fun fetchPopularManga(page: Int) = memoizedFetchPopularManga(page)
```
CarlosEsco, line 46: Why are you memoizing?  Just curious

h-hyuuga: I don't expect the results of invoking fetchPopularManga to change over  the lifetime of source after it gets loaded into Tachiyomi.

Pretty much everything I'm doing in that extension calls fetchPopularManga, so I figured why not just cache the results and re-use them. 

The extension ends up feeling faster, and I'm not needlessly invoking their api.

h-hyuuga: Did you notice anything potentially problematic with this behavior, or is it okay?

CarlosEsco: It's good just didn't have time to merge

PR #7006, opened 2021-05-13
Add a warning when the manga URL changes in NS

alessandrojean:
Reported by some users that sometimes the source changes the slug of the manga (I don't know if this is a measure to prevent 3rd party apps from accessing their content). Now, when this happens again, there will be a warning to inform the user to migrate the manga from the source to the same source to update the URL.

I also filtered the novels in pure text format, that the app does not support.


alessandrojean:
Hmm, I don't know why the build failed. I will take a look later, marking as draft for now.

PR #7066, opened 2021-05-17
Use correct field name for MD@H report endpoint

Korbeil:
Today, all report requests are denied by MD@H backend since the field doesn't match the required schema:

```Go
type reportRequest struct {
    URL      string  `json:"url"`
    Success  bool    `json:"success"`
    Bytes    int     `json:"bytes"`
    Duration float64 `json:"duration"` // milliseconds, can have decimal places
    Cached   bool    `json:"cached"`
}
```


CarlosEsco:
need to bump the build.gradle file version

CarlosEsco:
also that damn "d" smh

PR #7091, opened 2021-05-19
Added Asura Scans back

loocool2:
Might be buggy. Added back due to hosting other groups scanlations that Asura had no contributions to.


Riztard:
where is the generator

jopejoe1:
You probably want to remove it from the issue closer too

PR #7096, opened 2021-05-19
Add Phoenix Fansub extension

SuperP91:


src/es/phoenixfansub/src/eu/kanade/tachiyomi/extension/es/phoenixfansub/PhoenixFansub.kt
```diff
@@ -0,0 +1,238 @@
...
+
+    override fun popularMangaSelector() = ".listupd .bs"
+
+    override fun popularMangaFromElement(element: Element): SManga = SManga.create().apply {
+        setUrlWithoutDomain(element.select("a").attr("href"))
+        thumbnail_url = element.select("img").attr("src")
+        title = element.select(".bigor .tt").text().trim()
+    }
+
+    override fun popularMangaParse(response: Response): MangasPage {
```
alessandrojean, line 49: Tip: The `popularMangaParse` function is exactly the same as the one in `ParsedHttpSource`, you don't need to override it if you don't modify anything.

https://github.com/tachiyomiorg/tachiyomi/blob/0dbb79359bf01aa93305efedf4ef80c1e0c0ecb8/app/src/main/java/eu/kanade/tachiyomi/source/online/ParsedHttpSource.kt#L22-L34

src/es/phoenixfansub/src/eu/kanade/tachiyomi/extension/es/phoenixfansub/PhoenixFansub.kt
```diff
@@ -0,0 +1,238 @@
...
+    }
+
+    override fun searchMangaNextPageSelector(): String? = popularMangaNextPageSelector()
+
+    override fun searchMangaSelector(): String = popularMangaSelector()
+
+    override fun searchMangaFromElement(element: Element): SManga = popularMangaFromElement(element)
+
+    override fun searchMangaParse(response: Response): MangasPage {
+        if (!response.isSuccessful) throw Exception("Búsqueda fallida ${response.code}")
```
alessandrojean, line 114: If the response was not successful this method will not even get called, because of the `asObservableSuccess` in `fetchSearchManga`. The rest of the function have the exact same content from `searchMangaParse` of `ParsedHttpSource`, don't need to override too.

src/es/phoenixfansub/src/eu/kanade/tachiyomi/extension/es/phoenixfansub/PhoenixFansub.kt
```diff
@@ -0,0 +1,238 @@
...
+    override fun chapterFromElement(element: Element): SChapter = SChapter.create().apply {
+        name = element.select(".chapternum").text().trim()
+        setUrlWithoutDomain(element.select("a").attr("href"))
+        chapter_number = element.attr("data-num").toFloat()
+        date_upload = parseDate(element.select(".chapterdate").text())
+    }
+
+    private fun parseDate(date: String): Long {
+        val spanishLocale = Locale("es", "ES")
+        return SimpleDateFormat("MMMM d, yyyy", spanishLocale).parse(date)?.time ?: 0
```
alessandrojean, line 153: Nitpick: it's better to create the date parser as a constant in a `companion object`, so you don't need to instantiate it every time when parsing a chapter. Check [this](https://github.com/tachiyomiorg/tachiyomi-extensions/blob/9936731198a9ebb8719e652228cdf9541a295133/src/pt/mangatube/src/eu/kanade/tachiyomi/extension/pt/mangatube/MangaTube.kt#L286) as a reference.

src/es/phoenixfansub/src/eu/kanade/tachiyomi/extension/es/phoenixfansub/PhoenixFansub.kt
```diff
@@ -0,0 +1,238 @@
...
+    }
+
+    override fun pageListParse(document: Document): List<Page> {
+        var images = document.select("script")[25].html().split(".run(")[1]
+        images = images.subSequence(0, images.length - 2) as String
+        val json = JsonParser.parseString(images)
+        val imageArray = json["sources"][0]["images"]
```
alessandrojean, line 160: If this is an array of strings, you can replace this by:

```kotlin
return json["sources"][0]["images"].array
    .mapIndexed { i, img -> Page(i, "", img.string) }
```

src/es/phoenixfansub/build.gradle
```diff
@@ -0,0 +1,17 @@
...
+ext {
+    extName = 'Phoenix Fansub'
+    pkgNameSuffix = 'es.phoenixfansub'
+    extClass = '.PhoenixFansub'
+    extVersionCode = 1
+    libVersion = '1.2'
+    containsNsfw = false
+}
+
+dependencies {
```
alessandrojean, line 13: Can remove this dependencies section since you are not using rate limiting in the OkHttp client.

alessandrojean:
I have not checked when reviewing the code, but taking a look at the site now, it appears that it uses WPMangaStream, so you can probably just add this source to the `multisrc` generator.

SuperP91:
You were absolutely right! The site uses WPMangaStream. I changed the extension to use the multisrc generator. 

Thank you!

as280093:
Hey you missed 512 sized icon also there is less spacing so you can use this 
[PheonixFansubs.zip](https://github.com/tachiyomiorg/tachiyomi-extensions/files/6513433/PheonixFansubs.zip)

PR #7115, opened 2021-05-20
[SKIP CI] Added caseIgnore to relevant tasks

loocool2:
Another detail I missed


alessandrojean:
I think the action expects the parameter to be `ignoreCase` instead of `caseIgnore`.

arkon:
@loocool2 pls

PR #7160, opened 2021-05-21
[Mangadex] Show Manga Cover from seperate API + UUID in description.

FourTOne5:
Noticed the coverapi on Neko and now it's here.

Since everything redirects to mangadex.org this will help to see the Manga UUID until the frontend of the site is up.


CarlosEsco:
What benefit does showing the uuid provide?

PR #7177, opened 2021-05-22
Mangadex AltCover

FourTOne5:
It will try to get cover from related link like MAL, KU, Al etc. and if it fails then back to the original tempcover.


arkon:
Sorry, but this is redundant given MD is going to add covers soon (TM).

FourTOne5:
Well yes soon™ but as the time when soon™ will come isn't specified shouldn't this approach be taken. I mean people will like the actual cover than cover isn't available won't they?

arkon:
It was literally announced today.

FourTOne5:
ok.....

PR #7196, opened 2021-05-23
Added PizzaReader multisrc CMS

FedericoHeichou:
I added the multisrc for PizzaReader CMS.  

There is a comment in the Generator with the GTO's extension.  
When the website's CMS will be fully migrated I will move the current GTO FoolSlide's extension into PizzaReader and I will uncomment the line in the generator.

In this way I should update the extension without problems


FedericoHeichou:
Sadly seems it doesn't compile without sources in generator, I will add commits when everything is done

CarlosEsco:
There is a compile error for this PR

alessandrojean:
If you will still work on this, please replace the `org.json` usage with `kotlinx.serialization` as per #7392.

FedericoHeichou:
> If you will still work on this, please replace the `org.json` usage with `kotlinx.serialization` as per #7392.

Yes, I will do it

FedericoHeichou:
Hello, I migrated a valid source to this CMS, can you merge it if it is ok?

I need to add the icons of this scanlation team, moreover I need to replace `org.json` usage with `kotlinx.serialization`.
I hope I can do this in the next weeks.
Anyway I'm not a Android developer, so some help to replace `org.json` with `kotlin.serialization` would be really appreciated.
I looked some examples but seems hard to convert it and I have no much time.

I hope someone skilled can help me, but for the moment it's ok deploy this version, I will raise `baseVersionCode` to 1 when I will migrate to `kotlin.serialization`. Obviously if you want you can commits directly to replace `org.json`, but it would be a dream :smile:

alessandrojean:
No problem, I think you can get it merged on how it is for now and later on submit a new PR to convert the deserialization part to kotlinx.serialization.

multisrc/overrides/pizzareader/phoenixscans/src/PhoenixScans.kt
```diff
@@ -0,0 +1,5 @@
+package eu.kanade.tachiyomi.extension.it.phoenixscans
+
+import eu.kanade.tachiyomi.multisrc.pizzareader.PizzaReader
+
+class PhoenixScans : PizzaReader("Phoenix Scans", "https://www.phoenixscans.com", "it")
```
alessandrojean, line 5: Did they changed the URL structure in their new website?

If so, it's better to bump the `versionId` to force users to migrate to update the URLs from their library entries.

You can refer to the code below, just add it inside the PhoenixScans class

```kotlin
override val versionId = 2
```

FedericoHeichou: Yes, it's changed
The current PhoenixScans version in FoolSlide is 1.2.3, I set ovveride to 4 inside my CMS. It is not sufficient? I should add this `versionId = 2` too?

alessandrojean: That doesn't have relation with this.

I asked if the URL structure of the manga pages changed, for example: when they used Foolslide the URL was website.com/manga/title and now it's website.com/translation/title.

If it's the case, yes, it's better to bump the `versionId` inside the PhoenixScans class to force users to migrate and update the URLs of the manga in their libraries.

FedericoHeichou: Yes, sure, now I understand. Thank you very much, I applied the `versionId`

PR #7252, opened 2021-05-27
Attempt to bypass Cloudflare at NS

alessandrojean:
This source is very slow nowadays, and they are always changing the manga URLs, probably to block the app.

This PR fixes the WebView method of bypassing Cloudflare and also adds a more harsh rate limiting.

The fix probably will stop working soon, if it's really a cat and mouse game.


CarlosEsco:
Wouldn't be a pt-BR source unless they were trying to block tachi

PR #7271, opened 2021-05-28
Fix missing slash in api endpoint

kafkaien42:
Endpoints of both guya.moe and danke.moe, ends with a slash. This saved a http request.


CarlosEsco:
You need to update the build.gradle version

kafkaien42:
> You need to update the build.gradle version

Which build.gradle do I need to update?

arkon:
https://github.com/tachiyomiorg/tachiyomi-extensions/blob/0f7d576c60f2a0d37ab985fd335f4b97646b938c/multisrc/src/main/java/eu/kanade/tachiyomi/multisrc/guya/GuyaGenerator.kt#L12

PR #7343, opened 2021-05-31
[ZH-Dmzj] Move some logic to use dmzj's new v4api, fix #6858

### LINKED ISSUE #6858: [Bug] DMZJ: Marked Unnumbered Chapters as Unread Repeatly
## Device information
- Tachiyomi version: TachiJ2K 1.0.16
- Android version: 10

## Source information
- Name: DMZJ
- Extension version: 1.2.16

## Steps to reproduce
1. Adding manga which using specific title but not number.
2. Mark some chapters as read.
3. Update the library.

### Expected behavior
Chapters which already read still marked as read.

### Actual behavior
Some of chapters re-marked as unread.

## Other details
Some of manga chapters in DMZJ would named with a specific name, and does not have a certain chapter number. This chapter would have chapter number `-1` when you share the manga. For example:  `Helltaker推特短篇集: 第 -1 章, 第 4 页`.

This kind of chapters would re-unread randomly after update the library. Some chapters like `15.5` `12.1` would may re-marked as unread too. It's hard to reproduce and hard to figure which would be affected and which would not in next time library update.

I am only using TachiJ2K, so I can't confirm if it appears in upstream version. And since it only appears in DMZJ, so I pull it as a extension problem.

Oldwangtouchtouchdoge:
Move some logic to use dmzj's new v4api, Closes #6858.

In the dmzj extension version 1.2.16 or before, chapter list will be fetch from multiple api, so sometimes chapter urls in subscribed manga may be change by updating library. At the same time, chapter number were extract the first string of numbers from chapter title, so when title don't contain any number or extracted chapter number having any duplicate, tachiyomi can't recognize these chapters and mark them as unread. To solve the issue, new version of code will store chapter's url as unchange comic ID and chapter ID. When fetching the chapter list, url will be concat to the actual url depend on which api is using.

Protobuf relative files, decrypt private key and v4api usages are obtain from https://github.com/xiaoyaocz/dmzj_flutter project. All credit goes to their outstanding work. 👍


alessandrojean:
You can probably use `kotlinx.serialization` for Protobuf stuff, like it's used in Manga Plus extension.

https://github.com/tachiyomiorg/tachiyomi-extensions/tree/master/src/all/mangaplus/src/eu/kanade/tachiyomi/extension/all/mangaplus

Oldwangtouchtouchdoge:
> 
> 
> You can probably use `kotlinx.serialization` for Protobuf stuff, like it's used in Manga Plus extension.
> 
> https://github.com/tachiyomiorg/tachiyomi-extensions/tree/master/src/all/mangaplus/src/eu/kanade/tachiyomi/extension/all/mangaplus

Thanks for advice, I will check it out.

PR #7369, opened 2021-06-01
[ZH-dmzj] Move some logic to use dmzj's new v4api, fix #6858

### LINKED ISSUE #6858: [Bug] DMZJ: Marked Unnumbered Chapters as Unread Repeatly

Oldwangtouchtouchdoge:
Almost same as PR #7343, I made some mistake with git and closed that PR. 🤣
Protobuf deserialization library has changed to kotlinx.serialization now.

### DIFF
--- a/src/zh/dmzj/build.gradle
+++ b/src/zh/dmzj/build.gradle
@@ -13,16 +13,6 @@ ext {
 dependencies {
     implementation project(':lib-ratelimit')
     implementation 'org.jetbrains.kotlinx:kotlinx-serialization-protobuf:1.2.0'
-    coreLibraryDesugaring 'com.android.tools:desugar_jdk_libs:1.1.5'
-}
-
-android {
-    defaultConfig {
-        multiDexEnabled true
-    }
-    compileOptions {
-        coreLibraryDesugaringEnabled true
-    }
 }
 
 apply from: "$rootDir/common.gradle"
--- a/src/zh/dmzj/build.gradle
+++ b/src/zh/dmzj/build.gradle
@@ -1,16 +1,28 @@
 apply plugin: 'com.android.application'
 apply plugin: 'kotlin-android'
+apply plugin: 'kotlinx-serialization'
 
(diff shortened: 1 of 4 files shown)


src/zh/dmzj/build.gradle
```diff
@@ -1,16 +1,28 @@
...
     extClass = '.Dmzj'
-    extVersionCode = 16
+    extVersionCode = 17
     libVersion = '1.2'
 }
 
 dependencies {
     implementation project(':lib-ratelimit')
+    implementation 'org.jetbrains.kotlinx:kotlinx-serialization-protobuf:1.2.0'
+    coreLibraryDesugaring 'com.android.tools:desugar_jdk_libs:1.1.5'
```
alessandrojean, line 16: Is this desugaring stuff still used?

Oldwangtouchtouchdoge: One usage to get epoch second, I changed to other approach.

alessandrojean (approved):

Issue #7392, opened 2021-06-02
Replace Gson and org.json with kotlinx.serialization


alessandrojean:
##  Why/Project's Benefit/Existing Problem

There are in the repo actually:
- ~~[1 file](https://github.com/tachiyomiorg/tachiyomi-extensions/search?q=%22com.google.gson.Gson%22) that uses Gson;~~
- ~~[4 files](https://github.com/tachiyomiorg/tachiyomi-extensions/search?q=%22com.google.gson.JsonParser%22) that uses JsonParser from Gson;~~
- ~~[14 files](https://github.com/tachiyomiorg/tachiyomi-extensions/search?q=%22com.github.salomonbrys.kotson%22) that uses Kotson; and~~
- [7 files](https://github.com/tachiyomiorg/tachiyomi-extensions/search?q=%22org.json%22) that uses org.json

All the usages of other Json libraries for deserialization should be replaced by kotlinx.serialization.

## Other details

The `Json` singleton instance with `ignoreUnknownKeys = true` from the main app can be reused:

```kotlin
private val json: Json by injectLazy()
```

It is not needed to add the kotlinx.serialization dependency to the extensions that uses it (with Json or ProtoBuf), since it was already added to `common-dependencies.gradle`. However, extensions that make usage of the library **must** add this one line after the `kotlin-android` plugin at the extension's Gradle file:

```gradle
apply plugin: 'kotlinx-serialization'
```

## Additional documentation

- [Using Json from kotlinx.serialization](https://github.com/Kotlin/kotlinx.serialization/blob/master/docs/json.md)
- [Using ProtoBuf from kotlinx.serialization](https://github.com/Kotlin/kotlinx.serialization/blob/master/docs/formats.md#protobuf-experimental)
- [Example of Json usage in extension](https://github.com/tachiyomiorg/tachiyomi-extensions/tree/master/src/pt/bruttal/src/eu/kanade/tachiyomi/extension/pt/bruttal)
- [Example of ProtoBuf usage in extension](https://github.com/tachiyomiorg/tachiyomi-extensions/tree/master/src/all/mangaplus/src/eu/kanade/tachiyomi/extension/all/mangaplus)

ObserverOfTime:
I don't understand why you're replacing the builtin `org.json` with an external library that isn't actually any better..

CarlosEsco:
consistency across extensions.

h-hyuuga:
[...]

PR #7401, opened 2021-06-03
MangaMutiny - Migration from Gson to kotlinx.serialization

E3FxGaming:
With the changes of this pull request I'm getting rid of the GSON library for the MangaMutiny extension and instead use kotlinx.serialization (reason for this change is #7392).

The changes made with this pull request will not affect extension users.

I avoided adding data classes annotated with Serializable, since I did not want to introduce additional classes to the extension. DeserializationStrategy objects can be just as useful and don't require lots of instances at runtime.


alessandrojean:
This DeserializationStrategy thing is very cool.

PR #7451, opened 2021-06-04
CatManga QoL improvements

ivaniskandar:
One endpoint for all operations except for fetching pages. Should be faster when refreshing library.

### DIFF
--- a/src/en/catmanga/src/eu/kanade/tachiyomi/extension/en/catmanga/CatManga.kt
+++ b/src/en/catmanga/src/eu/kanade/tachiyomi/extension/en/catmanga/CatManga.kt
@@ -15,10 +15,11 @@ import org.json.JSONArray
 import org.json.JSONObject
 import org.jsoup.nodes.Document
 import rx.Observable
-import uy.kohesive.injekt.Injekt
-import uy.kohesive.injekt.api.get
+import uy.kohesive.injekt.injectLazy
 
 class CatManga : HttpSource() {
+
+    private val application: Application by injectLazy()
 
     override val name = "CatManga"
     override val baseUrl = "https://catmanga.org"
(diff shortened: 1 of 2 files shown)


src/en/catmanga/src/eu/kanade/tachiyomi/extension/en/catmanga/CatManga.kt
```diff
@@ -15,11 +15,12 @@ import org.json.JSONArray
 import org.json.JSONObject
 import org.jsoup.nodes.Document
 import rx.Observable
-import uy.kohesive.injekt.Injekt
-import uy.kohesive.injekt.api.get
+import uy.kohesive.injekt.injectLazy
 
 class CatManga : HttpSource() {
 
+    private val application: Application by injectLazy()
```
CarlosEsco, line 22: Can't you inject the source pref directly similar to 
https://github.com/tachiyomiorg/tachiyomi-extensions/blob/master/src/all/mangadex/src/eu/kanade/tachiyomi/extension/all/mangadex/MangaDex.kt#L43

src/en/catmanga/src/eu/kanade/tachiyomi/extension/en/catmanga/CatManga.kt
```diff
@@ -202,6 +226,14 @@ class CatManga : HttpSource() {
...
+        throw UnsupportedOperationException("Not used.")
+    }
+
+    override fun chapterListParse(response: Response): List<SChapter> {
+        throw UnsupportedOperationException("Not used.")
+    }
+
     override fun searchMangaParse(response: Response): MangasPage {
         throw UnsupportedOperationException("Not used.")
     }
```
CarlosEsco, line 239: I think in general we are trying to switch extensions from gson and the internal json to kotlinx.  So if you could do that in this PR

src/en/catmanga/src/eu/kanade/tachiyomi/extension/en/catmanga/CatManga.kt
ivaniskandar: I don't think so, the shared prefs is different for each manga. https://github.com/tachiyomiorg/tachiyomi-extensions/pull/7451/files#diff-f22fc81b84173dcff99a1c03c2f56b42717f74c380fd06982a9d5616be88c2f8R84

CarlosEsco: hmm why is the shared preference different per a manga and not just a different entry?

src/en/catmanga/src/eu/kanade/tachiyomi/extension/en/catmanga/CatManga.kt
ivaniskandar: It doesn't use any gson stuff here. Both [JSONArray](https://developer.android.com/reference/org/json/JSONArray) and [JSONObject](https://developer.android.com/reference/org/json/JSONObject) are available in Android's SDK.

I've tried switching to kotlinx.serialization without data classes and the code became way less readable. And as for now with the current JSON format in their site I don't know how to do it with data class without any workarounds.

src/en/catmanga/src/eu/kanade/tachiyomi/extension/en/catmanga/CatManga.kt
ivaniskandar: Should've request the change when I first implemented this 😅 But okay I'll make it populated inside single prefs

CarlosEsco: sorry i didn't notice its fine how it is then.

src/en/catmanga/src/eu/kanade/tachiyomi/extension/en/catmanga/CatManga.kt
ObserverOfTime: > I've tried switching to kotlinx.serialization without data classes and the code became way less readable.

I 100% agree with this statement.

h-hyuuga: kotlinx migration for catmanga is already done @CarlosEsco 

https://github.com/tachiyomiorg/tachiyomi-extensions/pull/7407

CarlosEsco: @h-hyuuga that will probably need a slight refactor after I merge this

ivaniskandar: I'll try to pick @h-hyuuga changes to my branch as discussed

h-hyuuga: Don't, someone else already is

ivaniskandar: Okay then I'll mark this PR ready.

PR #7508, opened 2021-06-06
add / to referrer url

OncePunchedMan:
thx @nar1n bby for the tip


nar1n (approved):
Can confirm this worked for paperback

PR #7532, opened 2021-06-07
add fumo

az4521:


CrepeTF:
This extension has changed my life for the better, I am now 110% more fumo than before and no longer shrimp vtubas. Fumopproved.

PR #7533, opened 2021-06-07
Added Colored Council

KojoZero:
Added Colored Council to the Guya Multisrc


multisrc/src/main/java/eu/kanade/tachiyomi/multisrc/guya/GuyaGenerator.kt
```diff
@@ -14,6 +14,7 @@ class GuyaGenerator : ThemeSourceGenerator {
     override val sources = listOf(
         SingleLang("Guya", "https://guya.moe", "en", overrideVersionCode = 18),
         SingleLang("Danke fürs Lesen", "https://danke.moe", "en", className = "DankeFursLesen"),
+        SingleLang("Colored Council", "https://coloredcouncil.moe", "en", className = "ColoredCouncil"),
```
Riztard, line 17: doesnt really need class name since it can be done automatically from "Colored Council"

CarlosEsco: Once this change is made will merge

PR #7549, opened 2021-06-08
Use correct field when parsing manga pub status

animusfracto:
After migrating all of my manga from (old) MangaDex to (new) MangaDex, I noticed that all of the statuses were showing up as 'Unknown', despite the api returning the correct status.

Note: this does nothing to address the 'completed' or 'canceled' statuses that MangaDex can return


CarlosEsco:
can you pr this to my branch (not it uses kotlinx now)
https://github.com/carlosesco/tachiyomi-extensions/tree/dexKotlinx

animusfracto:
PR opened to your branch as requested.

PR #7589, opened 2021-06-10
[Mangabox] Revert Mangakakalot dateformat, add Mangakakalots dateformat

### LINKED ISSUE #7318: [Bug] <mangakakalot>: <Some manga have different chapter dates>
---

## Device information
- Tachiyomi version: 1.0.17
- Android version: 11

## Source information
- Name: mangakakalot.com
- Extension version: 1.2.4

## Steps to reproduce
1. Check manga like Awkward Senpai

### Expected behavior
Date uploaded should show

### Actual behavior
Date uploaded is missing

FlaminSarge:
Closes #7318 (I think).

### DIFF
--- a/multisrc/overrides/mangabox/mangakakalot/src/Mangakakalot.kt
+++ b/multisrc/overrides/mangabox/mangakakalot/src/Mangakakalot.kt
@@ -2,10 +2,8 @@ package eu.kanade.tachiyomi.extension.en.mangakakalot
 
 import eu.kanade.tachiyomi.multisrc.mangabox.MangaBox
 import okhttp3.Headers
-import java.text.SimpleDateFormat
-import java.util.Locale
 
-class Mangakakalot : MangaBox("Mangakakalot", "https://mangakakalot.com", "en", SimpleDateFormat("MMM dd,yy", Locale.ENGLISH)) {
+class Mangakakalot : MangaBox("Mangakakalot", "https://mangakakalot.com", "en") {
     override fun headersBuilder(): Headers.Builder = super.headersBuilder().set("Referer", "https://manganato.com") // for covers
     override val simpleQueryPath = "search/story/"
     override fun searchMangaSelector() = "${super.searchMangaSelector()}, div.list-truyen-item-wrap"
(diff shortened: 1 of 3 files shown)


FlaminSarge:
Doublechecked both sources, the updated dateformats work.

Seishirou101:
it uses two formats, you'll just be breaking the current format if you revert it so you have to account for it if you want to fix for regular kakalot

FlaminSarge:
@Seishirou101 got example urls I can check against?

Seishirou101:
the linked issue tells you, one uses `MMM-dd-yy` which is the regular mangakakalot date format, and then `MMM dd,yy` when it uses manganelo's format and redirects there

Seishirou101:
actually, it looks like the fix for the dates went into effect b4 the fix for nelo -> nato. can you double check if this fixes when its redirected to nato's domain on kakalot? you can easily find a link by opening it in browser and clicking on manga that redirect you to nato. Tales of Demons and Gods is one and should be the first one if you hit browse on kakalot

curche:
Few examples that I noted when the issue was first raised in #support channel:
 1. [Solo Leveling](https://mangakakalot.com/search/story/solo_leveling) - the title redirects to readmanganato
 2. [Matchmaking of the Amagami Household](https://mangakakalot.com/search/story/amagami_household) - an actual mangakakalot entry
 3. [Awkward Senpai](https://mangakakalot.com/search/story/awkward_senpai) - same as 2

In fact, if you look at their [Latest page](https://mangakakalot.com/manga_list?type=latest&category=all&state=all&page=1), you see that some of the links are actually to readmanganato (as far as I can see, spread randomly)

The problem with #7318, is that manga in mangakakalot website are actually linking to readmanganato manga, which has a different date format ( Jun 03,21 instead of  Jun-03-21)

also, [Mangakakalots](https://ww2.mangakakalots.com/manga/) aka Mangakakalots Unoriginal is a different clone which has a date problem no one reported yet

FlaminSarge:
Ah yeah in my testing I checked specifically the Solo Leveling case (nato redirect) as well as kakalot native manga. I also fixed the unreported issue on kakalots. All cases worked.

PR #7604, opened 2021-06-11
[Madara] Fix dateformat for TheFluffyHangoutGroup

FlaminSarge:
Now uses e.g. '2021-06-10'


FlaminSarge:
These are the only changes needed to add a class override when the classname/pkg already match the source name, right?

PR #7608, opened 2021-06-11
Hentaimimi proper fix

arisudesu:
The previous pull request #7574 didn't fix the slashes issue. I'm really sorry for submitting not working fix.

To fix my situation, I had to put my hands in Kotlin and figure out how to fix the parsing of the list of pages. It turns out that simple js arrays, such as a list of pages on their site, can be parsed as json, and you don't have to bother with handling escaped strings at all.

This time, I carefully checked and tested on a device that the changes worked, and additionally checked other special elements such as ?, #, @, /, [, ] too - all pages parsed without errors.

### DIFF
--- a/src/en/hentaimimi/build.gradle
+++ b/src/en/hentaimimi/build.gradle
@@ -1,5 +1,6 @@
 apply plugin: 'com.android.application'
 apply plugin: 'kotlin-android'
+apply plugin: 'kotlinx-serialization'
 
 ext {
     extName = 'HentaiMimi'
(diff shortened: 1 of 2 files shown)


alessandrojean:
Need to add the following line to extension `build.gradle` (right after the `kotlin-android` plugin line):

```gradle
apply plugin: 'kotlinx-serialization'
```

jobobby04:
Thats only needed if your doing the parse to a dto, but there is no harm in adding it anyway

denveroh:
No registred instance of factory for type class kotlinx.serialization.json.json

PR #7620, opened 2021-06-12
Replace JsonParser with kotlinx.serialization in some extensions

alessandrojean:
Replace the usages of Gson's `JsonParser` by kotlinx.serialization as per #7392.


alessandrojean:
I'm done with this one for now, it's ready to merge.

PR #7656, opened 2021-06-14
Remove last Gson usage at MangaDex

alessandrojean:


CarlosEsco:
Can you make it a dto?  Since the rest of MangaDex uses dto's i would prefer it to stay consistent
see below
https://github.com/CarlosEsco/Neko/blob/master/app/src/main/java/eu/kanade/tachiyomi/source/online/MangaDex.kt#L130-L149

alessandrojean:
Updated to use the existing DTO.

Issue #7668, opened 2021-06-15
[Bug] <Dmzj>: Domain dmzj1.com is unreachable


Myuki:
## Device information
- Tachiyomi version: 0.11.1
- Android version: 11

## Source information
- Name: Dmzj
- Extension version: 1.2.19

## Steps to reproduce
1. Try any comic in this source

### Expected behavior
Reachable resource

### Actual behavior
Unreachable resource

## Other details

It seems as DMZJ has deprecated `dmzj1.com` and return to use `dmzj.com` as the main domain. `dmzj1.com` don't have the DNS record and all comic can't fetch now.

zhongfly:
确实，中午时还能检查更新的

PR #7712, opened 2021-06-17
zh-dmzj: replace dmzj.com with dmzj1.com for v4api only to make it work again

erinacio:
The new v4api doesn't work if using `nnv4api.dmzj.com` as api url, only `nnv4api.dmzj1.com` works. The `dmzj.com` version will always return an error message indicating `暂时无法观看` (with errno `3`) even if the manga is not blocked (i.e. can read directly from website).

The `dmzj1.com` domain is back online again, including its websites: https://www.dmzj1.com/ ([archive](https://web.archive.org/web/20210617223910/https://www.dmzj1.com/)) and https://manhua.dmzj1.com/ ([archive](https://web.archive.org/web/20210617224035/https://manhua.dmzj1.com/)). The previous issue (https://github.com/tachiyomiorg/tachiyomi-extensions/issues/7668) may be just a DNS misconfiguration incident. I only reverted change of v4api because it seems that other apis still work under `dmzj.com`, only v4api breaks.

### DIFF
--- a/src/zh/dmzj/build.gradle
+++ b/src/zh/dmzj/build.gradle
@@ -6,7 +6,7 @@ ext {
     extName = 'Dmzj'
     pkgNameSuffix = 'zh.dmzj'
     extClass = '.Dmzj'
-    extVersionCode = 20
+    extVersionCode = 21
     libVersion = '1.2'
 }
 
--- a/src/zh/dmzj/src/eu/kanade/tachiyomi/extension/zh/dmzj/Dmzj.kt
+++ b/src/zh/dmzj/src/eu/kanade/tachiyomi/extension/zh/dmzj/Dmzj.kt
@@ -48,7 +48,7 @@ class Dmzj : ConfigurableSource, HttpSource() {
     private val v3apiUrl = "https://v3api.dmzj.com"
     private val v3ChapterApiUrl = "https://nnv3api.dmzj.com"
     // v3api now shutdown the functionality to fetch manga detail and chapter list, so move these logic to v4api
-    private val v4apiUrl = "https://nnv4api.dmzj.com" // https://v4api.dmzj.com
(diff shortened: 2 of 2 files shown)


zhongfly:
but I find latest dmzj's app use muwai.com instead of dmzj.com, such like http://nnv4api.muwai.com/comic/detail/ or https://images.muwai.com/tuijian/750_480/210513xinman97tj5.jpg

erinacio:
> but I find latest dmzj's app use muwai.com instead of dmzj.com, such like http://nnv4api.muwai.com/comic/detail/ or https://images.muwai.com/tuijian/750_480/210513xinman97tj5.jpg
> edit: Finally I find https://m.muwai.com/ just same as https://m.dmzj.com.

Should we switch all domains to muwai.com instead of dmzj.com and dmzj1.com? Maybe we can also make it an option, and let users choice which domain to use, in case that the default domain is down.

zhongfly:
and I have tested nnv4api.dmzj.com now ,it work well and doesn't return error message.
Maybe it just your network problem or accident?

erinacio:
> and I have tested nnv4api.dmzj.com now ,it work well and doesn't return error message.
> Maybe it just your network problem or accident?

No, it still returns error on mangas that is not blocked. The manga I test with is https://m.dmzj.com/info/40135.html .

zhongfly:
> > and I have tested nnv4api.dmzj.com now ,it work well and doesn't return error message.
> > Maybe it just your network problem or accident?
> 
> No, it still returns error on mangas that is not blocked. The manga I test with is https://m.dmzj.com/info/40135.html .
> 
> If v4api failed, the plugin will fallback to v3api. In chapter list, something like `连载: 43话` indicates v4api is working, and `第43话` means v4api is not working, and v3api is used.

Yes,I see.I only test an chinese original comic,and I don't know it will be different to translate comics.
Making an option is a good idea, to choise `dmzj.com` `dmzj1.com` or `muwai.com`

Issue #7661, opened 2021-06-15
[Source Request] <Helvetica Scans>


FireNation323:
## Source website
- Name: Helvetica scans
- Link: https://helveticascans.com/
- Language(s): English

loocool2:
MangAdventure

FireNation323:
> MangAdventure

?

loocool2:
That's the site template it's built on

datreesezcup:
I'm not sure if a new extension is even needed: it would appear that Helvetica Scans is an exact mirror of Assorted Scans.
Take a look: the sites look to be identical:
- https://assortedscans.com/
- https://helveticascans.com

And: the site title in the tab for Helvetica Scans says: "Home · Assorted Scans"

So as of now, I believe the currently existing "Assorted Scans" extension will work for this source.
(Unless I'm missing something: I have not had time to check if everything is identical between the two sites)

ObserverOfTime:
Helvetica does indeed point to Assorted Scans and the extension already works.
https://github.com/tachiyomiorg/tachiyomi-extensions/blob/516dfdbd364ff933cf788c226bccaeb0f84a698b/multisrc/overrides/mangadventure/assortedscans/AndroidManifest.xml#L18

PR #7816, opened 2021-06-22
Revert "Add Helvetica Scans (#7773)"

ObserverOfTime:
The new extension is redundant. See https://github.com/tachiyomiorg/tachiyomi-extensions/issues/7661#issuecomment-865977036.


nar1n:
Blame @curche

arkon:
A lot of extensions are redundant, I'm not sure why this one is an issue?

ObserverOfTime:
I'm not sure why you merged the original PR without question when the extension's redundancy had already been explained by @datreesezcup and yet you're questioning the purpose of this one.

arkon:
Because we don't have a hard rule against "redundant" websites. If we did, we wouldn't have things like MangaSee/MangaLife or Mangakakalot/MangaNato.

This will cut down on the inevitable issues created to request it too, and the cost of supporting it is so low that it's not worth _not_ having it.

ObserverOfTime:
Then at least move the helveticascans.com deeplinks to it and copy the icons over as well.

arkon:
You could open a PR if you have time.

ObserverOfTime:
No, I don't think I will.

Soitora:
> No, I don't think I will.

Then we won't either, lol.

ObserverOfTime:
You said "the cost of supporting it is so low" so go ahead and support it properly then.

PR #7837, opened 2021-06-24
[RU]YagamiProject artbooks + artist and qualitative cover

e-shl:


src/ru/yagamiproject/src/eu/kanade/tachiyomi/extension/ru/mangabook/YagamiProject.kt
```diff
@@ -148,8 +148,16 @@ class YagamiProject : ParsedHttpSource() {
             }
         }
     }
-    override fun imageUrlParse(response: Response): String = response.asJsoup().select("#page img").attr("src")
-    override fun imageUrlParse(document: Document): String = throw Exception("imageUrlParse(document: Document) Not Used")
+    override fun imageUrlParse(document: Document): String {
```
Pavkazzz, line 151: override fun imageUrlParse(document: Document): String {
        val defaultimg = document.select("#page img").attr("src")
        return if (defaultimg != "string(1) ") {
            defaultimg
        } else {
            document.select("#get_download").first().attr("href")
        }
    }

PR #7841, opened 2021-06-24
[Mangabox] Separate authors with commas

FlaminSarge:


FlaminSarge:
Tested for all 5 sources with both manga that had single authors and multiple authors.

PR #7911, opened 2021-06-28
Add new multisrc extension (Weebreader)

### LINKED ISSUE #7908: Arang Scans: No Longer Using Madara
### Acknowledgements

- [X] I have updated the app to version **[0.11.1](https://github.com/tachiyomiorg/tachiyomi/releases/latest)**.

### Tachiyomi version

Not applicable

### Android version

Not applicable

### Source information

Arang Scans 1.2.10

### Steps to reproduce

Attempt to use the extension in any way

### Expected behavior

Normal extension things

### Actual behavior

Returnw HTTP 521

### Other details

Seems to be using: https://github.com/Dinokin/WeebReader

Users will have to migrate titles after a fix is pushed -> uses different URL scheme

h-hyuuga:
This pull request adds new multisrc extension weebreader to the repo.

Sources:

* NaniScans (Used as the base implementation)
* Arang Scans

Close #7908


Daktyl198:
Why not use the weebreader API that's available, like this other extension is doing?
https://github.com/tachiyomiorg/tachiyomi-extensions/blob/master/src/en/naniscans/src/eu/kanade/tachiyomi/extension/en/naniscans/NaniScans.kt

h-hyuuga:
> Why not use the weebreader API that's available, like this other extension is doing?
> https://github.com/tachiyomiorg/tachiyomi-extensions/blob/master/src/en/naniscans/src/eu/kanade/tachiyomi/extension/en/naniscans/NaniScans.kt

Had no idea there was an API.

Thanks for bringing this to my attention

CarlosEsco:
this WIP then so you can switch to api?

h-hyuuga:
> this WIP then so you can switch to api?

I already did.

PR #7966, opened 2021-07-04
Another mangaspw fix

DTecheira:
They changed some tags

### DIFF
--- a/multisrc/src/main/java/eu/kanade/tachiyomi/multisrc/mmrcms/MMRCMSSources.kt
+++ b/multisrc/src/main/java/eu/kanade/tachiyomi/multisrc/mmrcms/MMRCMSSources.kt
@@ -32,7 +32,7 @@ class MMRCMSSources {
                 override val overrideVersionCode: Int = 0,
             ) : SourceData()
         }
-        val version: Int = 4
+        val version: Int = 5
         val sourceList: List<SourceData.Single> = listOf(
             SourceData.Single("مانجا اون لاين", "https://onma.me", "ar", className = "onma"),
             SourceData.Single("Read Comics Online", "https://readcomicsonline.ru", "en"),
--- a/multisrc/src/main/java/eu/kanade/tachiyomi/multisrc/mmrcms/MMRCMSSources.kt
+++ b/multisrc/src/main/java/eu/kanade/tachiyomi/multisrc/mmrcms/MMRCMSSources.kt
@@ -32,7 +32,7 @@ class MMRCMSSources {
                 override val overrideVersionCode: Int = 0,
             ) : SourceData()
         }
-        val version: Int = 5
+        val version: Int = 4
         val sourceList: List<SourceData.Single> = listOf(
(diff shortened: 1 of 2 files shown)


multisrc/src/main/java/eu/kanade/tachiyomi/multisrc/mmrcms/MMRCMSSources.kt
```diff
@@ -32,7 +32,7 @@ class MMRCMSSources {
                 override val overrideVersionCode: Int = 0,
             ) : SourceData()
         }
-        val version: Int = 4
+        val version: Int = 5
```
Riztard, line 35: just use `overrideVersionCode` instead, since the fix only for mangas.pw

DTecheira: @Riztard I should put 5? or is ok start with 1?

`SourceData.Single("Mangas.pw", "https://mangas.in", "es", className = "MangasPw", overrideVersionCode = 5),`

Elcrack14: El manga.pw sólo muestra un capítulo y ese lleva a un hentai arreglen lo por favor

Riztard: 1

DTecheira: @Elcrack14 tu comentario no ayuda

DTecheira: Done

Clouddark75: It's intentional, they have been doing it for the past years. The best option is removing the source. They don't want their website here.

DTecheira: ok, we should close this PR and open an issue to remove the extension

DTecheira:
https://github.com/tachiyomiorg/tachiyomi-extensions/pull/7966#discussion_r663507544

multisrc/src/main/java/eu/kanade/tachiyomi/multisrc/mmrcms/MMRCMSSources.kt
Clouddark75: I was on their Discord like on Jun and they simply ignored my question, they said they could have a harsh limited timer for their page so user don't hit the website that badly, but I got no answers.
So they clearly don't want any scrapper on their website anymore. It's sad to see their behavior, but the people using Tachi aren't just two or three people.

PR #7972, opened 2021-07-04
[MangaKatana] Add server preference, alt names in description, and comma-separate author list

### LINKED ISSUE #5939: [Meta] Mangakatana 3 image servers
https://mangakatana.com/ does have 3 image servers, and currently only server 1 is used right?

All tachiyomi users use same image server, and this causes slow downs for sure, if there would be an option to change between 3 image servers, it would likely go more evenly, and less stress on server 1.

Similar option what mangadex is using, some image server even might be better option for different country.

### LINKED ISSUE #6617: [Enhancement] MangaKatana: Show Alternate Names, Add Punctuation in Author(s) Field
Title is pretty self-explanatory. The extension at the moment doesn't show alternate names, but accurate results still show up if searched using an alternate one. Maybe it could be displayed in the description itself, like how the Hentai2Read extension does it.

There's also no comma to separate the author and the artist in the extension, but this is not the case in the website. Just a small QoL feature. Example given below.

Screenshot_20210419-141401
Screenshot_20210419-141431

FlaminSarge:
Fixes #5939
Fixes #6617

### DIFF
--- a/src/en/mangakatana/src/eu/kanade/tachiyomi/extension/en/mangakatana/MangaKatana.kt
+++ b/src/en/mangakatana/src/eu/kanade/tachiyomi/extension/en/mangakatana/MangaKatana.kt
@@ -1,6 +1,11 @@
 package eu.kanade.tachiyomi.extension.en.mangakatana
 
+import android.app.Application
+import android.content.SharedPreferences
+import androidx.preference.ListPreference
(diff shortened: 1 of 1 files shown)


FlaminSarge:
Please let me know if there's better ways to do these things in Kotlin.

Issue #5767, opened 2021-02-11
Multisrc extensions silently shadow individual extensions


arkon:
Seen in #5763: if the package/module name shadows another one, only one of them gets built.

jopejoe1:
When the non Multisrc extension was installed before switching over to Multisrc the Multisrc extension is missing in the extension list in the Tachiyomi app. Leading to what i guess happened in #5763.

PR #8002, opened 2021-07-07
add two new sources to madara & remove two sources

CVIUS:
- add aqua manga & mangatone
- remove mangalazy, mangadna


Riztard:
there is already mangafast ext so i think it will cause some issue #5767

CVIUS:
I made it manga-fast.com, wdyt?

Riztard:
try build on pc?

Riztard:
why remove mangalazy and mangadna?

CVIUS:
> why remove mangalazy and mangadna?

its not working, and can't fix it 😅 and I'll make a request about it instead after this is merged

Riztard:
i think better to say something like `remove mangaduck`, cuz ext changelog is using github commit

PR #8086, opened 2021-07-12
Update url

matheuswr89:

### DIFF
--- a/multisrc/overrides/madara/neoxxxscans/src/NeoXXXScans.kt
+++ b/multisrc/overrides/madara/neoxxxscans/src/NeoXXXScans.kt
@@ -11,7 +11,7 @@ import java.util.concurrent.TimeUnit
 @Nsfw
 class NeoXXXScans : Madara(
     "NeoXXX Scans",
-    "https://xxx.neoxscans.net",
+    "https://xxx.neoxscans.com",
     "pt-BR",
     SimpleDateFormat("dd/MM/yyyy", Locale("pt", "BR"))
 ) {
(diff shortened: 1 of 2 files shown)


CarlosEsco:
@alessandrojean do you want to just make this configurable?

matheuswr89:
I think this would be a good thing for this extension, as it uses the domain of another one, that is, if it breaks there, the same happens here.

alessandrojean:
@CarlosEsco sure. I can update the hentai website URL to the new one and add the preference in a new PR, so users will not need to update the extension twice.

> if it breaks there, the same happens here.

Extensions can't read settings from each other, though.

matheuswr89:
> Extensions can't read settings from each other, though.
I know, I mean when you change Neox URL

jobobby04:
> 
> 
> @CarlosEsco sure. I can update the hentai website URL to the new one and add the preference in a new PR, so users will not need to update the extension twice.
> 
> > if it breaks there, the same happens here.
> 
> Extensions can't read settings from each other, though.

They can if you setup the shared preferences correctly, just use the same id for the key

alessandrojean:
> They can if you setup the shared preferences correctly, just use the same id for the key

Probably would be confusing for the users to setup the URL though, and when Neox changed the URL of the main website, it took some days for them to update the hentai one, so users in this period would not be able to use the extension if they changed the URL in the main preference.

I will just let them separate because it will be more easy for the users.

matheuswr89:
@alessandrojean can I close this PR?

PR #8097, opened 2021-07-13
Update NetTruyen.kt

haruki-takeshi:
Update Nettruyenvip.com


Riztard:
like the other one. update ext version?

haruki-takeshi:
> like the other one. update ext version?

Update URL only

arkon:
You need to update the extension version.

PR #8112, opened 2021-07-14
added gogomanga.org to wpmangastream & fix mangaowl

CVIUS:


src/en/mangaowl/src/eu/kanade/tachiyomi/extension/en/mangaowl/MangaOwl.kt
```diff
@@ -93,7 +93,7 @@ class MangaOwl : ParsedHttpSource() {
 
     override fun searchMangaFromElement(element: Element): SManga = popularMangaFromElement(element)
 
-    override fun searchMangaNextPageSelector() = popularMangaNextPageSelector()
+    override fun searchMangaNextPageSelector() = "div.blog-pagenat-wthree li.page-item.next"
```
CarlosEsco, line 96: :disappointed:

Riztard:
i think it better to make separate PR
since there is no similarities/connection

Riztard:
people be like...what is manga owl has to do with madara or add teenmanhua

Issue #8071, opened 2021-07-12
NetTruyen and NhatTruyen: HTTP error 403


Makro297:
### Acknowledgements

- [X] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open issue.
- [X] I have written a short but informative title.
- [X] I have updated the app to version **[0.11.1](https://github.com/tachiyomiorg/tachiyomi/releases/latest)**.
- [X] I have updated all installed extensions.
- [X] I have tried the [troubleshooting guide](https://tachiyomi.org/help/guides/troubleshooting/).
- [X] If this is an issue with the app itself, I should be opening an issue in the [app repository](https://github.com/tachiyomiorg/tachiyomi/issues/new/choose).
- [X] I will fill out all of the requested information in this form.

### Tachiyomi version

0.11.1

### Android version

Android 9

### Source information

NetTruyen version 1.2.3

### Steps to reproduce

1. Browse
2. Click on extension NetTruyen ( version 1.2.3, Vietnamese) and extension NhatTruyen ( version 1.2.3, Vietnamese)

### Expected behavior

manga loads normally

### Actual behavior

It showed HTTP error 403
![image](https://user-images.githubusercontent.com/87240370/125277252-66ab2580-e33b-11eb-9a4b-3140185917eb.png)

### Other details

I have check website in WebView, the website still work.
![image](https://user-images.githubusercontent.com/87240370/125277679-f224b680-e33b-11eb-9ae2-9a6c89888dcb.png)

PR #8194, opened 2021-07-20
update WPComics multiSrc

hntidat:
Update the header of WPComics.
Related issues:
Issue #8071


Riztard:
i think some of the source need that, and since that change only needed for those two source maybe just use override instead?

hntidat:
> i think some of the source need that, and since that change only needed for those two source maybe just use **override file** instead?

I don't think so, all of the sources in this WPcomic have the same issue.

hntidat:
> i think some of the source need that, and since that change only needed for those two source maybe just use **override file** instead?

Nettruyen, TruyenChon and Nhattruyen source have the same issue. But you are right, it should be used override file instead.

PR #8250, opened 2021-07-24
Inspect extensions to get source information

Syer10:
Continues #6410

This is an rewritten version of the Tachidesk Inspector that ArMor PR'd back in April. It takes the suggestions of other contributors, such as:

- Puts the array of source information(such as name, lang, and id) inside the index.json as a item in the extensions object. 
- Deletes all Tachidesk specific code, only adding whats needed for the Inspector
- Puts the Inspector.jar in the runner-files so that it is not downloaded externally

Sample from index.json
```json
[
  {
    "name": "Tachiyomi: Cubari",
    "pkg": "eu.kanade.tachiyomi.extension.all.cubari",
    "apk": "tachiyomi-all.cubari-v1.2.9.apk",
    "lang": "all",
    "code": 9,
    "version": "1.2.9",
    "nsfw": 0,
    "sources": [
      {
        "name": "Cubari",
        "lang": "en",
        "id": 6338219619148106000
      },
      {
        "name": "Cubari",
        "lang": "all",
        "id": 1470847599087460400
      },
      {
        "name": "Cubari",
        "lang": "other",
        "id": 2013845246758512400
      }
    ]
  },
  ...
]
```

Successful build of the index.json:
https://github.com/Syer10/tachiyomi-extensions/blob/repo/index.json

Successful run for the Inspector.jar:
https://github.com/Syer10/tachiyomi-extensions/actions/runs/1061635114
The Inspector.jar only takes around 30 seconds to go over around 1000 extensions.

Repo for the Inspector.jar:
https://github.com/Syer10/TachiExtension-Inspector
This can be transferred to the organization

### DIFF
--- a/.github/workflows/build_push.yml
+++ b/.github/workflows/build_push.yml
@@ -128,6 +128,11 @@ jobs:
           ref: master
           path: master
 
+      - name: Set up JDK
+        uses: actions/setup-java@v1
+        with:
+          java-version: 1.8
+
       - name: Sign APKs
         run: |
           cd master
@@ -136,6 +141,11 @@ jobs:
             ${{ secrets.ALIAS }} \
             ${{ secrets.KEY_STORE_PASSWORD }} \
             ${{ secrets.KEY_PASSWORD }}
+
+      - name: Run inspector
+        run: |
+          cd master
+          java -jar ./.github/runner-files/Inspector.jar "apk" "output.json" "tmp"
 
       - name: Create repo artifacts
         run: |
--- a/.github/workflows/build_push.yml
+++ b/.github/workflows/build_push.yml
@@ -145,7 +145,7 @@ jobs:
       - name: Run inspector
         run: |
           cd master
-          INSPECTOR_LINK="$(curl -s "https://api.github.com/repos/tachiyomiorg/tachiyomi-extensions-inspector/releases/latest" | grep -o "https.*download\/[a-zA-Z0-9.\/-]*")"
+          INSPECTOR_LINK="$(curl -s "https://api.github.com/repos/tachiyomiorg/tachiyomi-extensions-inspector/releases/latest" | jq -r '.assets[0].browser_download_url')"
           curl -L "$INSPECTOR_LINK" -o ./Inspector.jar
           java -jar ./Inspector.jar "apk" "output.json" "tmp"
 
--- a/.github/workflows/build_push.yml
+++ b/.github/workflows/build_push.yml
@@ -131,7 +131,7 @@ jobs:
       - name: Set up JDK
         uses: actions/setup-java@v1
         with:
-          java-version: 1.8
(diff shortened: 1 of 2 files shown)


.github/workflows/build_push.yml
```diff
@@ -128,6 +128,11 @@ jobs:
           ref: master
           path: master
 
+      - name: Set up JDK
+        uses: actions/setup-java@v1
+        with:
+          java-version: 1.8
```
arkon, line 134: Does it work with JDK11?

.github/workflows/build_push.yml
```diff
@@ -137,6 +142,11 @@ jobs:
             ${{ secrets.KEY_STORE_PASSWORD }} \
             ${{ secrets.KEY_PASSWORD }}
 
+      - name: Run inspector
+        run: |
+          cd master
+          java -jar ./.github/runner-files/Inspector.jar "apk" "output.json" "tmp"
```
arkon, line 148: It'd be ideal if the JAR came from a trusted place (a repo release artifact, maybe) from a tachiyomiorg repo.

arkon (commented):
Seems okay overall. I think I'd want the inspector stuff to be transferred over first though, so there's a vaguely higher trustworthiness for it.

For the future: I wonder if it'd also be useful for the inspector to extract our the base URLs? Could maybe even infer its NSFW-ness based on the annotations.

.github/workflows/build_push.yml
Syer10: Tachidesk works on any JDK after 8, so the inspector should be the same

arkon: Might as well bump this from the get-go then.

Syer10:
> For the future: I wonder if it'd also be useful for the inspector to extract our the base URLs? Could maybe even infer its NSFW-ness based on the annotations.

I can see this being useful for the repo actions and website, as well as the nsfw being useful in the app. It would be easy to add to the inspector.

.github/workflows/build_push.yml
```diff
@@ -137,6 +142,13 @@ jobs:
             ${{ secrets.KEY_STORE_PASSWORD }} \
             ${{ secrets.KEY_PASSWORD }}
 
+      - name: Run inspector
+        run: |
+          cd master
+          INSPECTOR_LINK="$(curl -s "https://api.github.com/repos/tachiyomiorg/tachiyomi-extensions-inspector/releases/latest" | grep -o "https.*download\/[a-zA-Z0-9.\/-]*")"
```
arkon, line 148:
```
curl -s https://api.github.com/repos/tachiyomiorg/tachiyomi-extensions-inspector/releases/latest | jq '.assets[0].browser_download_url'
```

Syer10: Done

PR #8264, opened 2021-07-25
fix mangahero url in override & update webtoon18 url

CVIUS:


multisrc/overrides/madara/webtoon18/src/WebToon18.kt
```diff
@@ -8,7 +8,7 @@ import java.text.SimpleDateFormat
 import java.util.Locale
 
 @Nsfw
-class WebToon18 : Madara("WebToon18", "https://webtoon18.net", "en", dateFormat = SimpleDateFormat("d MMMM, yyyy", Locale.US)) {
+class WebToon18 : Madara("WebToon18", "http://webtoon18.net", "en", dateFormat = SimpleDateFormat("d MMMM, yyyy", Locale.US)) {
```
CarlosEsco, line 11: they went to http?

CVIUS: yes

Issue #8178, opened 2021-07-19
Flame Scans: HTTP Error 403


DatHypnoboi:
### Acknowledgements

- [X] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open issue.
- [X] I have written a short but informative title.
- [X] I have updated the app to version **[0.11.1](https://github.com/tachiyomiorg/tachiyomi/releases/latest)**.
- [X] I have updated all installed extensions.
- [X] I have tried the [troubleshooting guide](https://tachiyomi.org/help/guides/troubleshooting/).
- [X] If this is an issue with the app itself, I should be opening an issue in the [app repository](https://github.com/tachiyomiorg/tachiyomi/issues/new/choose).
- [X] I will fill out all of the requested information in this form.

### Tachiyomi version

0.11.1

### Android version

11

### Source information

Flame Scans 1.2.15

### Steps to reproduce

1. Open Flame Scans source
2. Open any manga/manwha/manhua

### Expected behavior

Manga info should show

### Actual behavior

Gives "Http error 403" message instead

mcOz1:
I have the same problem. When I try to open a page in browser on application I got 'access denied', but when I try without application, for example in chrome, everything working normally.
![1](https://user-images.githubusercontent.com/71328023/126330860-008844ec-a5c8-492c-8ae7-2537b331dd7e.PNG)
![2](https://user-images.githubusercontent.com/71328023/126330861-d2cfeb0d-cebd-489a-91fc-556c7708feb1.PNG)

Seishirou101:
They block tachi at peak times/when they want. migrate or wait for them to unblock

8227846265:
Same time is happening from my end on Flame Scans extension.

hyoretsu:
> They block tachi at peak times/when they want. migrate or wait for them to unblock
Last update I received was from at most 3 weeks ago, so damn

Issue #7781, opened 2021-06-21
[Bug] Flame Scans : Cloudflare bypass not working (inf loop)


krakerz:
## Device information
- Tachiyomi version: 0.11.1
- Android version: 10

## Source information
- Name: Flame Scans
- Extension version: 1.2.11

## Steps to reproduce
1. Install the plugins
2. if by chance you get 403 error or no page, try opening the webview (random manga is fine in most case)
3. infinite redirect cloudflare

### Expected behavior
cloudflare bypassed normally even in webview

### Actual behavior
infinite loop @ cloudflare

## Other details

also, Manhwahentai.me (1.2.8) has same issue, tho this one can be fixed temporary by opening @ webview once, and after cloudflare bypassed, it's worked normally again (tho the auto update might not working, since it seems it's got cloudflare again)

Clouddark75:
Clean your cookies and cache (make sure you have the best/last version of Webview installed). 
They're still using a basic version of Cloudflare, so  everything does load fine here with differents devices.

krakerz:
already tried the cookies and cache clear from the app (either from vanilla tachiyomi / tachiyomiSY)
already checked the playstore as well for the webview update (no update tho)
still got the issue sadly ._.

krakerz:
here i attach the vid record for it
https://drive.google.com/file/d/1kE9o52Tght2_BvQV5vxeB39Ag4kkdaIr/view?usp=drivesdk

Clouddark75:
Do you have the same infinite loop with Chrome or another browser on your device?
The problem with Cloudflare is that anything can break their DDoS shield. If you have a dynamic IP try to restart your device. It could be that your IP got flagged. The other thing you can do is wait until it gets fixed by itself.

krakerz:
> Do you have the same infinite loop with Chrome or another browser on your device?
> The problem with Cloudflare is that anything can break their DDoS shield. If you have a dynamic IP try to restart your device. It could be that your IP got flagged. The other thing you can do is wait until it gets fixed by itself.

only within tachiyomi, when i'm using opera mobile, it's working as intended, sadly my IP is static right now so... RIP for me if my IP is blacklisted (inside the tachiyomi tho)
![Screenshot_2021-06-21-14-56-09-95_4641ebc0df1485bf6b47ebd018b5ee76](https://user-images.githubusercontent.com/8030983/122727243-088da400-d2a1-11eb-9aa3-360387d65745.jpg)

[...]

PR #8266, opened 2021-07-25
remove FLAME-SCANS

curche:
closes https://github.com/tachiyomiorg/tachiyomi-extensions/issues/8265


hyoretsu:
Should also close #8178 and #7781

PR #8289, opened 2021-07-27
Add Random Scans source

lord-ne:
Add Random Scans source


src/en/comikey/AndroidManifest.xml
```diff
@@ -23,11 +23,11 @@
                     android:scheme="https" />
 
                 <data
-                    android:host="comikey.com"
+                    android:host="www.comikey.com"
                     android:pathPattern="/comics/..*"
                     android:scheme="https" />
                 <data
-                    android:host="comikey.com"
+                    android:host="www.comikey.com"
```
CarlosEsco, line 30: Was this intended?  PR mentions nothing about this

lord-ne: That was fixing a small typo in Comikey, I didn't realize I accidentally committed it at the same time. Should I leave it and increment the Comikey version number, or should I remove it and put it in another PR?

CarlosEsco: remove and put in another PR

src/en/randomscans/src/eu/kanade/tachiyomi/extension/en/randomscans/RandomScans.kt
```diff
@@ -0,0 +1,279 @@
...
+                    status = when (element.selectFirst("p.has-large-font-size:containsOwn(Status: )")?.text()?.substringAfterLast("Status: ")) {
+                        "Ongoing" -> SManga.ONGOING
+                        "Completed" -> SManga.COMPLETED
+                        "Hiatus" -> {
+                            description = "Status: Hiatus\n\n$description"
+                            SManga.ONGOING
+                        }
+                        "Dropped" -> {
+                            description = "Status: Dropped\n\n$description"
+                            SManga.COMPLETED // Not sure what the best status is for "Dropped"
```
CarlosEsco, line 75: I would put Unknown, because this will break the setting "Only update completed"

src/en/randomscans/src/eu/kanade/tachiyomi/extension/en/randomscans/RandomScans.kt
```diff
@@ -0,0 +1,279 @@
...
+    // oneshots have manga pages which directly link to the chapter page, and have dates in their url like a chapter page
+    private fun isOneshot(manga: SManga): Boolean = manga.url.contains(Regex("/20\\d\\d/\\d\\d/\\d\\d"))
+
+    // We can only get the title
+    private fun oneshotDetailsParse(response: Response): SManga {
+        return SManga.create().apply {
+            val responseJson = response.asJsoup()
+            title = responseJson.selectFirst("h1.entry-title").text()
+            thumbnail_url = responseJson.selectFirst("img[src]").attr("src")
+            status = SManga.COMPLETED
```
CarlosEsco, line 179: why set status at all here?

src/en/randomscans/src/eu/kanade/tachiyomi/extension/en/randomscans/RandomScans.kt
lord-ne: Isn't the setting "Only update ongoing"?

src/en/randomscans/src/eu/kanade/tachiyomi/extension/en/randomscans/RandomScans.kt
lord-ne: A oneshot links directly to a single chapter page, which means that we know it's completed and there won't be any more chapters.

src/en/randomscans/src/eu/kanade/tachiyomi/extension/en/randomscans/RandomScans.kt
CarlosEsco: https://github.com/tachiyomiorg/tachiyomi/blob/88619145d816fdd3248b5c071bdd8059ab3219cb/app/src/main/java/eu/kanade/tachiyomi/data/library/LibraryUpdateService.kt#L258

src/en/randomscans/src/eu/kanade/tachiyomi/extension/en/randomscans/RandomScans.kt
```diff
@@ -0,0 +1,279 @@
...
+            description = "Add this manga from search to fetch details properly"
+        }
+    }
+
+    // the same as SManga.copyFrom, which we can't access, except it doesn't copy status
+    private fun SManga.copyFromCustom(other: SManga): SManga {
```
CarlosEsco, line 185: This seems to only be used when copying for the one shot, why does it not copy the status then?  Isn't other already set to complete? or do i have that backwards

src/en/randomscans/src/eu/kanade/tachiyomi/extension/en/randomscans/RandomScans.kt
lord-ne: I think setting `Dropped` to `Completed`  fits with `updateOnlyNonCompleted`, since we don't need to update a `Dropped` source since we know there are no new chapters

CarlosEsco: Does the status Dropped == They dropped it  or does it == The publication dropped it, but they are scanning those are two different things, also setting it Completed makes it seem to naive users that the manga is actually complete scanlated

src/en/randomscans/src/eu/kanade/tachiyomi/extension/en/randomscans/RandomScans.kt
lord-ne: If the oneshot was added from a regular search (as opposed to a deeplink slug search), the status might be `Unknown` if the extension wasn't able to parse it from the website (not every series has a status listed on the page, which is why I use safe calls when reading the status so `null` (element not found) can fall into the `else` case and default to `Unknown`)

src/en/randomscans/src/eu/kanade/tachiyomi/extension/en/randomscans/RandomScans.kt
lord-ne: It means the scanlators dropped it. I think that since we also set the first line of the description to "Status: Dropped" it won't be confusing to users.

PR #8294, opened 2021-07-27
fix two madara sources

CVIUS:
I just removed some parts that causing page looping.
Also I found that mangalibrary start using base64 encryption for covers and chapters? I'll removed it because it is beyond my knowledge and its just a new source so ig its ok to make it a request. I apologize.


Riztard:
so youre not testing them again before PR

PR #8301, opened 2021-07-28
Update FoolSlide multisrc

ObserverOfTime:
* Refactored the extension code to be more idiomatic.
  * Notably, `latestUpdatesParse` uses [`MutableSet.add`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/-mutable-set/add.html) now which reduces the number of operations.
* Added an adult content preference (`true` by default, forced in Baixar Hentai).
* Added The Cat Scans icons & removed trailing slash from URL.
* Moved `/reader` from Hyakuro URL to `urlModifier`.
* Added Mangatellers icons & updated to the new URL.
   * Removed obsolete method override.
 * Merged HNI-Scantrad French & English URLs.
   * Removed obsolete method overrides.
 * Moved LetItGo Scans from Comicake to FoolSlide.
 
 ---
 
 Initially, I planned to completely rewrite the extension using the [API](https://github.com/FoolCode/FoOlSlide/blob/master/application/controllers/api/reader.php) which, unlike the frontend, is consistent among all sources (except LupiTeam whose comic model has 2 extra fields) and doesn't require regular expressions or complicated date parsing functions. Unfortunately, I ran into several issues, one of which I was unable to solve.
 
 The paginated response has no indication of whether there's another page and can only be stopped by returning an empty `MangasPage` when it 404s. However, Tachiyomi shows "No results found" in that case and I haven't found a way to forcefully terminate the observable without throwing an exception which would display an error message.


CarlosEsco:
> The paginated response has no indication of whether there's another page and can only be stopped by returning an empty MangasPage when it 404s. However, Tachiyomi shows "No results found" in that case and I haven't found a way to forcefully terminate the observable without throwing an exception which would display an error message.

@arkon do you have an idea about this

arkon:
Rough ideas off the top of my head:

- Throw an exception with a more descriptive message (e.g. "No more pages")
- Actually make a request for the next page to determine if it exists for `hasNextPage` (kind of wasteful though, unless you also reuse the response for the actual next page?)

ObserverOfTime:
I think keeping the current implementation is the best option…

multisrc/src/main/java/eu/kanade/tachiyomi/multisrc/foolslide/FoolSlide.kt
```diff
@@ -13,79 +19,75 @@ import kotlinx.serialization.json.jsonArray
...
+) : ConfigurableSource, ParsedHttpSource() {
 
     override val supportsLatest = true
 
-    private val json: Json by injectLazy()
+    private val json by lazy { Injekt.get<Json>() }
+
+    private val userAgent = "Mozilla/5.0 (" +
+        "Android ${Build.VERSION.RELEASE}; Mobile) " +
+        "Tachiyomi/${BuildConfig.VERSION_NAME}"
```
CarlosEsco, line 48: Why are we sending a custom Tachiyomi user agent now?

multisrc/src/main/java/eu/kanade/tachiyomi/multisrc/foolslide/FoolSlide.kt
```diff
@@ -13,79 +19,75 @@ import kotlinx.serialization.json.jsonArray
...
+        )
     }
 
     override fun latestUpdatesSelector() = "div.group"
 
     override fun latestUpdatesRequest(page: Int): Request {
-        if (page == 1) {
-            latestUpdatesUrls.clear()
-        }
+        if (page == 1) latestUpdatesUrls.clear()
```
CarlosEsco, line 72: this goes against the kotlin style format

multisrc/src/main/java/eu/kanade/tachiyomi/multisrc/foolslide/FoolSlideGenerator.kt
```diff
@@ -25,15 +25,15 @@ class FoolSlideGenerator : ThemeSourceGenerator {
...
+        SingleLang("Mangatellers", "https://reader.mangatellers.gr", "en"),
         SingleLang("Iskultrip Scans", "https://maryfaye.net", "en"),
         SingleLang("Anata no Motokare", "https://motokare.xyz", "en", className = "AnataNoMotokare"),
         SingleLang("Yuri-ism", "https://www.yuri-ism.net", "en", className = "YuriIsm"),
         SingleLang("LupiTeam", "https://lupiteam.net", "it"),
         SingleLang("Zandy no Fansub", "https://zandynofansub.aishiteru.org", "en"),
         SingleLang("Kirishima Fansub", "https://www.kirishimafansub.net", "es"),
         SingleLang("Baixar Hentai", "https://leitura.baixarhentai.net", "pt-BR", isNsfw = true, overrideVersionCode = 3),
-        MultiLang("HNI-Scantrad", "https://hni-scantrad.com", listOf("fr", "en"), className = "HNIScantradFactory", pkgName = "hniscantrad", overrideVersionCode = 1),
+        SingleLang("HNI-Scantrad", "https://hni-scantrad.com", "fr", className = "HNIScantradFR", pkgName = "hniscantrad", overrideVersionCode = 1),
```
CarlosEsco, line 36: They still seem to have english?
https://hni-scantrad.com/lel/series/hajime-no-ippo/  are all english chapters

multisrc/src/main/java/eu/kanade/tachiyomi/multisrc/foolslide/FoolSlide.kt
ObserverOfTime: Why not? ¯\\\_(ツ)_/¯

multisrc/src/main/java/eu/kanade/tachiyomi/multisrc/foolslide/FoolSlide.kt
ObserverOfTime: https://developer.android.com/kotlin/style-guide#braces
> Braces are not required for `when` branches and `if` statement bodies which have no `else if/else` branches and which fit on a single line.

multisrc/src/main/java/eu/kanade/tachiyomi/multisrc/foolslide/FoolSlide.kt
CarlosEsco: No reason to send it, unless asked for by a scanlator.  This just provides something that an aggregator can block on

multisrc/src/main/java/eu/kanade/tachiyomi/multisrc/foolslide/FoolSlideGenerator.kt
ObserverOfTime: So I guess they just merged both languages with no distinction between them…

multisrc/src/main/java/eu/kanade/tachiyomi/multisrc/foolslide/FoolSlide.kt
CarlosEsco: yes yes i was misremembering it with 
```
if (string.isEmpty())
    return  // WRONG!
```

multisrc/src/main/java/eu/kanade/tachiyomi/multisrc/foolslide/FoolSlideGenerator.kt
Riztard: maybe just filter the chapter, if url contains (en-us/fr) similar to mangadex
url example:
	https://hni-scantrad.com/lel/read/hajime-no-ippo/en-us/132/1348/
	https://hni-scantrad.com/lel/read/one-punch-man-version-one/fr/0/30/

PR #8342, opened 2021-07-31
manga alarab:Add filters

ahmedgamal17:
add all filter
but search won't work with filters as the website does not support it


zeredouni:
Thanks .Can you fix these  #8338

PR #8361, opened 2021-08-01
add Tachidesk extension

AriaMoradi:
This is the most basic possible implementation of a Tachidesk client.


src/all/tachidesk/src/eu/kanade/tachiyomi/extension/all/tachidesk/Dto.kt
```diff
@@ -0,0 +1,79 @@
+package eu.kanade.tachiyomi.extension.all.tachidesk
+
+enum class MangaStatus(val status: Int) {
```
ObserverOfTime, line 5: You can use the [`ordinal`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/-enum/ordinal.html) property instead of a custom status property and replace `MangaStatus.valueOf(value)` with `MangaStatus.values()[value]`.

AriaMoradi: It conforms to Tachidesk's internal data classes, I have to convert Strings like "UNKNOWN" to it's ordinal value defined in Tachiyomi's manga status, do you have a better way?

ObserverOfTime: I just realised that you're not actually using your `valueOf(Int)`, but the built-in `valueOf(String)`.

In that case, you can just remove your custom method, along with the `status` property, and use `MangaStatus.valueOf(status).ordinal` instead.

src/all/tachidesk/src/eu/kanade/tachiyomi/extension/all/tachidesk/Tachidesk.kt
```diff
@@ -0,0 +1,184 @@
...
+    fun pageListParse(response: Response, chapter: SChapter): List<Page> {
+        val mangaId = chapter.url.split(" ").first()
+        val chapterIndex = chapter.url.split(" ").last()
+
+        val chapter = json.decodeFromString<ChapterDataClass>(response.body!!.string())
+
+        return (0 until chapter.pageCount).map {
+            val url = "$baseUrl/api/v1/manga/$mangaId/chapter/$chapterIndex/page/$it/"
+            Page(it, url, url)
+        }
```
ObserverOfTime, line 103:
```suggestion
        return List(chapter.pageCount) {
            Page(it + 1, "", "$baseUrl/api/v1/manga/$mangaId/chapter/$chapterIndex/page/$it/")
        }
```

src/all/tachidesk/src/eu/kanade/tachiyomi/extension/all/tachidesk/Dto.kt
CarlosEsco: To add to the above.
"ONGOING" -> SManga.Ongoing

src/all/tachidesk/src/eu/kanade/tachiyomi/extension/all/tachidesk/Tachidesk.kt
```diff
@@ -0,0 +1,190 @@
...
+
+    private fun MangaDataClass.toSManga() = SManga.create().also {
+        it.title = title
+        it.url = id.toString()
+        it.thumbnail_url = "$baseUrl$thumbnailUrl"
+        it.artist = artist
+        it.author = author
+        it.description = description
+        it.status = when (status) {
+            "UNKNOWN" -> 0
```
alessandrojean, line 176:
```suggestion
            "UNKNOWN" -> SManga.UNKNOWN
```

Same with the other ones below this line.

alessandrojean (changes requested):

src/all/tachidesk/src/eu/kanade/tachiyomi/extension/all/tachidesk/Tachidesk.kt
AriaMoradi: Tachidesk supports the same exact 4 possible status, else can only represent an impossible error

AriaMoradi: Using `SManga` constants makes sense, I'll change that bit.

CarlosEsco: I think what he means is remove the 0 branch.  And make it just else at the end. So if tachidesk ever adds different statuses it will fall back to unknown besides going to else when 0

AriaMoradi: I another one is ever added, it'll be mirroring the same thing in Tachiyomi

src/all/tachidesk/src/eu/kanade/tachiyomi/extension/all/tachidesk/Tachidesk.kt
```diff
@@ -0,0 +1,190 @@
...
+                }
+            }
+
+            setOnPreferenceChangeListener { _, newValue ->
+                try {
+                    val res = preferences.edit().putString(title, newValue as String).commit()
+                    Toast.makeText(context, "Restart Tachiyomi to apply new setting.", Toast.LENGTH_LONG).show()
+                    res
+                } catch (e: Exception) {
+                    e.printStackTrace()
```
CarlosEsco, line 132: Show log the exception here not print stacktrace

CarlosEsco (changes requested):

src/all/tachidesk/src/eu/kanade/tachiyomi/extension/all/tachidesk/Tachidesk.kt
CarlosEsco: Defensive programming. The manga shouldn't fail to load, just because it doesn't know the status.  Just because you plan at this time to only match Tachiyomi status, doesn't mean someone else who forks tachi desk will, or you won't change your mind down the line.  There is no reason to throw an exception here.

AriaMoradi: alright, makes sense

src/all/tachidesk/src/eu/kanade/tachiyomi/extension/all/tachidesk/Tachidesk.kt
AriaMoradi: Hmm, took it from Komga btw

CarlosEsco: Must have missed that.  But Log.e will print to the android console

AriaMoradi: done

src/all/tachidesk/src/eu/kanade/tachiyomi/extension/all/tachidesk/Tachidesk.kt
```diff
@@ -0,0 +1,181 @@
...
+
+    private fun MangaDataClass.toSManga() = SManga.create().also {
+        it.title = title
+        it.url = id.toString()
+        it.thumbnail_url = "$baseUrl$thumbnailUrl"
+        it.artist = artist
+        it.author = author
+        it.description = description
+        it.status = when (status) {
+            "UNKNOWN" -> SManga.UNKNOWN
```
CarlosEsco, line 163: now that you have else -> SManga.UNKNOWN  you dont need this check

CarlosEsco (approved):

PR #8429, opened 2021-08-07
add new sources to madara

CVIUS:
added hm2d, manga68.net, mangamoli, treemanga


Riztard:
just want to make sure, youve tested them before open pr right?

Riztard:
it should show/ make it easier to know which part is wrong with android studio

PR #8445, opened 2021-08-07
Change latest API endpoint in Bilibili

alessandrojean:
Should be good now, @FourTOne5.


FourTOne5:
Dto Cleanup? Andit should be good ☺️

alessandrojean:
Ah, sure, thanks for remembering :+1:

PR #8469, opened 2021-08-08
Remove status from MY

alessandrojean:
Website is marking everything as complete, so the status they provide isn't reliable any more.


alessandrojean:
Closing since the source can start to fixing it sometime in the future.

PR #8503, opened 2021-08-10
Implementing features for 3 sources

### LINKED ISSUE #1262: [Mangadex] Manually select default sort for catalogue
I would really appreciate if I could select the default sort from:   
Extension>Mangadex page.

I like to use Update time as sorting method as it allows me to find new/active manga but I find it a bit irritating to every time have to select the sort.

### LINKED ISSUE #7779: Manga-Fénix: Rate limit needed
They have to migrate many times now because of the heavy use of the site, it's probably they will ask for the source to be removed. So in order to have a less impact on the website. I recommend put a discrete limited for the scan of mangas and chapters loads from the website.

### LINKED ISSUE #8468: Mangahasu: Use popular section instead of the directory section
### Acknowledgements

- [X] I have updated the app to version **[0.12.1](https://github.com/tachiyomiorg/tachiyomi/releases/latest)**.

Mangahasu uses the directory section https://mangahasu.se/directory.html
instead of popular section https://mangahasu.se/most-popular.html
Screenshot_20210808-223614_Tachiyomi

### Other details

_No response_

Arraiment:
**Mangahasu** will use most popular of all time instead of directory, which closes #8468. Also, chapter details will no longer display if they are updating.
Adds default popular page sort order preference to **Mangadex**, which closes #1262.
Closes #7779 for **Manga-Fenix**


CarlosEsco:
Please split this into multiple prs (1 per source), only multisrc prs should really address multiple sources

src/all/mangadex/src/eu/kanade/tachiyomi/extension/all/mangadex/MangaDexHelper.kt
```diff
@@ -201,7 +201,7 @@ class MangaDexHelper() {
             val nonGenres = listOf(
                 (attr.publicationDemographic ?: "").capitalize(Locale.US),
                 contentRating,
-                Locale(attr.originalLanguage ?: "").displayLanguage
+                Locale(attr.originalLanguage).displayLanguage
```
CarlosEsco, line 204: Are you sure its not possible to have null here?

Arraiment: IDE tells me it shouldn't result in anything but a string. Regardless, I'll just leave the mangadex commits out of it all together.

PR #8645, opened 2021-08-18
Replace org.json with kotlinx.serialization in PizzaReader

alessandrojean:
Addresses #7392.

The PR also adds the missing Latest section.

If you want to review, @FedericoHeichou, feel free to do it.


FedericoHeichou:
Wait, really? I don't know how to thank you, my contributed to not explode my old pc debugging things I never did :smile: 

You added latest updates too.

Thank you so much again, now I only need to add:
`.sortedByDescending { comic -> comic.views }` in popular (or something similar, I have to check the right element) and the extension is done

PR #8735, opened 2021-08-22
Tachidesk: fix crash issues

AriaMoradi:


Riztard:
maybe later need to make some announcement in tachidesk server to update manual/reinstall ext.

AriaMoradi:
> maybe later need to make some announcement in tachidesk server to update manual/reinstall ext.

It's fine, I forgot to bump version on the last PR, only a handful of users should be effected....

PR #8740, opened 2021-08-22
add linting to multisrc

AriaMoradi:


multisrc/build.gradle.kts
```diff
@@ -6,6 +6,17 @@ plugins {
     kotlin("android")
 }
 
+buildscript {
```
arkon, line 9: Is this needed? It's already in the top-level build.gradle.

AriaMoradi: worked wthout

PR #8819, opened 2021-08-24
Reaper scans: fix upload date bleeding into chapter name

AriaMoradi:
before:
image
after:
image


multisrc/overrides/madara/reaperscans/src/ReaperScans.kt
```diff
@@ -1,9 +1,31 @@
...
+        with(element) {
+            select(chapterUrlSelector).first()?.let { urlElement ->
+                chapter.url = urlElement.attr("abs:href").let {
+                    it.substringBefore("?style=paged") + if (!it.endsWith(chapterUrlSuffix)) chapterUrlSuffix else ""
+                }
+                chapter.name = urlElement.ownText()
+            }
+            // Dates can be part of a "new" graphic or plain text
+            // Added "title" alternative
+            chapter.date_upload = select("img").firstOrNull()?.attr("alt")?.let { parseRelativeDate(it) }
```
alessandrojean, line 24: Need to make the `parseRelativeDate` as `protected` or `protected open` in the base Madara class.

alessandrojean (changes requested):

multisrc/src/main/java/eu/kanade/tachiyomi/multisrc/madara/Madara.kt
```diff
@@ -542,7 +542,7 @@ abstract class Madara(
 
     // Parses dates in this form:
     // 21 horas ago
-    private fun parseRelativeDate(date: String): Long {
+    fun parseRelativeDate(date: String): Long {
```
alessandrojean, line 545: Removing the `private` will still make the method as `private` by default. 

It needs to be `protected` or `protected open` to allow its access in the child classes.

alessandrojean (changes requested):

Issue #4564, opened 2020-10-09
[Source Request] CommitStrip: The blog relating the daily life of web agency developers


Noriller:
Name:
CommitStrip

Small description:
The blog relating the daily life of web agency developers

Link:
https://www.commitstrip.com/

Thank you

AbdullahM0hamed:
Should have the source request label instead of enhancement.

kbhasi:
Add info:

*Languages: English and French

PR #8878, opened 2021-08-27
Add extension CommitStrip

curche:
Closes https://github.com/tachiyomiorg/tachiyomi-extensions/issues/4564

Inspired by RealLifeComics, PatchFriday, Dilbert and other extensions

Still a work in progress hence the Draft
A Checklist: 
- [x] Overall basic ext
- [x] FactorySource : supports `en` and `fr` and so, I put it under `all`
- [x] Manga thumbnails : tried [this](https://www.commitstrip.com/wp-content/themes/krds_blog/images/logo_small.png) and [this](https://www.commitstrip.com/wp-content/themes/krds_blog/images/logo_small.png), but didn't seem to have worked. Not sure why :thinkiiing:
- [x] Extension icon
- [x] **_Pagination_**


ObserverOfTime:
Why not remove `siteLang` if it's always the same as `lang`?

curche:
> Why not remove `siteLang` if it's always the same as `lang`?

Actually made a copy of MangadexFactory and just tweaked it only enough to suit my needs. 
Looking at it, the way dex (and batoto) has it is: `lang` is for the tachi side of things and `siteLang` is for the website side of things; I believe this is useful in cases where the language code that tachi uses is different ( "pt-BR" and "pt_br", "es-419" and "es_419"). 
Thought I'll keep it as is if at all CommitStrip adds a new lang like that (unlikely but the ext can support easily :D).

Maybe have it like BatotoFactory i.e. `val lang: String, val siteLang: String = lang`? I'll leave it up to the review.

> For the icon, you could use the one from their [social media](https://pbs.twimg.com/profile_images/799176331623800832/ggd6_JbJ_400x400.jpg).

*If only I saw this earlier*. I ended up using their [smaller logo](https://www.commitstrip.com/wp-content/themes/krds_blog/images/logo_small.png) which I think is alright and it shows prominent characters as well

curche:
PR ready for review

Notes:
- loading the pages is slow and so the chapter list might take a while to load. Tried a different approach but in vain. If there's any other idea or any other ext that does pages faster, I can try again.
- custom thumbnails that I uploaded onto imgur 
  | EN | FR |
  |------|------|
  |  https://i.imgur.com/HODJlt9.jpg     |    https://i.imgur.com/I7ps9zS.jpg   |
  
- lets hope this doesn't end up like this (jk XD)

Strip-Hackers-VS-Security-team-650-finalenglish1

Singest:
The webview  and reading the comics in French doesn't work for me. The French comics just show up in English, and the webview URL has the base URL in front of the actual URL.

curche:
I was aware of webview not working when I checked after the PR was merged. I hadn't figured out how to fix it yet but guess I'll try something.

As for the other issue, can reproduce. I messed the lang when I reworked it. @Singest please create a new issue for it. I'll send a fix once I can

Singest:
#8910

PR #8888, opened 2021-08-28
Filtered comikey chapters

### LINKED ISSUE #8877: MangaDex: Chapters Uploaded by Comikey return No Pages found
### Source information

1.2.129

### Steps to reproduce

Find a manga with a Comikey chapter
Open it

### Expected behavior

The chapter should be able to be opened, or if it's a redirect to another site, the chapter should be hidden

### Actual behavior

The chapter returns no pages found

### Tachiyomi version

0.12.1

### Android version

Android 9

### Other details

_No response_

### Acknowledgements

- [X] I have updated the app to version **[0.12.1](https://github.com/tachiyomiorg/tachiyomi/releases/latest)**.

loocool2:
Closes #8877

### DIFF
--- a/src/all/mangadex/build.gradle
+++ b/src/all/mangadex/build.gradle
@@ -6,7 +6,7 @@ ext {
     extName = 'MangaDex'
     pkgNameSuffix = 'all.mangadex'
     extClass = '.MangaDexFactory'
-    extVersionCode = 129
+    extVersionCode = 130
     libVersion = '1.2'
     containsNsfw = true
 }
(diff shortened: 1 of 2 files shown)


paulo27ms:
Bilibili Comics will also be external-only from what's planned, if you want to add that.

alessandrojean:
> Bilibili Comics will also be external-only from what's planned, if you want to add that (for now they have pages in the API but in the future they won't I guess).

Bilibili is a special case because the Bilibili team that is uploading stuff to Dex, it's kinda like a deal between them iirc.

paulo27ms:
I'm pointing it out because the original issue was about chapters not loading pages, in the future I'd imagine Bilibili's stuff won't have pages (I'm not sure why they have them now in the API since you can't read them on the site) but I guess that can wait until they really start uploading page-less chapters.

CarlosEsco:
Billibilli will eventually have the latest 3 chapters redirect and older will be available on dex

CarlosEsco:
Also comikey might be under the same boat

paulo27ms:
Which is why something like I mentioned in https://github.com/tachiyomiorg/tachiyomi-extensions/issues/8877 would be better rather filtering out the entire group. Only filter stuff if there's no pages in the API, not hardcoded group filters.

CarlosEsco:
Confirmed that this will all redirect

PR #8896, opened 2021-08-30
[LHTranslation] Migrate from FMReader to Madara

### LINKED ISSUE #8727: LHTranslation - Madara
### Acknowledgements

### Source name

LHTranslation

### Source link

https://lhtranslation.net

### Language

English

### Other details

Source Theme: Madara

New temporary site which is being used due to the recent hacking incident of LHTranslation. Will close #8724 (Needs to use different name than LHTranslation due to a extension already existing in that name. "LHTranslation - Madara" should work.)

FlaminSarge:
Closes #8727


FlaminSarge:
@FourTOne5 dyou know if they're staying on Madara or if they're going back to FM eventually?

FourTOne5:
IMG_20210830_123622_388

This is their last public statement. So I think it'll be better to just make a separate Madara extension.

FourTOne5:
@FlaminSarge are you working on it? I'll take over if you aren't.

FlaminSarge:
What's the timeline for them getting the old site back up? If it's within 1-2 weeks, then it's not worth doing either of these. If it's longer, then it may be worth migrating entirely as the old extension won't function at all for that amount of time anyways; if they eventually ever do return to FMReader, the inverse of this PR can be opened at that time.

To me, it doesn't make sense to try to preserve the broken extension while also creating a fixed one; that's what git history is for.

FourTOne5:
> 
> 
> What's the timeline for them getting the old site back up? If it's within 1-2 weeks, then it's not worth doing either of these. If it's longer, then it may be worth migrating entirely as the old extension won't function at all for that amount of time anyways; if they eventually ever do return to FMReader, the inverse of this PR can be opened at that time.
> 

There is no specific statement they only said that there working on the old site.
> To me, it doesn't make sense to try to preserve the broken extension while also creating a fixed one; that's what git history is for.


Yeah I was being dumb xD. So I guess PR is ready to be merged.

PR #8901, opened 2021-08-30
Fix LectorManga loading pages, add rate limit for new image CDN hosts

### LINKED ISSUE #8859: Lector Manga: chapter images do not load
Tachiyomi version 0 .12 .1 (67)

Android  version 11

device redmi note 5+

Lectormanga versión 1. 2 .19

to reproduce the issue: 
open any manga from lectormanga you will get a black screen with the word retry 
expected behavior: load the manga successfully

additional information: 

i checked lectormanga website is online and working well, i checked the web view feature of tachiyomi and also is working well for lectormanga, i search in the issue list and i can't find nothing about this, two pull requests were done by @clouddark75 and then merge  regarding to this #8843 and this #8851 but after the update of the extension two times the issue still persist

edgarMejia:
Closes #8859


Clouddark75:
I was thinking of "something" like that because the extension did need some rework. They did add a couple of new server mask for the images, but they're just a mask for img1.followmanga.com or img1.japanreader.com (I was thinking of replacing them with the original images URL domain like an option, but my knowledge it isn't that good for that, haha), then there was the problem that some of the recent Mangas chapters did have a different order for the data-src image location so it will break the older chapters if I tried to fixed it that way. 


**Note**: One more thing, there's another mask server "https://img1.recipesandcook.com" it loads on some type of mangas. 


Thanks for fix, Edgar!

PR #8980, opened 2021-09-06
TMO ratelimit options

TheeoCornnaro:
list of values for ratelimit is not enough,  chapter ratelimit is 10, but image ratelimit is higher.

added more values for the user to choose the right ratelimit, this variates from image hosting server. 

it is annoying to be reading and then stop for a minute on the images because the choices were below the server cap.


alessandrojean (approved):

arkon:
Is there really a point of a rate limit if you can set it to 999...?

TheeoCornnaro:
> Is there really a point of a rate limit if you can set it to 999...?

not really, it was a way to set it to disabled. that is more for the images rate limit, in the chapter ratelimit, 10 is the number, but in the image rate limit, I guess it depends on the chapter number of images, I reduced it to 100 as maximum.

CarlosEsco:
fix merge conflict please

TheeoCornnaro:
> fix merge conflict please

done

PR #9012, opened 2021-09-08
MegaTokyo: Fix Chapter Order

loocool2:
Reverses chapter order so that the first chapter isn't the latest one anymore

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,12 +1,9 @@
-<!--
-If you are updating extensions, please remember to:
+Checklist:
 
-- Update the `extVersionCode` value in `build.gradle`
-- Annotate `Source` or `SourceFactory` classes with `@Nsfw` when appropriate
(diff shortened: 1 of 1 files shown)


src/en/megatokyo/src/eu/kanade/tachiyomi/extension/en/megatokyo/Megatokyo.kt
```diff
@@ -19,7 +20,8 @@ import javax.net.ssl.SSLContext
 import javax.net.ssl.TrustManager
 import javax.net.ssl.X509TrustManager
 
-class Megatokyo : ParsedHttpSource() {
+class
+Megatokyo : ParsedHttpSource() {
```
arkon, line 24: What happened here?

loocool2: Accidentally pressed enter there I guess

PR #8965, opened 2021-09-04
Fix chapters and other updates for Mangadex

Arraiment:
Closes #8928 
As for the updated icon, I couldn't find one with a high enough res so I'll leave that up to someone else.
Closes #8872 
Closes #8036, now ext will show titles based on preferred lang->en->first available title, any title is better than no title.
Also changed latest updates since api now supports sorting by update date.

PR #9043, opened 2021-09-12
Mangadex: add missing sort options

jopejoe1:
Added sort options that where missing in extension
~~Updated `latestUpdatesRequest` to use `latestUploadedChapter` sort filter instead of the current work around.~~
No longer updating latestUpdateRequest in this PR

Checklist:

- [x] Have explicitly kept the `versionId` if a source's name or language were changed


FourTOne5:
In #8965 Carlos said:

> Other issue, I'm not merging that latest change
> Sorting by latest chapter on manga list doesn't take into account the current users language (not available in api). So if some group uploads a bunch chapter in Spanish for manga 1 and you are browsing the English dex extension you are seeing the manga 1 near the top even though no English chapters were uploaded

jopejoe1:
> In #8965 Carlos said:
> 
> > Other issue, I'm not merging that latest change
> > Sorting by latest chapter on manga list doesn't take into account the current users language (not available in api). So if some group uploads a bunch chapter in Spanish for manga 1 and you are browsing the English dex extension you are seeing the manga 1 near the top even though no English chapters were uploaded

One thing I'm not seeing mentioned in that PR is that the current implementation has a lot of repeats. Which i find more annoying than having updates from another language in there. The old mangadex also had updates from other languages in its latest update tab.

Comparison:
Current

https://user-images.githubusercontent.com/34899572/132995711-e9fe9c19-c117-4d44-9ea2-454c36e64790.mp4



Updated 

https://user-images.githubusercontent.com/34899572/132995648-db306028-9f7f-496c-a9cb-4da38d5246b3.mp4

FourTOne5:
How about some kind of caching system?

It'll store manga id and chapter update time as a pair (or something else) if a repetation comes for the manga id and the chapter update time is higher it'll swap with the higher one.

Now if the cache has item from 1:00am - 1:30am and if the time is like 5 minutes more( or some user preferred value maybe?) it'll start to add item to cache until the lowest time is less than 1:30am

Though there will also need to be a way to bakfill the cache (which I have no idea how to do)

CarlosEsco:
> 
> 
> > In #8965 Carlos said:
> > > Other issue, I'm not merging that latest change
> > > Sorting by latest chapter on manga list doesn't take into account the current users language (not available in api). So if some group uploads a bunch chapter in Spanish for manga 1 and you are browsing the English dex extension you are seeing the manga 1 near the top even though no English chapters were uploaded
> 
> One thing I'm not seeing mentioned in that PR is that the current implementation has a lot of repeats. Which i find more annoying than having updates from another language in there. The old mangadex also had updates from other languages in its latest update tab.
> 
> Comparison:
> Current
> Current.mangadex.latest.1.mp4
> 
> Updated
> Updated.mangadex.1.mp4

It's only not annoying cause billibilli is slamming uploads. When they finally stop your implementation won't be similar at all.  Rather you do a set with the manga id and remove it before the query to remove the duplicates

jopejoe1:
> > > In #8965 Carlos said:
> > > > Other issue, I'm not merging that latest change
> > > > Sorting by latest chapter on manga list doesn't take into account the current users language (not available in api). So if some group uploads a bunch chapter in Spanish for manga 1 and you are browsing the English dex extension you are seeing the manga 1 near the top even though no English chapters were uploaded
> > 
> > 
> > One thing I'm not seeing mentioned in that PR is that the current implementation has a lot of repeats. Which i find more annoying than having updates from another language in there. The old mangadex also had updates from other languages in its latest update tab.
> > Comparison:
> > Current
> > Current.mangadex.latest.1.mp4
> > Updated
> > Updated.mangadex.1.mp4
> 
> It's only not annoying cause billibilli is slamming uploads in english. When they finally stop your implementation won't be similar at all. Rather you do a set with the manga id and remove it before the query to remove the duplicates

Removed it from this PR

PR #9068, opened 2021-09-15
Remove Random Scans source

lord-ne:
They moved to a new website which would require a complete rewrite of the extension


FourTOne5:
Someone will rewrite the existing extension there is no point in removing it.

ObserverOfTime:
The [new website](https://randomscans.xyz/) uses WPMangaStream so removing the current extension is the way to go.
Just add the site to the generator, move the assets to the overrides, and update the manifest.

PR #9085, opened 2021-09-16
[zh-copymanga]Fix an issue that the "change CDN" option was incorrectly handled

SnowArk:
Fix an issue that the "change CDN" option was incorrectly handled. It would use the china mainland CDN if you didn't choose the  change cdn option.

Checklist:

- [ ] Annotated `Source` or `SourceFactory` classes with `@Nsfw` when appropriate
- [ ] Added the `containsNsfw = true` flag in `build.gradle` when appropriate
- [x] Have explicitly kept the `versionId` if a source's name or language were changed


zhongfly:
You misunderstood this option, or the explanation for this option is misleading.This option refers to changing the cdn of the picture to mainland, but in fact the original link obtained by the software is originally from the mainland. Therefore, when the option is true, the link should not be modified, and when it is false, it should be changed to overseas cdn starting with `mirror` or `mirror2`.

zhongfly:
So, we should modify this option now to make it in line with the actual effect: "Convert image CDN to overseas CDN (to load the image using an overseas CDN, it is recommended to turn on this option if you can use a proxy)/转换图片cdn为境外cdn(加载图片使用境外CDN,可以使用代理的情况下推荐打开此选项)". When this option is turned on, as you now, modify the cdn to the overseas cdn like `mirror.mangafunc.fun`.
`mirror.mangafunc.fun` is not a cdn for mainland!

SnowArk:
> You misunderstood this option, or the explanation for this option is misleading.This option refers to changing the cdn of the picture to mainland, but in fact the original link obtained by the software is originally from the mainland. Therefore, when the option is true, the link should not be modified, and when it is false, it should be changed to overseas cdn starting with `mirror` or `mirror2`.

OK seems I misunderstood. I'm now totally understand what you mean. The reason why I changed this is because sometime the overseas cdn will have some delay and will cause the replaced link not be able to load the picture. I think it is better to use the original link by default. Maybe I can set this option to true by default. What do you think?

SnowArk:
> So, we should modify this option now to make it in line with the actual effect: "Convert image CDN to overseas CDN (to load the image using an overseas CDN, it is recommended to turn on this option if you can use a proxy)/转换图片cdn为境外cdn(加载图片使用境外CDN,可以使用代理的情况下推荐打开此选项)". When this option is turned on, as you now, modify the cdn to the overseas cdn like `mirror.mangafunc.fun`.
> `mirror.mangafunc.fun` is an overseas CDN!

It's OK to me! It is better to include the possible issue that the overseas cdn might not be able to load the picture if the manga is newly updated in the description.

zhongfly:
> > So, we should modify this option now to make it in line with the actual effect: "Convert image CDN to overseas CDN (to load the image using an overseas CDN, it is recommended to turn on this option if you can use a proxy)/转换图片cdn为境外cdn(加载图片使用境外CDN,可以使用代理的情况下推荐打开此选项)". When this option is turned on, as you now, modify the cdn to the overseas cdn like `mirror.mangafunc.fun`.
> > `mirror.mangafunc.fun` is an overseas CDN!
> 
> It's OK to me! It is better to include the possible issue that the overseas cdn might not be able to load the picture if the manga is newly updated in the description.

Sure, waiting for your pr.

PR #9241, opened 2021-09-26
Add Kouhai Scanlations

### LINKED ISSUE #9194: Kouhai Scanlations
### Source name

Kouhai Scanlations

### Source link

https://kouhai.work

### Language

English

### Other details

_No response_

### Acknowledgements

ObserverOfTime:
I ran into a weird issue and I don't know if I'm doing something wrong or if there's a bug in the app.

The source has 4 filters:

- Genres *(checkbox list)*
- Themes *(checkbox list)*
- Demographics *(select)*
- Status *(select)*

Whenever demographics and/or status are set, everything seems to work as expected.

However, if I only set genres and/or themes, Tachiyomi calls `popularMangaRequest` instead.

Furthermore, if I search for a query while setting genres and/or themes, but not demographics or status,
it fails with `NoSuchElementException` which means that the `FilterList` is missing one or more of the filters.

---

Closes #9194

Checklist:


alessandrojean:
I've run with this issue once, the problem if I remember correctly was related to instantiating the list inside the class.

One solution that worked for me is creating a method that return the list with each filter item that you're having problem (like `getGenreList()`) inside the source class and add it as a parameter in the filter constructor, so you can pass it in the `getFilterList()` as `GenreFilter(getFilterList())`.

PR #9242, opened 2021-09-26
[zh-copymanga] Fix the Library Updates only show the last chapter of updated chapters

SnowArk:
Fix the Library Updates only show the last chapter of updated chapters if there are several chapters is updated.
If only set the `date_upload` for the last chapter, the other updated chapter will be ignored in Library Updates.
Set the time for all chapters will make the time showed in the detail page not accurate. But it can make the Library Updates function works properly.

Checklist:

- [ ] Annotated `Source` or `SourceFactory` classes with `@Nsfw` when appropriate
- [ ] Added the `containsNsfw = true` flag in `build.gradle` when appropriate


sunbeams001:
Wait, I don't think it's a proper fix, if you don't know the certain update time of a chapter, it should be left blank. and it's compatible with other extensions.
If the Library Updates function works not well, it should be fixed or redesigned.

SnowArk:
> Wait, I don't think it's a proper fix, if you don't know the certain update time of a chapter, it should be left blank. and it's compatible with other extensions.
> If the Library Updates function works not well, it should be fixed or redesigned.

Unfortunately this issue is raised long long ago [tachiyomi/issues/2089](https://github.com/tachiyomiorg/tachiyomi/issues/2089). So for now in the contributing guide it is recommended to use `System.currentTimeMillis()` as a substitute when real data is not available. I prefer to make the Library Updates function works normally by giving the `date_upload` since many other extensions use similar way to handle this scenario.

PR #9246, opened 2021-09-26
[skip ci] Add a new item to the PR checklist

alessandrojean:
Add a new item asking if the user actually compiled and tested the modifications, since some of them are just editing the files through the web interface of GitHub instead of cloning and following the CONTRIBUTING.md guide.

Checklist:


arkon:
Should be more explicit that it's actually tested by running it. Compilation doesn't mean too much.

alessandrojean:
Sure, you're right.

PR #9306, opened 2021-10-01
Use SimplyHentai API

ObserverOfTime:
This PR converts the SimplyHentai extension to use the API. A self-migration will be required.

Checklist:


src/all/simplyhentai/src/eu/kanade/tachiyomi/extension/all/simplyhentai/SimplyHentai.kt
```diff
@@ -1,221 +1,229 @@
...
-    override val baseUrl = "https://old.simply-hentai.com"
+    override val baseUrl = "https://www.simply-hentai.com"
 
     override val supportsLatest = true
 
-    override val client: OkHttpClient = network.cloudflareClient
+    override val client = network.cloudflareClient
 
-    private val json: Json by injectLazy()
+    override val versionId = 2
```
CarlosEsco, line 35: why override versionId?  Is this new extension not compatible with old entries?

ObserverOfTime: The base URL is different.

PR #9808, opened 2021-11-14
[RU]Newbie new domain

e-shl:
Checklist:


e-shl:
> Have not changed source names

and it is possible to change?

CarlosEsco:
> > Have not changed source names
> 
> and it is possible to change?

Yes but you need to hardcode the source id

e-shl:
> > > Have not changed source names
> > 
> > 
> > and it is possible to change?
> 
> Yes but you need to hardcode the source id
Thanks for the future.
I will leave it that way, because the site in social networks is still under this name.

PR #9853, opened 2021-11-19
ru-comx. Fix chapters parsing from window.__data__

Pavkazzz:
Checklist:


src/ru/comx/src/eu/kanade/tachiyomi/extension/ru/comx/ComX.kt
```diff
@@ -1,5 +1,9 @@
 package eu.kanade.tachiyomi.extension.ru.comx
 
+import com.github.salomonbrys.kotson.get
+import com.github.salomonbrys.kotson.nullArray
+import com.github.salomonbrys.kotson.obj
+import com.google.gson.JsonParser
```
alessandrojean, line 6: We're in the process of removing any usage of Gson and Kotson, please use kotlinx.serialization if possible.

alessandrojean (changes requested):

Issue #4414, opened 2020-09-20
[Source Request] <Japanese sources group>


sohilsayed:
## Source Website
- Name: tonari young jump - sai zen sen - gangan online - seiga nico - urasunday - pocket shonenmagazine - sunday webry - kuragebunch
- Link: https://tonarinoyj.jp/ - https://sai-zen-sen.jp/comics/twi4/ -https://www.ganganonline.com/ - https://seiga.nicovideo.jp/ - https://urasunday.com/ - https://pocket.shonenmagazine.com/ - https://www.sunday-webry.com/ - https://kuragebunch.com/
- Language(s): japanese

## Other Details
A group of japanese oficcial sources beacuse there isnt a lot of oficcial sources in the app and there isnt that many japanese sources

alessandrojean:
Some of these sources are already added as extensions. It would be better for the organization of the issues if you open an individual ticket for the websites you still want to be added.

PR #9928, opened 2021-11-26
Adding Nicovideo Seiga source

ringosham:
Addressing one of the sources listed in #4414
Finally I can ditch the Nicomanga app! Reverse engineering the JavaScript was tough but fun.

Checklist:


ringosham:
I noticed that requests frequently fails due to failure in SSL library during testing. Refreshing makes it work again.
Not sure why that happens, but I don't think that's my fault.

PR #10023, opened 2021-12-07
[Mangadex] Fallback to uploader name for chapters that are missing group info entirely

FlaminSarge:
Checklist:


FlaminSarge:
Verified with the following title:
https://api.mangadex.org/manga/d793e3d0-145c-4f25-b38b-6846b31e87cd/feed?includes[]=scanlation_group&includes[]=user&translatedLanguage[]=en

CarlosEsco:
Only issue i see is there is nothing to distinguish this was the uploader and not the group.  
I rather it say something like 
No Group & Uploaded by: XXXXX
or just the uploaded by section

FlaminSarge:
When it comes to uploads without a group, is their a difference? For all intents and purposes, a no-group upload's scanlator _is_ the uploader, isn't it?

CarlosEsco:
> When it comes to uploads without a group, is their a difference? For all intents and purposes, a no-group upload's scanlator _is_ the uploader, isn't it? (In practice of course it may not necessarily be but for the purpose of distinguishing separate no-group uploads I think this approach makes sense). One thing I'm unsure of is if any username conflicts with a group name (e.g. for DL folder).
> 
I guess my main point is what benefit does showing the uploader even provide?  Seeing Hashu_HS  instead of No Group provides nothing but confusion to the end user that the scanlator field in the app is showing who uploaded but without distinction, and now I am searching for a Hashu_HS scanlation group

> I'm a bit worried about the download folder name if we made the scanlator string include "uploaded by:", though.
This is a valid concern and and can easily be rectified by just having spaces Uploaded by Hashu_HS. As there are plenty of groups that have spaces.

FlaminSarge:
> I guess my main point is what benefit does showing the uploader even provide? Seeing Hashu_HS instead of No Group provides nothing but confusion to the end user that the scanlator field in the app is showing who uploaded but without distinction, and now I am searching for a Hashu_HS scanlation group

I'd argue that having 'No Group' for every single chapter is less useful than having an uploader name (particularly for differentiating duplicate chapters by different uploaders, as in the example manga above), and I also feel it's implicit that the scanlator might as well be the uploader for uploads that lack a group. I do not think anybody's gonna get anywhere searching for 'No Group' over a given username, either. In fact, they'd have more luck searching for e.g. 'Hashu_HS scanlator' or 'Hashu_HS manga' than anything 'No Group' would provide. If the uploader later creates a group and updates their chapters, the ext will fill in the group name properly at refresh anyways, too.

> This is a valid concern and and can easily be rectified by just having spaces Uploaded by Hashu_HS. As there are plenty of groups that have spaces.

Fair.

Seishirou101:
I agree that having all of them all no group wouldn't be very useful but its not really implicit that an uploader is the scanlator. On their discord you have users called power uploaders who upload 500 or more chapters to mangadex and more likely than not, they're not the scanlator of the chapter they upload. So something like what carlos said would probably work with the "No Group & uploaded by" so at least you can differentiate between them and not be too confusing.

FlaminSarge:
I'm leaning towards just `Uploaded by x` given the above info about power uploaders, though I think power uploaders usually properly set a group as well so this may not be relevant for that. Do I need translations/i18n for that string, though?

paulo27ms:
Extensions don't have translations.

FlaminSarge:
Update and added a bit of extra logic to ignore a `no group` group: I've confirmed with MD folks that the group ID I've added as a constant here is their MDv3 "no group" group and can be treated as if it doesn't exist.

PR #10033, opened 2021-12-08
Mangadex: Filter Popular to show only Manga TLd in that particular lang

curche:
While adding #10002, realized Latest/Popular in Mangadex (NE) were showing Manga which just didn't have chapters in NE
Digging a bit, turns out:
  - Turns out Latest isn't broken
  - but Popular could be improved using `availableTranslatedLanguages` as said in [Manga endpoint section](https://api.mangadex.org/docs.html#operation/get-search-manga)

Example api request: [https://api.mangadex.org/manga?availableTranslatedLanguage[]=ne](https://api.mangadex.org/manga?availableTranslatedLanguage[]=ne)
Not sure when this was added to the API, but the earliest #dev-talk-api discussion was back in Sept

Tested on Pixel 3 for Mangadex Nepali, Indonesian, and Russian and it looks to be working as expected

Checklist:


CarlosEsco:
Yea this was added in the last couple months.  I just never got around to applying it

PR #8205, opened 2021-07-21
MangaDex: Make chapter URLs deeplink to series page in Tachiyomi

lord-ne:
Make MangaDex chapter links (i.e. ```https://mangadex.org/chapter/...```) open the corresponding series page in Tachiyomi.

Closes #8066

PR #10060, opened 2021-12-09
Mangadex: View all Manga by a Group (+handle group links)

curche:
While I was hoping for a search for manga by a scanlation group feature, I realised I can at least do the same as in [Feed tab](https://mangadex.org/group/e6615424-f51c-4563-8fc4-cd4b64ce5261/tonari-no-scanlation?tab=feed) of Groups page

- Inspired by https://github.com/tachiyomiorg/tachiyomi-extensions/pull/8205
- Uses the /chapter endpoint's `groups[]` parameter
- Reuses creation of Manga out of Chapters like how Latest section does it
  - Paged; while it does clear duplicates in one page it cannot remove those from next pages (a limitation of the Latest section as well, although its less prevalent there now)
- Adds a search phrase `grp:<group id>` for searching (if you're a madlad and remember group UUIDs) and (for the sane folks) an intent support to open /group links in the App using the extension. Same [example](https://mangadex.org/group/e6615424-f51c-4563-8fc4-cd4b64ce5261/tonari-no-scanlation?tab=feed) as before works. Chose a small keyword like `id:` and `ch:` but I think `group:` is a good alternative too
- Respects `dexLang` so that if ever a particular Group does multi-language TLs, only the one relevant to the extension chosen is shown

Tested on Pixel 3 (blueline) and across the forks. Tested deeplinks on A12 and it works.

Checklist:


CarlosEsco:
I'm gonna let this one sit a little and look over the code more. I was avoiding this because I really hated how hacky the v3 way was. 

As an end user with the v3 I have no idea what group the manga is actually showing up for.  I.e I search grp: dog and it would pick the first scanlator and would have no idea what the group actually was.

curche:
> I'm gonna let this one sit a little and look over the code more.

fair. I'll mark it as Draft for the time being but after adding a few final improvements (down below)

> As an end user with the v3 I have no idea what group the manga is actually showing up for. I.e I search grp: dog and it would pick the first scanlator and would have no idea what the group actually was.

About that:
  - I didn't intend this to be group search feature like we had before (I never used that myself XD). Instead, the extension gets a link to the group's page and shows the feed of that id in the link 
    i.e. link - https://mangadex.org/group/e6615424-f51c-4563-8fc4-cd4b64ce5261/tonari-no-scanlation
          groupID - e6615424-f51c-4563-8fc4-cd4b64ce5261
          and then send a request to /chapter with this groupID
  - a search like `grp:dog` isn't supposed to work since grp: expects a UUID. I'll add a check beforehand so that we do not have to deal with Error 400 responses

I wrote this checking at how website shows it. Like here: https://mangadex.org/group/e6615424-f51c-4563-8fc4-cd4b64ce5261/tonari-no-scanlation?tab=feed
I now realize there's a better way using /manga and `group=` which also lets us use Filters correctly (thanks to @FourTOne5 for making me check again)
 
As a side note, If we need *actual group search*:
  - it's best to be done by /group endpoint (meanwhile I just ~~abuse~~ use the /chapter endpoint again). 
  - Link to that: https://api.mangadex.org/docs.html#operation/get-search-group
  - the /group endpoint accepts searching by name and its what is used in the homepage (omni search). The search might be better now I think. but wouldn't that be something that would be better to have in Neko :thinking:
  - why I say so is because we could have something extra that shows you all the group search results and select one group and then show manga TLd by them
  - if we were to use /group
    - you have a search request like so: https://api.mangadex.org/group?name=tonari
    - the results include: "Tonari no scanlation", "Tonari no Eroneko", etc
    - (select one of the results)
    - then with the group's UUID, get their feed

Lastly, I'll add a few commits before marking this PR as Draft:
  - A check to see if a UUID is being passed
  - Making filters work: I'll rewrite to use /manga instead of /chapter
 
If you feel this is too much, then we can close this and wait for a better way from V5, I don't mind

FourTOne5:
just wondering but we can do something like this `grp:group-id query` right?
Though I guess it's not necessary

CarlosEsco:
group id i am fine with actually.  When i wrote the comment i assumed it was the psuedo group search.

CarlosEsco:
> just wondering but we can do something like this `grp:group-id query` right? Though I guess it's not necessary

what would the query entail?  also @curche if you switch to the manga endpoint are you keeping the deeplinking?

FourTOne5:
> > just wondering but we can do something like this `grp:group-id query` right? Though I guess it's not necessary
> 
> what would the query entail?
it'll search for `query` in that group. i.e. `grp:mangashushu-id solo leveling` it'll search for solo leveling in that Mangashushi's worked manga.
>  also @curche if you switch to the manga endpoint are you keeping the deeplinking?
he is

curche:
> if you switch to the manga endpoint are you keeping the deeplinking?

yes. ~~the deeplinking is the entire reason I am adding this XD~~
Just pushed two commits and tested again (had to reinstall and install to actually see results) and it works the same as before

curche:
> it'll search for query in that group. i.e. grp:mangashushi-id solo leveling it'll search for solo leveling in that Mangashushi's worked manga.

yeah I can see how it's useful. Right now, I don't intend to add it but for groups that actually release a ton of chapters (e.g. the now ded BlackCatScans), it might actually be useful

CarlosEsco:
Is there a reason to use the manga endpoint?  deeplink the group and grp:id is really all that is needed.  Having a filter with a grp:id seems bit excessive and rarely used.

FourTOne5:
> Is there a reason to use the manga endpoint? deeplink the group and grp:id is really all that is needed. Having a filter with a grp:id seems bit excessive and rarely used.

Because it will ~~in most case~~ give more than 1 manga

CarlosEsco:
> > Is there a reason to use the manga endpoint? deeplink the group and grp:id is really all that is needed. Having a filter with a grp:id seems bit excessive and rarely used.
> 
> Because it will ~in most case~ give more than 1 manga

And?  Its not like the view isnt paged, and most scanlators probably do <20 manga.   If you already went into the hassle to copy a group id, or open a group through deeplink.  Does it really matter if you can't filter the results.  

There definitely is no use case for 
grp:randomIdIhadTodeepLink Solo Leveling vs just searching solo leveling.  

grp:randomIdIhadTodeepLink then filtering for say shounen again seems very low usage

curche:
> Is there a reason to use the manga endpoint? deeplink the group and grp:id is really all that is needed. Having a filter with a grp:id seems bit excessive and rarely used.

well, previously I was using the /chapter endpoint, using /manga means I can just create the request and let `searchMangaParse` handle the response like it does already. Having filters is really only an added bonus. If we don't want filters, I can remove filtering for `grp:` (and also add a Note on no filtering while `grp:` searching in Filter sheet)

I was checking out getting the rest of the search after grp:UUID in `grp:UUID stuff` and it seems a bit too much and so, won't add that

CarlosEsco:
@curche let me know when this is good to review

curche:
Not adding any more stuff. PR is ready for review

PR #10221, opened 2021-12-25
Mangadex: Use updated /at-home/server/ endpoint for pages

curche:
Mangadex Dev team's Christmas/New Year's present for its servers :tada: 

This PR does the following (in order of each commit):
- update `Chapter(Attributes)Dto` and move `hash`, `data`, `data-saver` to /at-home/server/ endpoint's `AtHome(Chapter)Dto`
- uses the new `AtHome` response to fetch chapter pages (mostly the same steps as before but with a few additions)
- `pageListRequest` (which creates the request for where we can find the pages) now calls the `/at-home/server` endpoint directly instead of the chapter endpoint while making sure old chapter token practice is still followed
- `pageListParse` (which creates the final url for fetching individual pages from chosen MD@H node) only uses the `AtHomeDto` (no more `chapterDto` required)
  - There is no 204 response as per the API docs (instead there is a 404 response; which is handled by the response caller). 
  - /at-home/server also does not seem to have a 503 response as per API docs but I have kept it anyways (this isn't reproducible at my end but there haven't been issues reported recently so I'll keep it as is for now)
- the original URL (`atHomeRequestUrl), which is used to track tokens, is now obtained using Respone.request.url (the original URL stored in the Response as per [okhttp docs](https://square.github.io/okhttp/4.x/okhttp/okhttp3/-response/request/)). Note: there are issues mentioned with the URL in the okhttp link but as far as I have checked, these will not affect the extension's usecase (building the ext and reading chapters using it is working fine).
- removes additional check for externalUrl uploaders that checks for chapter.attribute.data.isEmpty() - since it no longer exists (Not sure if that'd cause additional problems) 
- increment mangadex.extVersionCode (as per Checklist)

Tested on Tachiyomi Main 0.12.3 (running on A12) and reading chapters/oneshots (instantly and after a while) works as expected. 

Checklist:


src/all/mangadex/src/eu/kanade/tachiyomi/extension/all/mangadex/MangaDex.kt
```diff
@@ -375,40 +375,39 @@ abstract class MangaDex(override val lang: String, val dexLang: String) :
...
+            "${MDConstants.apiUrl}/at-home/server/$chapterId?forcePort443=true"
+        } else {
+            "${MDConstants.apiUrl}/at-home/server/$chapterId"
+        }
+
+        return helper.mdAtHomeRequest(atHomeRequestUrl, headers, CacheControl.FORCE_NETWORK)
     }
 
     override fun pageListParse(response: Response): List<Page> {
         if (response.isSuccessful.not()) {
```
alessandrojean, line 391: 404 is considered a non successful response, so it will throw the exception instead of returning the empty list. Also, it will probably even never enter this method since there's a `asObservableSuccess()` in the `fetchPageList()` method by default, so it's probably needed to override there to handle this too.

curche: Ah so its redundant and dead code now. I had a hunch but I kept it anyways. I'll just remove it completely

curche: Ah wait, I think I didn't understand this fully. Were you also referring to the Error 504 that [getMdAtHomeUrl](https://github.com/tachiyomiorg/tachiyomi-extensions/blob/a478a446a325799a4ac65061680474355f7420bf/src/all/mangadex/src/eu/kanade/tachiyomi/extension/all/mangadex/MangaDexHelper.kt#L159) handles?

alessandrojean: No, because in that case the response is coming directly from a client call above, so the check in this method is valid and works if needed.

alessandrojean: What I was referring is that you probably need to override the `fetchPageList` to remove the `asObservableSuccess` and handle it manually to ignore 404 (and in this case returning an `Observable.just(emptyList())`)

curche: Ah okay. Yeah, there wasn't a particular need to handle 404 explicitly; I was refactoring the code around it while checking the api docs.
After looking at it again, I'll just let `fetchPageList` handle it like it normally does. Thanks for catching the dead code!

PR #10286, opened 2021-12-30
fix hitomi images

Basara-Hatake:
They changed it again...
If they do it again we have to think of a better way but i dont know any

Checklist:


alessandrojean:
Sounds like a cat and mouse game already.

PR #10366, opened 2022-01-06
Added option for author's notes at end of chapters

altaccosc:
Hi there, this is my first commit to the project, so I hope I've done everything correctly. I've tested the app out on my own phone, and it works well there.

I'm not sure if this is a feature that will be accepted, but I always like reading author's notes, and it wasn't possible to read those in Tachiyomi before this.

I'm planning on adding the same thing to the WebToon source, but I'll do that later and in another PR. (If this feature is even one that will be accepted in the first place)

Inspired by the XKCD source: https://github.com/tachiyomiorg/tachiyomi-extensions/blob/05addfb95e833e8f5a98c2b48a2fd554fa8867bb/src/all/xkcd/src/eu/kanade/tachiyomi/extension/all/xkcd/Xkcd.kt#L50-L67
https://github.com/tachiyomiorg/tachiyomi-extensions/blob/05addfb95e833e8f5a98c2b48a2fd554fa8867bb/src/all/xkcd/src/eu/kanade/tachiyomi/extension/all/xkcd/Xkcd.kt#L112
Checklist:


ObserverOfTime:
FWIW placehold.jp handles word wrapping automatically.

altaccosc:
Indeed, it does, and I'd have loved to use that, but the problem with that is that the height of the image is not known beforehand if we let placehold.jp handle the word wrapping.

If you (or anyone else) have a better way of doing this, please let me know!

Issue #10291, opened 2021-12-31
MangaDex: Option to block Bilibili and Comikey


FourTOne5:
### Describe your suggested feature

Since Group block is now live in API.

### Acknowledgements

- [X] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open issue.
- [X] I have written a short but informative title.
- [X] If this is an issue with the app itself, I should be opening an issue in the [app repository](https://github.com/tachiyomiorg/tachiyomi/issues/new/choose).
- [X] I have updated the app to version **[0.12.3](https://github.com/tachiyomiorg/tachiyomi/releases/latest)**.
- [X] I will fill out all of the requested information in this form.

CarlosEsco:
This would only be useful in the latest view since the ability to do so is only on the chapter list endpoint.  Browse and sorting by latest will still show them all

alessandrojean:
/edit-title "MangaDex: Option to block Bilibili and Comikey"

CarlosEsco:
@curche For this i would prefer
1.  bilibili, mangaplus, comikey and whatever the other one to be blocked always since they dont show in manga view anyways
2. a preference where additional UUID's can be added for a user, so it blocks during latest, and in the manga view

curche:
1. would be good to block all groups marked `official` in /group attributes; sadly I have to find all those manually
2. comma separated UUIDs entered in preference would have to do for now

CarlosEsco:
for 1.  ask dex to search the db for you they should be able to provide a list of uuids

jasontbright:
Hello, found this when a bunch of chapters disappeared from a manga I was reading.  Why are you blocking bilibili by default?  I tried following the issue line and didn't find any explanation.

PR #10376, opened 2022-01-06
MangaDex: Scanlation Group blocking

curche:
*The setting everyone's been waiting for (but not how they expected to be: random UUIDs waves hello)*

Adds option to block Scanlation groups by entering them as a comma-separated list of UUIDs

By Default, this blocks the External Link Official Groups (Bilibili, Comikey, Azuki, Mangaplus).
Note that this disables ALL bilibili chapters: including the *exception-causing* ones that are uploaded to dex and readable on dex ([Eg like this:](https://mangadex.org/title/e1c2aace-d3fb-453a-a580-886405ca27f4/ceo-s-sudden-proposal))

### Notes: 

 - First time having to deal with Preferences, that too a string plain text
 - Checked how Komga, Lanraragi does it and worked from that
-  Tried to make it a Set of Strings but ended up crashing the Extension settings page oof
   Welcoming any changes for dealing with this, thanks!

 PS. I feel this might be the jankiest PR I have sent uptil now phew (jk its [not](https://github.com/tachiyomiorg/tachiyomi-extensions/blob/01ecbdc8337681d9a943ca18bc05bf95f1baa7ab/src/en/reallifecomics/src/eu/kanade/tachiyomi/extension/en/reallifecomics/RealLifeComics.kt#L72) :p)

Checklist:


curche:
Closes #10291

curche:
Should I block the External Link Official Groups (Bilibili, Comikey, Azuki, Mangaplus etc) explicitly as well? Right now, its the default value for blocking which can edited by just clearing the setting

alessandrojean:
> Tried to make it a Set of Strings but ended up crashing the Extension settings page oof

Probably happened because you used the same key for the preference (the setting expect the key to be a String, since you're using a `EditTextPreference` and it crashes when it sees the preference is a String set instead, I guess).

You can try to save it as a String set using another key.

curche:
> > Tried to make it a Set of Strings but ended up crashing the Extension settings page oof
> 
> Probably happened because you used the same key for the preference (the setting expect the key to be a String, since you're using a `EditTextPreference` and it crashes when it sees the preference is a String set instead, I guess).
> 
> You can try to save it as a String set using another key just for it.

Nah, it was crashing on the first fresh install. Looking at the docs later, [EditTextPreference](https://developer.android.com/reference/androidx/preference/EditTextPreference) only saves as a String so that was probably it

IPTN:
Where do you go to edit these preferences for the extension/source?

PR #10400, opened 2022-01-09
Move TT from Madara to WpMangaReader

alessandrojean:
I removed the hardcoded ID because the URL path schema changed, and a migration will be necessary.

The PR also have some linting done automatically by AS.

Checklist:


Riztard:
> I removed the hardcoded ID

i think the old one will become obsolete and they need to uninstall and install the new one

alessandrojean:
No because the package name is still the same, so it will appear as an update.

PR #10469, opened 2022-01-15
Add extension fields for hasReadme and hasChangelog

ghostbear:
Depends on https://github.com/tachiyomiorg/tachiyomi/pull/6469

PR in Draft so it doesn't accidentally gets merged

Checklist:


common.gradle
```diff
@@ -24,11 +24,23 @@ android {
         versionCode extVersionCode
         versionName project.ext.properties.getOrDefault("libVersion", "1.2") + ".$extVersionCode"
         setProperty("archivesBaseName", "tachiyomi-$pkgNameSuffix-v$versionName")
+        def readmes = project.projectDir.listFiles({ File file ->
+            file.name.endsWith("README.md") ||
+            file.name.endsWith("CHANGELOG.md")
```
arkon, line 29: You seem to only allow specifically `README.md` and `CHANGELOG.md` in the app PR. Should these checks just check for those specific filenames?

src/all/mangapark/README.md
```diff
@@ -0,0 +1,20 @@
+# MangaPark
+
+Table of Content
+- [FAQ](#FAQ)
+
+[Uncomment this if needed; and replace &#40; and &#41; with ( and )]: <> (- [Guides]&#40;#Guides&#41;)
```
arkon, line 6: TODO?

template/README-REMOVED-TEMPLATE.md
```diff
@@ -0,0 +1,11 @@
+# Extension name
```
arkon, line 1: Is the intention to keep READMEs around even when the rest of an extension's assets are removed?

src/all/mangapark/README.md
ghostbear: No, just left it so one doesn't need to go to the template to get it. It is a comment so it will only show up when viewing the file RAW

template/README-REMOVED-TEMPLATE.md
ghostbear: Yes, I was thinking they could be left for a while that way it can explain to users why it was removed. But it will only be needed for extensions that have a README that way it doesn't throw 404 the first thing after getting removed

src/all/mangapark/README.md
```diff
@@ -0,0 +1,20 @@
...
+
+Don't find the question you are look for go check out our general FAQs and Guides over at [Extension FAQ](https://tachiyomi.org/help/faq/#extensions) or [Getting Started](https://tachiyomi.org/help/guides/getting-started/#installation)
+
+## FAQ
+
+### How do I deal with duplicate chapters?
```
FourTOne5, line 12: MangaPark v3 doesn't have duplicate chapters only v2 (en.mangapark) has duplicate chapter.

template/README-REMOVED-TEMPLATE.md
FourTOne5: Why not bundle the readmes with the app?
Before removing the extension we can give it a higher rate limit and update the extension with the reason it was removed under a special format which will only be shown app side if the extension is obsolete.

arkon: It'd be easier to point to up-to-date documentation without forcing users to update first. I can certainly see scenarios of updating the Markdown files without needing to actually trigger APK rebuilds too.

src/all/mangapark/README.md
arkon: Ah interesting, I wasn't aware of this comment syntax.

common.gradle
```diff
@@ -24,11 +24,23 @@ android {
         versionCode extVersionCode
         versionName project.ext.properties.getOrDefault("libVersion", "1.2") + ".$extVersionCode"
         setProperty("archivesBaseName", "tachiyomi-$pkgNameSuffix-v$versionName")
+        def readmes = project.projectDir.listFiles({ File file ->
+            file.name.equals("README.md") ||
+            file.name.equals("CHANGELOG.md")
```
arkon, line 29: What's the reasoning behind a changelog also counting as a readme?

ghostbear: Well Komga has a [CHANGELOG.md ](https://github.com/tachiyomiorg/tachiyomi-extensions/blob/master/src/all/komga/CHANGELOG.md) which is more human readable than the [commit history](https://github.com/tachiyomiorg/tachiyomi-extensions/commits/master/src/all/komga). This also includes Komga server requirements for the extension which can be helpful troubleshoot, if users read the changelog/commit history why the extension isn't functioning properly or want to know what changed.

The variable name may be a bit misleading it should probably be named `markdownFiles` or `filteredFiles` as the next two variable definitions check for the README and CHANGELOG

template/README-REMOVED-TEMPLATE.md
ghostbear: Only time an update is required is when README.md or CHANGELOG.md is added to update the AndroidManifest, but they are hopefully always added along actual changes, after that it will always point to up-to-date documentation

arkon (approved):
I didn't actually review the actual README contents, but the general approach of flagging the existence of the files seems fine to me.

You may want to bump the version of Komga too, since it already has a CHANGELOG.md file.

ghostbear:
> I didn't actually review the actual README contents

For the content of the actual READMEs, it's what is currently on the website. So when this gets merged FAQ needs to be updated and the extension list on the website needs to be updated to filter and link to these READMEs

arkon:
So were you intending on this to be merged? It's still in draft.

PR #10481, opened 2022-01-16
MangaDex: Group Blocking improvements

curche:
The PR does the following:
- Replace "Abandoned" with the new "Cancelled" Status Enum (was replaced in V5 API few months ago)
- No longer expose default blocked external-link/official scan grps in Dex
- Users can still block UUIDs for /chapter and /manga/ID/feed
- Use `excludedGroups` for external-link/official scan grps in Latest section **only** (uses /chapter)
- Use the newly added `pages` attrib in Chapter to filter out external-link/official scan grps in manga chapter list (uses /manga/ID/feed)
  - chapters with non-empty externalUrl cannot be read on MD and are filtered out
  - however, in some cases (bilibili comics), a non-empty externalUrl with a non-zero pages attrib means that it can be read on MD and so aren't filtered out

Checklist:


src/all/mangadex/src/eu/kanade/tachiyomi/extension/all/mangadex/MangaDexHelper.kt
```diff
@@ -328,17 +326,7 @@ class MangaDexHelper() {
...
-            }
-
-            // Bilibili special check. If it's a Bilibili chapter and the
-            // publishAt date is < now, it can be read on MD.
-            if (
-                !attr.externalUrl.isNullOrEmpty() &&
-                attr.externalUrl.contains(BILIBILI_URL) &&
-                parseDate(attr.publishAt) >= Date().time
-            ) {
+            if (!attr.externalUrl.isNullOrEmpty() && attr.pages == 0) {
```
Syer10, line 329: You shouldnt have to worry about empty strings, so it could be simplified to 
```suggestion
            if (attr.externalUrl != null && attr.pages == 0) {
```
Otherwise this looks good

Syer10 (approved):

PR #10538, opened 2022-01-21
Genkan.io add language support

jopejoe1:
Checklist:


src/all/genkanio/src/eu/kanade/tachiyomi/extension/all/genkanio/GenkanIOFactory.kt
```diff
@@ -0,0 +1,17 @@
...
+import eu.kanade.tachiyomi.source.SourceFactory
+
+class GenkanIOFactory : SourceFactory {
+    override fun createSources() = listOf(
+        GenkanIO("all"),
+        GenkanIO("ar"),
+        GenkanIO("en"),
+        GenkanIO("fr"),
+        GenkanIO("pl"),
+        GenkanIO("pt"),
```
alessandrojean, line 12: Can you give me any example of title that have Portuguese so I can check if is indeed `pt` or if it should be `pt-BR`?

alessandrojean: I would suggest that is better to change the source lang to `pt-BR` since I've particularly never saw a scanlator from Portugal and the vast majority of Portuguese translations in manga/manhwa/etc are Brazilian.

alessandrojean (approved):

Issue #6849, opened 2021-05-06
Standardised sub-language codes


jopejoe1:
##  Why/Project's Benefit/Existing Problem
Currently, there are 3 different language codes for Chinese Simple and Traditional each. For a total of 7 language codes just for Chinese alone, 6 of them with only one extension in it.
There are also 2 Lang codes for Portuguese Brazilian.

## Other Details
Some of it is just that some extension use different capitalization for the language codes than other extension resulting in multiple language categories in tachiyomi.

The majority uses full caps sub language codes.

CarlosEsco:
Would be nice to put it as a constant in the extensions lib so people are setting to the constant variable rather then hard coding

alessandrojean:
> There are also 2 Lang codes for Portuguese Brazilian.

I will work on this and submit a PR later to fix this part and replace all the usages of `pt-br` to `pt-BR`.

jopejoe1:
@CarlosEsco 
Should be reopened as only fixes the Portuguese Brazilian language.

CarlosEsco:
didn't realize he did fixes # lol

matheuswr89:
In Mangadex extension too necessary replace `pt-br` to `pt-BR`

alessandrojean:
@matheuswr89 As said above, it's replaced in CarlosEsco/tachiyomi-extensions#526, that's why I didn't change this part on my PR.

alessandrojean:
Fixed by #10562.

PR #10562, opened 2022-01-23
Update lang codes in a few extensions

jopejoe1:
Reduces the  amount of Chinese lang codes from 7 to 3
Edit: Also fixed some wrong lang codes

Related to #6889
Checklist:


alessandrojean:
Shouldn't be better to use `zh-Hans` and `zh-Hant` as Bilibili and MangaDex (and soon MangaToon) uses it?

jopejoe1:
> Shouldn't be better to use `zh-Hans` and `zh-Hant` as Bilibili and MangaDex (and soon MangaToon) uses it?

Yeag, for some reason i read it as `zh-HANS` and `zh-HANT` in dex
 
> Also needs to keep the old IDs in Batoto and Mangapark to avoid unnecessary migrations.

Done

alessandrojean:
I think MangaPark and Batoto constructor `lang` parameter needs to be changed to final, otherwise there will be a compile error.

```kotlin
final override val lang: String
```

jopejoe1:
> I think MangaPark and Batoto constructor `lang` parameter needs to be changed to final, otherwise there will be a compile error.
> 
> ```kotlin
> final override val lang: String
> ```

Worked fine in Android Studio for me, but changing it is not gone hurt

alessandrojean (approved):

alessandrojean (approved):

alessandrojean:
Will close #6849 now.

What needs to be done in another future PR is review the `zh`-only sources and change the language to `-Hans` or `-Hant`, so they get grouped together in the listing.

PR #10615, opened 2022-01-30
BlogTruyen: Fix parse page list not return enough pages in some manga

VietAnh14:
Checklist:

### DIFF
--- a/src/vi/blogtruyen/src/eu/kanade/tachiyomi/extension/vi/blogtruyen/BlogTruyen.kt
+++ b/src/vi/blogtruyen/src/eu/kanade/tachiyomi/extension/vi/blogtruyen/BlogTruyen.kt
@@ -14,6 +14,7 @@ import okhttp3.HttpUrl.Companion.toHttpUrlOrNull
 import okhttp3.OkHttpClient
 import okhttp3.Request
 import okhttp3.Response
+import org.json.JSONArray
 import org.jsoup.nodes.Document
 import org.jsoup.nodes.Element
 import java.text.SimpleDateFormat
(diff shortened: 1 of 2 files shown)


src/vi/blogtruyen/src/eu/kanade/tachiyomi/extension/vi/blogtruyen/BlogTruyen.kt
```diff
@@ -14,6 +14,7 @@ import okhttp3.HttpUrl.Companion.toHttpUrlOrNull
 import okhttp3.OkHttpClient
 import okhttp3.Request
 import okhttp3.Response
+import org.json.JSONArray
```
alessandrojean, line 17: Why not use `kotlinx.serialization` like the other extensions?

VietAnh14: I haven't looked at other extensions and get used to use `JSONArray`, I will change to use `kotlinx.serialization` as suggested.

alessandrojean: It's just a suggestion, we didn't remove the org.json usages in some extensions yet. You can let as it is if you want :+1:

alessandrojean (approved):

src/vi/blogtruyen/src/eu/kanade/tachiyomi/extension/vi/blogtruyen/BlogTruyen.kt
VietAnh14: Thank you, but I will just use it as suggested to make it consistent with other extensions.

PR #10675, opened 2022-02-03
Added Cloudflare client to Lanraragi

jferdous70:
Checklist:


alessandrojean:
> Updated `extVersionCode` value in build.gradle for individual extensions

Bump the `extVersionCode`.

arkon:
Did you test this?

jferdous70:
I don't have an example of non-protected server to test, if anybody has a URL i could check

alessandrojean:
Cloudflare client only do something if it detects that the `Server` header is `Cloudflare` and the website returned `403` and some other codes, so it should be fine, I guess?

jopejoe1:
You could always test with the demo server at `https://lrr.tvc-16.science` pretty sure that doesn't have Cloudflare.

jferdous70:
I tested both https://lrr.tvc-16.science and the cloudflare website, both work perfectly fine

PR #44, opened 2017-05-05
Add webtoons.com and Mangazuki.co (#48)

TonyToscana:
Add Webtoon.com source
Add mangazuki.co source (fixes #48 )

Issue #25, opened 2017-03-19
Page List Empty Error


jsantuele96:
(CATALOGUE) SOURCE REQUESTS:
can you add kissmang and m as nga fox to my
BUGS:
  tachiyomi is not connecting to any sources after the update. stull getting the page list is empty error. please help inees tachiyomi

Issue #24, opened 2017-03-19
Page List Empty Error


jsantuele96:
(CATALOGUE) SOURCE REQUESTS:
please help
BUGS:
 not being able to read any manga or download. page liest empty is showing up after having the news for updates v5.0. tried to update to fix the issue.nothing happened

inorichi:
It's already fixed. Please next time write which source is failing.

jsantuele96:
i am still getting the Page List is Empty error

PR #10860, opened 2022-02-18
Update Gmanga.kt

Mohamed-Ahmed-64597:
Changa Gmanga Domain

Checklist:

### DIFF
--- a/src/ar/gmanga/src/eu/kanade/tachiyomi/extension/ar/gmanga/Gmanga.kt
+++ b/src/ar/gmanga/src/eu/kanade/tachiyomi/extension/ar/gmanga/Gmanga.kt
@@ -35,7 +35,7 @@ import uy.kohesive.injekt.injectLazy
 
 class Gmanga : ConfigurableSource, HttpSource() {
 
-    private val domain: String = "gmanga.me"
+    private val domain: String = "gmanga.org"
 
     override val baseUrl: String = "https://$domain"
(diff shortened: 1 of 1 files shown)


alessandrojean:
You also didn't followed the PR checklist in this one either.

> Updated extVersionCode value in build.gradle for individual extensions

Please read the contributing guide and follow it. Also, have you tried and compiled the extension through Android Studio as the checklist asks?

vetleledaal:
Both domains are up, hosting the same content. Is there any announcement that they're moving? Or is this a mirror?

specterflare:
@vetleledaal Spent a fun couple of minutes playing internet sleuth and looking into it. Seems to be that .me is the older, notably less-trafficked original and .org is the newer, much more popular mirror.

1. The FAQ page for the .org has a link to a part of the .me site, just like the .me FAQ page does, which says to me that the .org is a copy.
2. Domain registrars indicate that the .me URL was registered almost 8 years ago, but the .org is less than two years old.
3. Incidently, but perhaps notably, alexa.org reports that the .org site is significantly more trafficked than the .me site (`#25,058` vs `#44,492` in terms of engagement). Importantly, the .me site was ranked as high as `#24,067` just last November, but has fallen rapidly, suggesting users are switching to the newer mirror.

I have no idea whether this means we should use the original, less popular site or the newer mirror for the extension.

PR #10936, opened 2022-02-26
MangaSwat : Fix images not loading 

zeredouni:


ahmedgamal17:
@zeredouni  i checked the selectors for manga swat and the did't change it to data-src and it's still src 
and for the issue that happens that the pics are not loading it's cloudflare issue i think beacaus it 's fixed after you open it in webview

zeredouni:
No the problem is not in cloudflare , It's because the site prevents copying images .
You can use this modified application to make sure .
[Tachiyomi_ MangaSwat 1.2.17.apk
](https://drive.google.com/u/1/uc?id=1i7WBhOen89n2rrHhiX7RqyPtlQ7F70dW&export=download)

ahmedgamal17:
@zeredouni ok I tested it now and it says no pages found so I was right and you don't need to change the attr class nome

zeredouni:
That's weird because it works so well for me .
did you clear chapter cache and database and cookies ?

ahmedgamal17:
Well I tested on j2k and it didn't work even clearing cache but worked an aniyomi I had delete the j2k data and it worked 🤔🤔👍🏼

PR #10814, opened 2022-02-13
Add support to string l10n in Madara

alessandrojean:
Marking it as WIP since it may have a build error, since I didn't check all overrides.

Changelog:

- Add support for string localization in the base class;
- Hide non-manga titles (text and video) by default in the selector;
- Add additional manga title selector;
- Fix missing SS icons.

Checklist:

- [x] Updated `extVersionCode` value in `build.gradle` for individual extensions
- [x] Updated `overrideVersionCode` or `baseVersionCode` as needed for all multisrc extensions
- [x] Referenced all related issues in the PR body (e.g. "Closes #xyz")
- [x] Added the `isNsfw = true` flag in `build.gradle` when appropriate
- [x] Have not changed source names
- [x] Have explicitly kept the `id` if a source's name or language were changed
- [x] Have tested the modifications by compiling and running the extension through Android Studio

PR #10262, opened 2021-12-28
Madara removed hardcoded genres and a bit of cleanup

jopejoe1:
Checklist:

- [x] Updated `extVersionCode` value in `build.gradle` for individual extensions
- [x] Updated `overrideVersionCode` or `baseVersionCode` as needed for all multisrc extensions
- [x] Referenced all related issues in the PR body (e.g. "Closes #xyz")
- [x] Added the `isNsfw = true` flag in `build.gradle` when appropriate
- [x] Have not changed source names
- [x] Have explicitly kept the `id` if a source's name or language were changed
- [x] Have tested the modifications by compiling and running the extension through Android Studio

PR #10998, opened 2022-03-02
Add back versionId to LHTranslation

FlaminSarge:
versionId 1 was for their FMReader case, versionId 2 is for their Madara case.

If they ever migrate back to FMReader, versionId will (probably) no longer be needed.

Checklist:


alessandrojean (approved):

FlaminSarge:
Originally removed by #10262, but the ext version only updated when #10814 was merged. Some time in the middle, an extra apk for 1.2.15 was built that had the new sourceId, but those that already had the extension didn't get that update.

jopejoe1 (approved):
Sorry that was my mistake.

PR #11150, opened 2022-03-16
Fix date and chapter name from ReaperScans EN

SilverBeamx:
Checklist:

This PR aims to fix chapter names and dates in the ReaperScan En extension, which are currently broken and set to null.

### DIFF
--- a/multisrc/overrides/madara/reaperscans/src/ReaperScansFactory.kt
+++ b/multisrc/overrides/madara/reaperscans/src/ReaperScansFactory.kt
@@ -21,27 +21,23 @@ abstract class ReaperScans(
     override val baseUrl: String,
     lang: String,
     dateFormat: SimpleDateFormat = SimpleDateFormat("MMMM dd, yyyy", Locale.US)
-) : Madara("Reaper Scans", baseUrl, lang, dateFormat) {
+) : Madara("Reaper Scans", baseUrl, lang, dateFormat)
+
+class ReaperScansEn : ReaperScans("https://reaperscans.com", "en", SimpleDateFormat("MMM dd,yyyy", Locale.US)) {
 
     override fun chapterFromElement(element: Element): SChapter {
         val chapter = SChapter.create()
-
         with(element) {
             select(chapterUrlSelector).first()?.let { urlElement ->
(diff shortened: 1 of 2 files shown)


alessandrojean:
Since you removed the `chapterFromElement` override from the base `ReaperScans` class (which is used by the other languages of the website), you need to check if the chapter parsing them are still working too.

SilverBeamx:
Yes, i moved the generic one, that was wrong, into the EN version, because the BR version already has its custom implementation, and it isn't needed elsewhere. As i mentioned on Discord, should i ovverride the BR versionId to the old version because i didn't touch that part, or just bumping the overrall versionCode is fine?

alessandrojean:
Bumping the `overrideVersionCode` in `MadaraGenerator` the way you did is right 👍🏻 .

multisrc/overrides/madara/reaperscans/src/ReaperScansFactory.kt
```diff
@@ -21,29 +21,25 @@ abstract class ReaperScans(
...
-            chapter.date_upload = select("span.chapter-release-date i").firstOrNull()?.text().let { parseChapterDate(it) }
+            chapter.date_upload = select("span.chapter-release-date > i").firstOrNull()?.text().let { parseChapterDate(it) }
         }
-
         return chapter
     }
 }
 
-class ReaperScansEn : ReaperScans("https://reaperscans.com", "en") {
-    override val versionId = 2
```
jopejoe1, line 44: The versionId override should not be removed as it would change the extension id

jopejoe1 (changes requested):

multisrc/overrides/madara/reaperscans/src/ReaperScansFactory.kt
alessandrojean: Good catch!

SilverBeamx: Yea, this is what i was asking previously. I probably explained myself poorly. Should be fixed now!

arkon:
@mergifyio rebase

Issue #10733, opened 2022-02-06
Imhentai is broken


Hounddraco:
### Source information

Imhentai 1.2.4

### Steps to reproduce

Open IMhentai 

### Expected behavior

Should show content 

### Actual behavior

Doesn't show any content 

### Tachiyomi version

0.13.1

### Android version

Android 11

### Acknowledgements

- [X] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open issue.
- [X] I have written a short but informative title.
- [X] I have updated the app to version **[0.13.1](https://github.com/tachiyomiorg/tachiyomi/releases/latest)**.
- [X] I have updated all installed extensions.
- [X] I have tried the [troubleshooting guide](https://tachiyomi.org/help/guides/troubleshooting/).
- [X] If this is an issue with the app itself, I should be opening an issue in the [app repository](https://github.com/tachiyomiorg/tachiyomi/issues/new/choose).
- [X] I will fill out all of the requested information in this form.

Basara-Hatake:
their search is broken thats why it doesnt work
![image](https://user-images.githubusercontent.com/53382992/152691804-54f6c334-1580-4db3-b4c3-22de882463a4.png)

alessandrojean:
Closing since it's a site issue.

PR #11187, opened 2022-03-20
imhentai overload message

Basara-Hatake:
IMHentai's search is currently overloaded and people are complaining in support that it doesnt work and this has already happened a few times before e.g #10733
So I decided it would be a good idea to throw an error so that it gets shown.

Checklist:


alessandrojean:
The way you did will probably cause exceptions because it's consuming the body from response twice. I think it would probably be better to do that in an interceptor maybe, throwing an IOException there instead.

Basara-Hatake:
> The way you did will probably cause exceptions because it's consuming the body from response twice (in the case it's not overloaded). I think it would probably be better to do that in an interceptor maybe, throwing an IOException there instead.

i added it now but im not sure if what i did is correct it seems to work tho

alessandrojean:
Looks good now 👍🏻

PR #11312, opened 2022-04-02
Add some missing information to CONTRIBUTING.md guide

alessandrojean:
Checklist:


CONTRIBUTING.md
```diff
@@ -157,20 +159,19 @@ The class which is referenced and defined by `extClass` in `build.gradle`. This
 | ----- | ----------- |
 | `name` | Name displayed in the "Sources" tab in Tachiyomi. |
 | `baseUrl` | Base URL of the source without any trailing slashes. |
-| `lang` | An ISO 639-1 compliant language code (two letters in lower case). |
+| `lang` | An ISO 639-1 compliant language code (two letters in lower case in most cases, but can also include the country/dialect part by using a simple dash character). |
```
vetleledaal, line 162: should mention that the 2nd part must be uppercase

CONTRIBUTING.md
```diff
@@ -183,7 +184,24 @@ a.k.a. the Latest source entry point in the app (invoked by tapping on the "Late
 
 - When the user searches inside the app, `fetchSearchManga` will be called and the rest of the flow is similar to what happens with `fetchPopularManga`.
     - If search functionality is not available, return `Observable.just(MangasPage(emptyList(), false))`
-- `getFilterList` will be called to get all filters and filter types. **TODO: explain more about `Filter`**
+- `getFilterList` will be called to get all filters and filter types.
+
+##### Filters
```
vetleledaal, line 189: Maybe mention `UriPartFilter`, which is commonly used. There are multiple variants, but the most popular one used on 57 occasions looks like:
```kt
open class UriPartFilter(displayName: String, private val vals: Array<Pair<String, String>>) :
    Filter.Select<String>(displayName, vals.map { it.first }.toTypedArray()) {
    fun toUriPart() = vals[state].second
}
```
In 9 cases first and second is swapped. The most common variation is adding/removing private from the class or vals, and adding a default selection which defaults to 0 when not set.

CONTRIBUTING.md
jopejoe1: > should mention that the 2nd part must be uppercase

Not the case with the chinese ones.

CONTRIBUTING.md
```diff
@@ -194,27 +212,27 @@ a.k.a. the Latest source entry point in the app (invoked by tapping on the "Late
...
-    - If `Page.imageUrl`s are available immediately, you should pass them here. Otherwise, you should set `page.url` to a page that contains them and override `imageUrlParse` to fill those `imageUrl`s.
+    - **The list should be sorted descending by the source order**.
 
 #### Chapter
 
 - After a chapter list for the manga is fetched and the app is going to cache the data, `prepareNewChapter` will be called.
-- `SChapter.date_upload` is the [UNIX Epoch time](https://en.wikipedia.org/wiki/Unix_time) **expressed in miliseconds**.
-    - If you don't pass `SChapter.date_upload`, the user won't get notifications for new chapters. refer to [this issue](https://github.com/tachiyomiorg/tachiyomi/issues/2089) for more info. `System.currentTimeMillis()` works as a substitute when real data is not available.
+- `SChapter.date_upload` is the [UNIX Epoch time](https://en.wikipedia.org/wiki/Unix_time) **expressed in milliseconds**.
+    - If you don't pass `SChapter.date_upload`, the user won't get notifications for new chapters and the app will use the fetch date instead. Refer to [this issue](https://github.com/tachiyomiorg/tachiyomi/issues/2089) for more info.
```
FourTOne5, line 223: Since the app assigns fetch date as update date the chapters will show up in Updates tab.

CONTRIBUTING.md
alessandrojean: I will leave as it is then.

CONTRIBUTING.md
alessandrojean: Hmm, I can include it as an example :+1:

PR #11402, opened 2022-04-09
Guya: update domain to fix covers and pages

scb261:
No opened issues

Checklist:


funkyhippo:
I wouldn't recommend merging this as we're not expecting to move the domain over. The current redirect is temporary as we ran into some hiccups with Cloudflare.

Covers should be back now since we've set up a catch-all redirect to the subdomain for now.

scb261:
Huh, and the fix doesn't seem to work anymore. I'm pretty sure it loaded pages just fine right after I've build it

funkyhippo:
> Huh, and the fix doesn't seem to work anymore. I'm pretty sure it loaded pages just fine right after I've build it

Hmm, that's a bit strange since the redirect on our end should be equivalent to your fix. It's likely due to the higher latency to hit our host since the `manga` subdomain isn't behind a CDN at the moment.

Regardless, the Guya extension will have impacted performance for the next little bit until Cloudflare stops throttling our traffic.

Issue #11509, opened 2022-04-18
Hentai Cube: Changed their domain


mihawk1890:
### Source information

Hentai Cube 1.2.20

### Steps to reproduce

1.open source
2.select hentai cube
3.doesn't find any results, got HTTP error 400

### Expected behavior

This should show the lastest update of hentaicb.top

### Actual behavior

Doesn't find any results, got HTTP error 400

### Tachiyomi version

1.8.2

### Android version

Android 12

### Acknowledgements

- [X] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open issue.
- [X] I have written a short but informative title.
- [X] I have updated the app to version **[0.13.3](https://github.com/tachiyomiorg/tachiyomi/releases/latest)**.
- [X] I have updated all installed extensions.
- [X] I have tried the [troubleshooting guide](https://tachiyomi.org/help/guides/troubleshooting/).
- [X] If this is an issue with the app itself, I should be opening an issue in the [app repository](https://github.com/tachiyomiorg/tachiyomi/issues/new/choose).
- [X] I will fill out all of the requested information in this form.

alekc:
@alessandrojean I can do this one, but I am unsure in regard to "Have explicitly kept the id if a source's name or language were changed". 

The source changed the name as well "Hentai Cube"+>"Hentai CB", do you have any example of how to maintain the id? (contributing file doesn't provide any info in regard). 

Cheers.

alessandrojean:
You can get the current ID by inspecting the `index.json` file in the `repo` branch. Then you do like this:

```kotlin
override val id: Long = <the-id>
```

PR #11552, opened 2022-04-21
feat: moved hentaicube

alekc:
Closes https://github.com/tachiyomiorg/tachiyomi-extensions/issues/11509

Checklist:


alekc:
for some reason, images are duplicated.

alessandrojean:
You need to keep the same package name, otherwise users will not receive the update if they already have the extension.

So moving the res and src isn't necessary. You can force the package name to be the same as the old one in the MadaraGenerator.kt by using `pkgName = "hentaicube"`.

alekc:
@alessandrojean ok, latest pushed commit should have everything needed (p.s. might be a good idea to add some info about how to deal with source renaming in Contributing page).

PR #11564, opened 2022-04-22
[skip ci] feat: added section about http interception in CONTRIBUTING.md

alekc:
Checklist:


CONTRIBUTING.md
```diff
@@ -429,6 +429,56 @@ Instead, once you've built and installed your extension on the target device, us
...
+and point your browser to http://0.0.0.0:8081/
+
+#### OkHttp proxy setup
+Since most of the manga sources are going to use https we need to disable ssl verification in order to use web debugger.
+
+Amend your source kotlin file (here we are using MangaBuddy)
+```kotlin
+class MangaBuddy : MadTheme(
+    "MangaBuddy",
+    "https://mangabuddy.com",
```
CarlosEsco, line 452: Just make this reference a random non real source / theme.  and its good to go.

alekc: done

alessandrojean:
Why not just use Logcat and filter for `OkHttpClient` tag instead? The app already has a logger inside it if you use the Preview version (as intended for developing extensions).

Using Logcat doesn't need any SSL bypass and other workarounds done that mitm requires.

alekc:
Well, first of all, I was not aware of that possibility :)

As for the rest: I've tried to follow your instruction, but I still can't see the HTTP flow from the logcat.

Also, I'd rather prefer having webui which I can easily click through/expand as I need rather then parsing logcat output.

PR #11803, opened 2022-05-10
ManhwaIndo error 404

TheKingTermux:
Just fix error 404

* Renaming Manhwa Indo to ManhwaIndo on ManhwaIndo.kt override WPMangaStream directory 
* Updated overrideVersionCode 

Checklist:


multisrc/overrides/wpmangareader/manhwaindo/src/ManhwaIndo.kt
```diff
@@ -5,7 +5,7 @@ import java.text.SimpleDateFormat
 import java.util.Locale
 
 class ManhwaIndo : WPMangaReader(
-    "Manhwa Indo", "https://manhwaindo.id", "id",
+    "ManhwaIndo", "https://manhwaindo.id", "id",
```
alessandrojean, line 8: This will change the generated `id` and will make users being obligated to unnecessarily migrate back to the source. How is this supposed to fix the 404 error?

TheKingTermux: The ekstension say error 404
But the WebView is work normaly
And I'm confused what should I do

PR #11381, opened 2022-04-08
Add Haoman6

anenasa:
Closes  #11158

Checklist:

- [x] Updated `extVersionCode` value in `build.gradle` for individual extensions
- [x] Updated `overrideVersionCode` or `baseVersionCode` as needed for all multisrc extensions
- [x] Referenced all related issues in the PR body (e.g. "Closes #xyz")
- [x] Added the `isNsfw = true` flag in `build.gradle` when appropriate
- [x] Have not changed source names
- [x] Have explicitly kept the `id` if a source's name or language were changed
- [x] Have tested the modifications by compiling and running the extension through Android Studio

PR #11839, opened 2022-05-13
MCCMS multisrc theme

kasperskier:
- Makes Haoman6 (#11381) into MCCMS theme
- Makes Haomanwu (#11514, #11538) into MCCMS theme
- Adds Haoman6 (g-lens) as per #11158 (which is not a mirror site; it has the same name but different data)
- Adds Haoman8

Checklist:


alessandrojean (approved):

alessandrojean:
If you happen to know what language they are, `zh-Hans` or `zh-Hant`, it's good to make it specific. Also, the ones migrated to this new theme generator, will the version be greater than the ones that already existed before? Otherwise users will not receive the update.

kasperskier:
@alessandrojean As per https://github.com/tachiyomiorg/tachiyomi-extensions/pull/11840#issuecomment-1126051575 (your comment), I'll change the extension names back to English. But the building script is using the name specified in the generator as extension name **and** the source name, and there is no other way to change the source name than to write an override class. This is a bit confusing.

alessandrojean:
Yeah, to have different names you need to have an override, it's how it is now. It definitely could be improved in the future though.

alessandrojean:
If it's mixed you can let as `zh` only, no problem 👍🏻

PR #11913, opened 2022-05-20
Bilibili Manga: Read unlocked chapters

anenasa:
No related issues found.
I override headersBuilder() to log in from webview and chapterListParse() to change filter.

Checklist:


multisrc/src/main/java/eu/kanade/tachiyomi/multisrc/bilibili/BilibiliDto.kt
```diff
@@ -38,6 +38,8 @@ data class BilibiliComicDto(
 @Serializable
 data class BilibiliEpisodeDto(
     val id: Int,
+    @SerialName("is_in_free") val is_in_free: Boolean,
```
alessandrojean, line 41:
```suggestion
    @SerialName("is_in_free") val isInFree: Boolean,
```

Change it to camelCase. Update other usages.

Does these appear in the Bilibili Comics API as well? If they don't, it will cause issues and they need to be nullable and have a default value of false.

multisrc/src/main/java/eu/kanade/tachiyomi/multisrc/bilibili/BilibiliDto.kt
```diff
@@ -38,6 +38,8 @@ data class BilibiliComicDto(
...
+    @SerialName("is_locked") val is_locked: Boolean,
```
alessandrojean, line 42:
```suggestion
    @SerialName("is_locked") val isLocked: Boolean,
```

multisrc/overrides/bilibili/bilibilimanga/src/BilibiliManga.kt
```diff
@@ -22,6 +26,22 @@ class BilibiliManga : Bilibili(
         .addInterceptor(SpecificHostRateLimitInterceptor(COVER_CDN_URL.toHttpUrl(), 2))
         .build()
 
+    override fun headersBuilder() = Headers.Builder().apply {
+        add("User-Agent", "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/88.0.4324.150 Safari/537.36 Edg/88.0.705.63")
```
alessandrojean, line 30: The app already sends this same User-Agent if not specified, I don't get why it got added.

multisrc/overrides/bilibili/bilibilimanga/src/BilibiliManga.kt
```diff
@@ -22,6 +26,22 @@ class BilibiliManga : Bilibili(
...
+
+    override fun chapterListParse(response: Response): List<SChapter> {
+        val result = response.parseAs<BilibiliComicDto>()
+
+        if (result.code != 0) {
+            return emptyList()
+        }
+
+        return result.data!!.episodeList
+            .filter { episode -> episode.is_in_free || !episode.is_locked }
```
alessandrojean, line 41: Just to confirm, this `isLocked` is false only if the user has bought or has the chapter in their account, right?

The Bilibili Manga API don't check for any access token like Bilibili Comics does?

alessandrojean (changes requested):

multisrc/src/main/java/eu/kanade/tachiyomi/multisrc/bilibili/BilibiliDto.kt
anenasa: Fix.
Yes, they do appear in the Bilibili Comics, just not useful for checking if the chapter is unlocked.

multisrc/overrides/bilibili/bilibilimanga/src/BilibiliManga.kt
anenasa: No, it doesn't. headersBuilder is overriden in https://github.com/tachiyomiorg/tachiyomi-extensions/blob/5716819a8da71e597f46a912e159638b38dabf24/multisrc/src/main/java/eu/kanade/tachiyomi/multisrc/bilibili/Bilibili.kt#L48.
This is what I get in webview without override: Mozilla/5.0 (Linux; Android 12; Redmi Note 8T Build/SP2A.220405.004; wv) AppleWebKit/537.36 (KHTML, like Gecko) Version/4.0 Chrome/101.0.4951.61 Mobile Safari/537.36, so it open the mobile version of the site.

alessandrojean: Hmm, probably an app bug, it should send the same Edge UA if not specified. Can keep it then, but make it a constant in a `companion object` for better code legibility.

multisrc/overrides/bilibili/bilibilimanga/src/BilibiliManga.kt
anenasa: > Just to confirm, this isLocked is false only if the user has bought or has the chapter in their account, right?

1. Free chapter
2. User has bought the chapter
3. 等免 (wait to unlock for three days)

alessandrojean: The 3rd one refers to when the user have waited these three days, right? Just to make sure.

anenasa: > The Bilibili Manga API don't check for any access token like Bilibili Comics does when reading the chapter...?
Yes, it is in cookie, I think cookie name is SESSDATA.

alessandrojean: The warning thing could be improved later then. I will test it tomorrow probably and then it can be merged.

anenasa: Is hiding locked chapters really necessary? Some other extensions show a lock in chapter name for locked chapters and I think this is a better idea. Here is an example:
https://github.com/tachiyomiorg/tachiyomi-extensions/blob/5716819a8da71e597f46a912e159638b38dabf24/src/zh/tencentcomics/src/eu/kanade/tachiyomi/extension/zh/tencentcomics/TencentComics.kt#L52

anenasa: > The 3rd one refers to when the user have waited these three days, right? Just to make sure.

The waiting time is different for different comics, chapters unlocked with this methed will be re-locked after three days. (Just my observation, not sure if it is always three days.)

alessandrojean: > Is hiding locked chapters really necessary? Some other extensions show a lock in chapter name for locked chapters and I think this is a better idea

That would potentially make users asking how to unlock these chapters, so it's better to hide them if they're not in the user account or free. The warning and the paid genre should do the job instead.

anenasa: For reference, here is my test result:
|                            | is_in_free | is_locked | pay_gold | pay_mode | unlock_type |
|----------------------------|------------|-----------|----------|----------|-------------|
| free                       | false      | false     | 0        | 0        | 0           |
| paid_unlocked              | false      | false     | >0       | 1        | 1           |
| wait_unlocked              | false      | false     | >0       | 1        | 2           |
| free for limited time 限免 | true       | true      | >0       | 1        | 0           |
| wait                       | false      | true      | >0       | 1        | 0           |
| paid                       | false      | true      | >0       | 1        | 0           |

multisrc/overrides/bilibili/bilibilimanga/src/BilibiliManga.kt
anenasa: Changed.

alessandrojean (approved):

alessandrojean:
@anenasa Can you update the `BilibiliIntl.hasPaidChaptersWarning` Chinese text to reflect the most updated English version that tells about sign in through WebView?

PR #11928, opened 2022-05-21
Change: sort by chapter number.

mao09:
Checklist:

### DIFF
--- a/multisrc/src/main/java/eu/kanade/tachiyomi/multisrc/madara/Madara.kt
+++ b/multisrc/src/main/java/eu/kanade/tachiyomi/multisrc/madara/Madara.kt
@@ -787,7 +787,7 @@ abstract class Madara(
 
         countViews(document)
 
-        return chapterElements.map(::chapterFromElement)
+        return chapterElements.map(::chapterFromElement).sortedBy { it -> it.chapter_number }.reversed()
     }
 
     override fun chapterListSelector() = "li.wp-manga-chapter"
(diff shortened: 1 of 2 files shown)


multisrc/src/main/java/eu/kanade/tachiyomi/multisrc/madara/Madara.kt
```diff
@@ -787,7 +787,7 @@ abstract class Madara(
 
         countViews(document)
 
-        return chapterElements.map(::chapterFromElement)
+        return chapterElements.map(::chapterFromElement).sortedBy { it -> it.chapter_number }.reversed()
```
saud-97, line 790: I'm not sure if I understood this correctly but you overridden the source sort and forced it to be by chapter number here?

multisrc/src/main/java/eu/kanade/tachiyomi/multisrc/madara/Madara.kt
```diff
@@ -814,6 +814,7 @@ abstract class Madara(
                 ?: select("span a").firstOrNull()?.attr("title")?.let { parseRelativeDate(it) }
                 ?: parseChapterDate(select(chapterDateSelector()).firstOrNull()?.text())
         }
+        chapter.chapter_number = Regex("""\d+\.?\d*""").find(chapter.name)?.value?.toFloat() ?: -1f
```
alessandrojean, line 817: This will make any number in the chapter to being recognized as the chapter number, even if it's part of the title and not the number (position) itself. The app already has a chapter number recognition system so this doesn't need to be specified necessarily. Individual sources with reversed chapter lists can be fixed manually through overrides.

saud-97: `parseChapterNumber `method inside `ChapterRecognition.kt`

arkon:
As @alessandrojean said, sources shouldn't be trying to parse/set chapter numbers unless they're explicitly known. The app handles that through its own parsing, which users can use from the sorting menu.

Issue #9075, opened 2021-09-15
Comikey: HTTP Error 403/Cloudflare


Canlethan:
### Source information

Comikey

### Steps to reproduce

Trying to read any chapter with the comikey extension gives back a 403 error. Webview works. 

### Expected behavior

Being able to read and download the chapter. 

### Actual behavior

Http error 403

### Tachiyomi version

0.12.2

### Android version

Android 10

### Other details

@lord-ne who wrote the extension. 

### Acknowledgements

- [X] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open issue.
- [X] I have written a short but informative title.
- [X] I have updated the app to version **[0.12.2](https://github.com/tachiyomiorg/tachiyomi/releases/latest)**.
- [X] I have updated all installed extensions.
- [X] I have tried the [troubleshooting guide](https://tachiyomi.org/help/guides/troubleshooting/).
- [X] If this is an issue with the app itself, I should be opening an issue in the [app repository](https://github.com/tachiyomiorg/tachiyomi/issues/new/choose).
- [X] I will fill out all of the requested information in this form.

GizmoTheGreen:
Got 403 on comikey too. Can't update chapters or read any chapters. WebView works fine and I'm logged in as well when i check it.

alessandrojean:
They added Cloudflare protection to the website. Nothing much you can do besides waiting for them to lower the protection or migrating to other source.

Alveel:
How does that work exactly? It works perfectly fine in Tachi's WebView, but in the reader I get an instant 403 every time. I never see anything Cloudflare related, like you sometimes see on other websites.

I only just got into some Comikey series, so it's sad to see this happening :(

lord-ne:
They're blocking access to their website if you don't have JavaScript enabled (and an HTTP GET request does not run JavaScript). They're even blocking the API, which seems to defeat the purpose of an API if you ask me.

It probably is possible to bypass this within Tachiyomi, I assume that's what the `cloudflareClient` is for, but I don't know how to do it.

alessandrojean:
> I assume that's what the `cloudflareClient` is for

This client only have an additional interceptor that tries to resolve the challenge Cloudflare provides in order to access the website, but it's outdated at the moment, and it does not work with the majority of sites.

[...]

PR #11971, opened 2022-05-25
Delete Comikey Extension

lord-ne:
Comikey extension hasn't worked for a while (about 8 months), due to changes made to the site. Specifically, they added Cloudflare protection, encryption, and scrambling images. I don't think the extension is realistically fixable, and it's been a while and no one else has fixed it either, so I'm opting to remove the extension.

Closes https://github.com/tachiyomiorg/tachiyomi-extensions/issues/9075

Checklist:


alessandrojean (approved):

ObserverOfTime:
Wouldn't it be possible to use their API?

lord-ne:
Yeah, several things actually are using the API, but even those parts don't work anymore due to the DRM encryption

CarlosEsco:
Undecided on this.  Its not a cat and mouse game afaik.  Just hasnt worked in 8 months

lord-ne:
I'll explain my reasoning for why I think this can't really be fixed:

Currently, there are two issues preventing the extension from working:

1. The API requests are protected by Cloudflare. Example: https://comikey.com/sapi/comics/37/read?format=json&content=EPI-nDq4Mk

2. The actual page listing the image urls is DRM protected. Example: Load the above URL, and click the link (relay-us.ebub.rocks/...) in the href attirbute of the response. You'll get an encrypted JSON document. Note that these relay-us.ebub.rocks links are the only way to access the URLs of the chapter images, as far as I'm aware.

(The third issue I mentioned earlier, images being scrambled, doesn't seem to be a things anymore, or at least not on most series)

The first issues if probably fixable. Maybe making the user open the url in webview and do the Cloudflare captcha once would allow future requests, or something like that, idk. But the second issue is my main concern. Breaking DRM inside a Tachiyomi extension seems incredibly difficult, maybe even not possible.

Issue #11881, opened 2022-05-15
Manga Pro Z: Pagelist returns Error 404 


Saud-97:
### Source information

Manga Pro Z [ar] 1.2.15

### Steps to reproduce

open manga "I Am Carrying Gold From the Post-Apocalyptic World"
try to download or browse latest chapter

### Expected behavior

Should work normally like other chapters in the same manga

### Actual behavior

Throwing Error 404 and no pages get downloaded at all. I checked webview and the chapter is working normally. But is protected first by an ad URL shortner.

### Tachiyomi version

j2k 1.5.2

### Android version

Android 9

### Other details

I only found the issue at latest chapter, the earlier ones are okay.

### Acknowledgements

- [X] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open issue.
- [X] I have written a short but informative title.
- [X] I have updated the app to version **[0.13.4](https://github.com/tachiyomiorg/tachiyomi/releases/latest)**.
- [X] I have updated all installed extensions.
- [X] I have tried the [troubleshooting guide](https://tachiyomi.org/help/guides/troubleshooting/).
- [X] If this is an issue with the app itself, I should be opening an issue in the [app repository](https://github.com/tachiyomiorg/tachiyomi/issues/new/choose).
- [X] I will fill out all of the requested information in this form.

Saud-97:
Can someone point out what method I need to override to fix this issue of cover fetching?
As it pulls the cover correctly during first fetch but after updating manga info it will delete the cover.

Issue #11809, opened 2022-05-10
Manga Pro Z (AR): Cover image gets deleted after first manga details update


Saud-97:
### Source information

Manga Pro Z (AR) 1.2.15

### Steps to reproduce

1. Browse the source
2. Notice the cover is loaded correctly first time
3. go inside any manga and update details
4. the cover image will get deleted
5. go back and notice the cover is no longer shown when viewing the manga from browse/library

### Expected behavior

Cover should get updated

### Actual behavior

Cover gets deleted

### Tachiyomi version

0.13.4

### Android version

Android 9

### Acknowledgements

- [X] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open issue.
- [X] I have written a short but informative title.
- [X] I have updated the app to version **[0.13.4](https://github.com/tachiyomiorg/tachiyomi/releases/latest)**.
- [X] I have updated all installed extensions.
- [X] I have tried the [troubleshooting guide](https://tachiyomi.org/help/guides/troubleshooting/).
- [X] If this is an issue with the app itself, I should be opening an issue in the [app repository](https://github.com/tachiyomiorg/tachiyomi/issues/new/choose).
- [X] I will fill out all of the requested information in this form.

Saud-97:
Can someone point out what method I need to override to fix this issue of cover fetching?
As it pulls the cover correctly during first fetch but after updating manga info it will delete the cover.

PR #12028, opened 2022-05-31
Manga Pro Z -> Manga Pro

Saud-97:
The source have removed Z from their name and changed their logo. I confirmed it by speaking to them at their discord and from their website URL and logo.

Also I removed a code that was removing FREE from the chapters names. Since this causes readers confusion as the last chapter that is shown in the app is considered paid or protected by URL shortner so it would always cause https://github.com/tachiyomiorg/tachiyomi-extensions/issues/11881

I was planning to fix the two issues https://github.com/tachiyomiorg/tachiyomi-extensions/issues/11881 and https://github.com/tachiyomiorg/tachiyomi-extensions/issues/11809 but unfortunately the project wouldn't compile in my PC. I only did the rebranding in this PR.

Checklist:


Saud-97:
> Is the name change strictly necessary? It _will_ end up orphaning downloads, so people would need to go rename the folder.

I changed it to differentiate between free chapters and the last chapter that is usually not free and will require user to go through an ad shortened URL similar to https://github.com/tachiyomiorg/tachiyomi-extensions/issues/11881

PR #12047, opened 2022-06-02
ManhwaIndo: using new changelog

Riztard:
add CHANGELOG.md

Checklist:

- [x] Updated `extVersionCode` value in `build.gradle` for individual extensions
- [x] Updated `overrideVersionCode` or `baseVersionCode` as needed for all multisrc extensions
- [x] Referenced all related issues in the PR body (e.g. "Closes #xyz")
- [x] Added the `isNsfw = true` flag in `build.gradle` when appropriate
- [x] Have not changed source names
- [x] Have explicitly kept the `id` if a source's name or language were changed
- [x] Have tested the modifications by compiling and running the extension through Android Studio

PR #12051, opened 2022-06-02
Copy README/CHANGELOG files for multisrc overrides correctly

arkon:
Fixes https://github.com/tachiyomiorg/tachiyomi-extensions/pull/12047#issuecomment-1145400458


arkon:
Apparently didn't account for files in non-override folders, so I'll address that in a bit.

PR #12129, opened 2022-06-08
[RU] MangaClub: Update status to extensions-lib 1.3

D3xron:
Checklist:


src/ru/mangaclub/src/eu/kanade/tachiyomi/extension/ru/mangaclub/MangaClub.kt
```diff
@@ -100,15 +100,15 @@ class MangaClub : ParsedHttpSource() {
         status = when (document.select("div.info a[href*=status_translation]").text().trim()) {
             "Продолжается" -> if (licensedStatus) SManga.LICENSED else SManga.ONGOING
             "Завершен" -> if (licensedStatus) SManga.LICENSED else SManga.COMPLETED
-            else -> SManga.UNKNOWN
+            else -> SManga.ON_HIATUS
```
alessandrojean, line 103: Won't this make all failed parsing attempts to be marked as _on hiatus_? What if they change the words that are used to mark things as completed or ongoing?

Instead, do they indicate if the series is on hiatus on their website, so you can use it in the checks? It's better to let the `else` case to these cases where the parsing fails and it should return `SManga.UNKNOWN`.

D3xron: This site doesn't have more statuses, only LICENSED, ONGOING, COMPLETED and ON_HIATUS, but well, I will changed it

alessandrojean (approved):

PR #12237, opened 2022-06-19
MDB multisrc theme

stevenyomi:
Rewrite and merge ManhuaDB + Maofly into multisrc theme as promised in https://github.com/tachiyomiorg/tachiyomi-extensions/pull/11930#issuecomment-1133655606.
- Better parsing
- Category filters
- Change icon for ManhuaDB
- Chapter upload date for Maofly
- Replace Duktaping [JS `LZString`](https://github.com/pieroxy/lz-string) with [native `LZString`](https://github.com/rufushuang/lz-string4java) for Maofly
- Add rate limit (2 per second)

Checklist:

- [ ] Updated `extVersionCode` value in `build.gradle` for individual extensions
- [x] Updated `overrideVersionCode` or `baseVersionCode` as needed for all multisrc extensions
- [ ] Referenced all related issues in the PR body (e.g. "Closes #xyz")
- [ ] Added the `isNsfw = true` flag in `build.gradle` when appropriate
- [x] Have not changed source names
- [ ] Have explicitly kept the `id` if a source's name or language were changed
- [x] Have tested the modifications by compiling and running the extension through Android Studio

PR #5679, opened 2021-02-08
[SKIP CI] Contributing update for multisrc

AriaMoradi:
I guess it's enough for an introduction, I'll add more in future PRs.

PR #12244, opened 2022-06-19
[skip ci] Fix additional gradle code path for multisrc themes

stevenyomi:
Replaces all occurrences of `additional.gradle.kts` with `additional.gradle`.

The reason is that additional gradle code is merged into a normal gradle file in Groovy. The formatter is unhappy with Groovy code in a Kotlin DSL file and will throw `ParseException` (without printing the filename 💢), preventing #12237 from building.

Looks like this is the first usage of additional gradle code - there's currently no `additional.gradle[.kts]` files in the repo.


stevenyomi:
This feature was added in #5845 and docs in #5679. (for reference)

PR #12246, opened 2022-06-19
Mangakawaii: Fix pagination displayed as manga

nzoba:
The pagination appeared as if each page number was a manga when searching, so I fixed the selector and added the one for the next pages. The covers were also loading hella slow so I increased the rate-limit value by 1, I hope this won't affect the source too much

Checklist:


arkon:
> The covers were also loading hella slow so I increased the rate-limit value by 1, I hope this won't affect the source too much

You could potentially rate limit them separately (based on domain/regex/etc.), but we can revisit that if you see issues with it with the new rate.

PR #12252, opened 2022-06-20
[Manga Plus] Add URL intent for chapter links

lord-ne:
When clicking a Manga Plus chapter links (e.g. https://mangaplus.shueisha.co.jp/viewer/1013489), open the corresponding series in Tachiyomi.

Checklist:


src/all/mangaplus/src/eu/kanade/tachiyomi/extension/all/mangaplus/MangaPlusUrlActivity.kt
```diff
@@ -14,13 +14,18 @@ class MangaPlusUrlActivity : Activity() {
 
         val pathSegments = intent?.data?.pathSegments
         if (pathSegments != null && pathSegments.size > 1) {
-            val titleId = if (!pathSegments[1].equals("sns_share")) pathSegments[1] else
-                intent?.data?.getQueryParameter("title_id")
+            val query: String = if (pathSegments[0].equals("viewer")) {
```
alessandrojean, line 17: Could rewrite this using a `when` block instead.

src/all/mangaplus/src/eu/kanade/tachiyomi/extension/all/mangaplus/MangaPlus.kt
```diff
@@ -139,7 +139,29 @@ abstract class MangaPlus(
...
+    private fun titleIdParse(response: Response): Int {
+        val result = response.asMangaPlusResponse()
+
+        checkNotNull(result.success) { result.error!!.langPopup(langCode).body }
+
+        return result.success.mangaViewer!!.titleId!!
+    }
+
     override fun fetchSearchManga(page: Int, query: String, filters: FilterList): Observable<MangasPage> {
+        if (query.startsWith(PREFIX_CHAPTER_ID_SEARCH) && query.matches(CHAPTER_ID_SEARCH_PATTERN)) {
```
alessandrojean, line 160: Alternatively to using Observables and adding another check to the fetch search method, you could just do this check in the search request method, making it call the viewer endpoint instead, and then in the search parse you can check if the response.request.url is from the viewer and do another call inside the parse.

src/all/mangaplus/src/eu/kanade/tachiyomi/extension/all/mangaplus/MangaPlus.kt
```diff
@@ -139,7 +139,29 @@ abstract class MangaPlus(
...
+            titleIdParse(response)
+        }
+    }
+
+    private fun titleIdParse(response: Response): Int {
+        val result = response.asMangaPlusResponse()
+
+        checkNotNull(result.success) { result.error!!.langPopup(langCode).body }
+
+        return result.success.mangaViewer!!.titleId!!
```
alessandrojean, line 156: What if the chapter expired? I think the API will not return the titleId in this case, so you do need to check if it's null or it's valid and throw an exception if it's not the case.

src/all/mangaplus/src/eu/kanade/tachiyomi/extension/all/mangaplus/MangaPlus.kt
```diff
@@ -434,5 +459,7 @@ abstract class MangaPlus(
 
         const val PREFIX_ID_SEARCH = "id:"
         private val ID_SEARCH_PATTERN = "^id:(\\d+)$".toRegex()
+        const val PREFIX_CHAPTER_ID_SEARCH = "ch_id:"
```
alessandrojean, line 462: Also could maybe use a more readable argument, like `chapterId` or `chapter-id`

src/all/mangaplus/src/eu/kanade/tachiyomi/extension/all/mangaplus/MangaPlus.kt
lord-ne: I don't have any expired chapter links lying around, but I tested with a nonexistent link (https://mangaplus.shueisha.co.jp/viewer/1013490), and it just returns "No results found" as expected.

It throws a `NullPointerException` if `titleId` is null since we do `!!`. If you have a more specific exception in mind, I can check and throw that instead. I wasn't sure how much we needed to worry about specific error messages.

src/all/mangaplus/src/eu/kanade/tachiyomi/extension/all/mangaplus/MangaPlus.kt
lord-ne: I think ch is a common enough abbreviation for "chapter" that we don't need to worry about it

src/all/mangaplus/src/eu/kanade/tachiyomi/extension/all/mangaplus/MangaPlusUrlActivity.kt
lord-ne: Sure, I can do that

src/all/mangaplus/src/eu/kanade/tachiyomi/extension/all/mangaplus/MangaPlus.kt
lord-ne: That also works, but tbh I find the current approach a little cleaner.

Another possibility is to do the call to the viewer API, then just return an SManga with only the url and title set, and initialized set to false. Then, fetchMangaDetails will be automatically called (I think), so we don't have to worry about it.

alessandrojean: No problems really, is just that Observables are a bit weird to understand at first, so I try to always avoid using it when I can. It was just a suggestion.

src/all/mangaplus/src/eu/kanade/tachiyomi/extension/all/mangaplus/MangaPlus.kt
alessandrojean: Could check if it's null with the `checkNotNull` and throw a simple "Chapter expired" exception.

src/all/mangaplus/src/eu/kanade/tachiyomi/extension/all/mangaplus/MangaPlusUrlActivity.kt
alessandrojean: Can you try to check if the deep link is still working? I recall sometimes using Kotlin features in the activity classes in the extensions sometimes causes crashes.

lord-ne: I checked the deep links (for both series links and chapter links) and it works

alessandrojean: Great! I will take a further look and review it one more time soon and then it should be okay to be merged 👍🏻

src/all/mangaplus/src/eu/kanade/tachiyomi/extension/all/mangaplus/MangaPlus.kt
lord-ne: Added

PR #12111, opened 2022-06-07
Manga Pro cover and chapter list bugs fixed

Saud-97:
Fixes #11881 and fixes #11809

Checklist:

- [ ] Updated `extVersionCode` value in `build.gradle` for individual extensions
- [x] Updated `overrideVersionCode` or `baseVersionCode` as needed for all multisrc extensions
- [x] Referenced all related issues in the PR body (e.g. "Closes #xyz")
- [ ] Added the `isNsfw = true` flag in `build.gradle` when appropriate
- [x] Have not changed source names
- [ ] Have explicitly kept the `id` if a source's name or language were changed
- [x] Have tested the modifications by compiling and running the extension through Android Studio

PR #12321, opened 2022-06-25
GigaViewer: show all chapters, fix Corocoro, cleanup

stevenyomi:
- Show all chapters (including paid/locked)
  - so that chapters downloaded during limited free and those already paid can be accessed afterwards
  - existing downloads (if previously hidden) need to be renamed to be detected (prepend title with emoji plus space)
  - emojis: [paid](https://emojipedia.org/yen-banknote/) or [locked](https://emojipedia.org/locked/)
- Compress the descrambed image to JPEG with quality 90
  - the source is already lossy, low-res and scrambled
  - not worth saving as big, "lossless" PNG files
- Fix Corocoro manga list parsing
- Cleanups
- Optimize icon assets

Checklist:


alessandrojean:
> existing downloads (if previously hidden) need to be renamed to be detected (prepend title with emoji plus space)

I am kinda against this to be honest, if the user will already have to rename stuff, it's just better it just use the local source in this case instead. Showing paid chapters (even if they can't be read) is confusing and eventually will only lead to users asking how they can unlock stuff (as happened in the past with other extensions). Would be better to really only show the free to read chapters.

stevenyomi:
> if the user will already have to rename stuff, it's just better it just use the local source in this particular case instead.

They don't have to. They can still keep to their local source for old chapters.

For new chapters onwards, the app will automatically rename and sync the downloaded folder, so there's no more copying from online source folder to local source folder. I think it is worth keeping the locked chapters for the convenience (no need to mess with local source).

Moreover, they can still manually copy to local source if they wish.

> lead to users asking how they can unlock stuff (as happened in the past with other extensions).

I don't find issues on this when doing a search. It can be easy as replying "wait for them to be free or pay for them" and closing the issue. Is there any example of that kind of issue?

> Would be better to really only show the free to read chapters as the vast majority of extensions with paid content does.

I did [search](https://github.com/tachiyomiorg/tachiyomi-extensions/search?q=DD12) the repo for lock icon and found several of them.

> could at least leave it a JPEG 100% maybe?

Check "Selecting a Quality Level" section in https://hacks.mozilla.org/2014/08/using-mozjpeg-to-create-efficient-jpegs/. The quality loss is way lower than what you've expected. Usually 75~80 is suggested and 90 is already high enough. IIRC quality 100 JPEG is almost no compression.

Also the images are low-res and very blurry. Scrambling plus JPEG introduces even more artifacts along the edges. This is simply not a good source for high-quality manga images.

Saud-97:
> Show all chapters (including paid/locked)
> so that chapters downloaded during limited free and those already paid can be accessed afterwards

Isn't better to only show accessible chapters to user? Hence if a user has paid he can login throw WebView and then refresh the manga to access paid chapters? No need to show a non-paying user that there are paid chapters unavailable for downloads.

stevenyomi:
> Isn't better to only show accessible chapters to user? No need to show a non-paying user that there are paid chapters unavailable for downloads.

Sometimes chapters are set to be free for a limited time; for some sites, the latest N chapters are free. Users might download them when they are free and read them afterwards. At that time, the chapter might be locked/paid, thus hidden and inaccessible, so users have to manually open the downloaded files or copy to local source. Showing all chapters make downloaded chapters always accessible to make things more convenient.

> Hence if a user has paid he can login throw WebView and then refresh the manga to access paid chapters?

Another problem is that we don't know how the webpage looks like if user really paid for the chapter. Even if the user paid, current implementation might still hide those chapters and show only the free ones.

Another insignificant benefit is that you get to see the complete history of the chapters (chapter numbers, titles, dates, etc) instead of broken pieces.

alessandrojean:
I am okay in merging this but it should be better to indicate in the chapter.url the chapter is locked (maybe by appending an `&locked=true` or something similar, and then in the pageListRequest or fetchPageList throw an error to avoid calling the website/API in this case. If the user have it downloaded, the app doesn't do the request and priorize the download IIRC (but need to be tested).

stevenyomi:
> it should be better to indicate in the chapter.url the chapter is locked (maybe by appending an `&locked=true` or something similar), and then in the pageListRequest or fetchPageList throw an error to avoid calling the website/API in this case.

Just in case someone bought the chapter, the request should be made. If it's not accessible, the response will parse to an empty list anyway. I can make it throw an exception "Chapter is not available!" though.

> I think the user should be forced to read in local source at their own responsability and the extension indeed should not show them if they're not free anymore

The downloaded images is indeed local, just that you can read it in the list. It's similar to recording a TV program and watching it afterwards (just for yourself), which is perfectly legal. It's just a more convenient interface. If there were an easy function to transfer downloaded chapters to local source in one tap, we can indeed show free ones only.

Showing them does not mean they are accessible - the original website also shows them. We are not abusing any APIs anyway.

Saud-97:
If the paid chapter pageList is having a unique indication that its paid (like having a redirection). You can filter out any paid or locked chapters via that. Similar to what I did in #12111. I filtered out non free protected chapters.

alessandrojean:
> I can make it throw an exception "Chapter is not available!" though.

If you can, would be better, yes. Potentially in Japanese maybe, as there's only Japanese sources in Gigaviewer.

> Showing them does not mean they are accessible - the original website also shows them. We are not abusing any APIs anyway.

Indeed, you're right on this. We just don't know if people from the corporates that search for this will think like this, but should be okay for now.

PR #12332, opened 2022-06-26
Copymanga: Fix SSL error while cover loading, add rate limit preference

constfold:
I met the cover loading issue on about half of manga, the error message shows it's caused by SSL certificate(I have no idea how it can). Re-introduce the SSL ignore seems fixed this problem.
`<-- HTTP FAILED: javax.net.ssl.SSLHandshakeException: java.security.cert.CertPathValidatorException: Trust anchor for certification path not found.`

Could you please review? @stevenyomi 

By the way, there are still some issues like chapter's upload-date got refresh to now when every new chapter in.
Checklist:


stevenyomi:
I'm trying to rewrite the extension with mobile H5 APIs and see how that affects loading speed.

stevenyomi:
Have you enabled this preference? This seems to be redirecting cover images to an outdated host:

https://github.com/tachiyomiorg/tachiyomi-extensions/blob/da995033e0a57a5dc59a8ade839eab1ab766c5b0/src/zh/copymanga/src/eu/kanade/tachiyomi/extension/zh/copymanga/CopyManga.kt#L102-L105

constfold:
> Have you enabled this preference? This seems to be redirecting cover images to an outdated host:

The regex itself is also outdated, so this couldn't have any effect IMO. 
In fact the cover url server responsed points to `mangafuna.xyz`(which I have no idea why) before I added the SSL back. After changes in PR, I can got the updated image url same as visit it with browser (`hi.est152.com`).

zhongfly:
> I'm trying to rewrite the extension with mobile H5 APIs and see how that affects loading speed.

I have completed an extension that uses the official app api. According to my research, it uses the same api as the mobile h5 version, only the headers are different.
https://github.com/zhongfly/tachiyomi-extensions/tree/copymangas/src/zh/copymangas

alessandrojean:
> the SSL ignore seems fixed this problem.

JFYI this is not safe and it's not recommended. The certificate extension is in general always something related to their HTTPS certificate that they need to fix it.

stevenyomi:
> I am very concerned that if the api is widely used it will lead to an escalation of the cat-and-mouse game with the website

Nope, there's not much difference between web page and API. In fact serializing JSON might even be faster than rendering an HTML. I'll try to organize the code.

constfold:
prefer #12338

Issue #9096, opened 2021-09-17
您好，請停止解析 CopyManga 並下架 tachiyomi-extensions 裡 copymanga 代碼


hotmanga:
### Source information

CopyManga.kt

### Steps to reproduce

您好，尊敬的 Tachiyomi 開發者。

我們無意中發現了 Tachiyomi 這款開源產品，覺得十分的優秀。裡面的代碼結構和編寫方式讓我們由衷的喜歡和羨慕。

但目前我們在閱讀源代碼時遭遇到一些情況，經受 CopyManga 的委託，特來與您方協商一下。

經過閱讀代碼我們發現，在您程序代碼 CopyManga.kt 中發現正在解析 CopyManga 的頁面和API接口，如

- /comics
- /api/v3/search/comic
- /comic/$comicPathWord/chapter/${chapter.getString("id")}
- /comic/${obj.getString("path_word")}

希望 Tachiyomi 方可以下架該功能並刪除這些代碼。講這些代碼共享開源出來之後，給 CopyManga 帶來了大量的工作阻礙。

我們發現 Tachiyomi 是有一款優秀的產品，有著十分廣泛的用戶，故大量用戶在頻繁訪問 CopyManga 時，也給服務器帶來了不小的影響。

我們十分喜歡和欣賞 Tachiyomi 這款產品，但還是希望您方可以下架該功能。

再次感謝。

### Expected behavior

希望可以下架該功能。

### Actual behavior

希望可以下架該功能。

### Tachiyomi version

Tachiyomi 0.12.2

### Android version

Tachiyomi 0.12.2

### Acknowledgements

- [X] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open issue.
- [X] I have written a short but informative title.
- [X] I have updated the app to version **[0.12.2](https://github.com/tachiyomiorg/tachiyomi/releases/latest)**.
- [X] I have updated all installed extensions.
- [X] I have tried the [troubleshooting guide](https://tachiyomi.org/help/guides/troubleshooting/).
- [X] If this is an issue with the app itself, I should be opening an issue in the [app repository](https://github.com/tachiyomiorg/tachiyomi/issues/new/choose).
- [X] I will fill out all of the requested information in this form.

arkon:
It doesn't look like the website is for a scanlator nor an official one.

We'll add some aggressive rate limiting to help alleviate pressure.

whghjyh:
[...]

PR #12338, opened 2022-06-27
CopyManga: rewrite

### LINKED ISSUE #12320: copymanga : Extremely slow and unsteady loading speed
### Source information

copymanga 1.3.28

### Steps to reproduce

1.open the source
2.the interface circling for a long time and much probably the error 504 take place
3.select a random manga
4.cover image can't loaded(or just in a extremely slow speed) , and when the chapters was loaded , the error 502 often take place

### Expected behavior

the error 502 and error 504 shouldn't take place frequently , and the covers and chapters should be loaded successfully

### Actual behavior

the error 502 and error 504 take place frequently , and the covers and chapters can't be loaded successfully the most time

### Tachiyomi version

0.13.4

### Android version

Android 12

### Other details

_No response_

### Acknowledgements

- [X] I have updated the app to version **[0.13.4](https://github.com/tachiyomiorg/tachiyomi/releases/latest)**.

### LINKED ISSUE #12346: CopyManga: cannot load, update or read any manga from it
### Source information

CopyManga 1.3.28

### Steps to reproduce

1.Install this extension
2.search some mangas
3.select the manga from CopyManga and view details

### Expected behavior

normally display the introduction and chapter list

### Actual behavior

load name, cover and introduction normally, but cannot load chapter list, and shows:
"Attempt to invoke virtual 
method 'java.lang.String
org.jsonp.nodes.Element.data()'on a
 null object reference"

### Tachiyomi version

0.13.4

### Android version

11

### Other details

MIUI 12.5.6 based on Android 11

### Acknowledgements

- [X] I have updated the app to version **[0.13.4](https://github.com/tachiyomiorg/tachiyomi/releases/latest)**.

stevenyomi:
Supercedes #12332

- Rewrite using APIs, closes #12346
  - no longer depends on org.json
  - better filters
  - better parsing
  - better chapter lists
  - better preferences
    - add option to prefer WebP images, reducing site load
    - add option to change domain
    - all preferences take effect without restart
- Changes to rate limit, fixes #12320
  - Change rate limit from 1 per 2 seconds to 2 per 4 seconds
  - Make rate limit interceptor nonblocking in https://github.com/tachiyomiorg/tachiyomi/pull/7389 and integrate that into the extension
- Update icons
- Exclude unused dictionary files in charater converter library to reduce size

Checklist:


zhongfly:
The rate limit is too strict, because checking for an update of a comic requires at least 2 requests, so it takes 4 seconds to update a comic at the fastest, which is too slow.
I think it needs to be changed to a preference that the user can modify

stevenyomi:
> The rate limit is too strict, because checking for an update of a comic requires at least 2 requests, so it takes 4 seconds to update a comic at the fastest, which is too slow. I think it needs to be changed to a preference that the user can modify

Regarding the math, the new limit allows 2 requests to be executed at the same time, so ideally it would not take 4 seconds. As you know, some titles have multiple chapter groups, and the app might re-download the thumbnail, so it might take 3~5 requests. Considering that network requests are already quite slow, 4+ seconds is very acceptable.

The rate limit was added to close #9096, where the site owner said they were unhappy with the extension. Traffic from Tachiyomi is useless for them, increasing site load while not producing ad clicks. Therefore, for a third-party aggregator website (non-official or licensed), rate limit is rather necessary to relieve pressure on them, so we should not add such preference unless the website is really inaccessible. According to my tests, after the fix the load speed is reasonable and acceptable.

stevenyomi:
By the way, I cached the chapter group information locally, so for updates it will take just that amount (1~3 requests).

zhongfly:
There is one more question: I found that you modified the image file in every pr. Is it something that your development tool does automatically? Because Git can not diff birnay file, it stores every single version of every single binary file within the repository, which also makes the git data larger and larger.
I think it would be better to only adjust the picture when necessary.

stevenyomi:
In #12226 I optimized the old, blocky icons and in this PR I replaced it with the app icons. It won't take up too much space. If you find your clone too big, it's most likely because git fetched objects from the `repo` branch and all other unused extensions. Follow the steps that I wrote [here](https://github.com/tachiyomiorg/tachiyomi-extensions/blob/master/CONTRIBUTING.md#cloning-the-repository) and you will get a smaller repo.

PR #12354, opened 2022-06-28
[skip ci] Send a bit more purge requests to jsDelivr

stevenyomi:
Send more purge requests to purge the cache.


alessandrojean:
@arkon Should it be okay to be merged?

stevenyomi:
Looks like jsDelivr purge requests are throttled to some extent. Furthermore, it works by chance and doesn't always purge the cache. I still hope that some GitHub proxy will be allowed, or index will be hosted on Netlify if possible. (related to https://github.com/tachiyomiorg/tachiyomi/pull/7400)

PR #12371, opened 2022-06-29
Ganma: cleanup for chapter list and preferences

stevenyomi:
Checklist:


alessandrojean:
Curiosity: any particular reason to switch from data classes to normal classes in the DTO?

stevenyomi:
> Curiosity: any particular reason to switch from data classes to normal classes in the DTO?

It's just that I wasn't using any data class features and using normal classes may reduce compiled code size. (stripping the auto-generated `toString`, `equals`, etc)

src/ja/ganma/src/eu/kanade/tachiyomi/extension/ja/ganma/Ganma.kt
```diff
@@ -75,7 +75,7 @@ open class Ganma : HttpSource(), ConfigurableSource {
 
     override fun fetchMangaDetails(manga: SManga): Observable<SManga> =
         client.newCall(realMangaDetailsRequest(manga)).asObservableSuccess()
-            .map { mangaDetailsParse(it).apply { initialized = true } }
+            .map { mangaDetailsParse(it) }
```
alessandrojean, line 78: I think it's better to keep the `initialized = true` as the app uses it to some internal checks too.

Reference of the base `HttpSource` class: https://github.com/tachiyomiorg/tachiyomi/blob/master/app/src/main/java/eu/kanade/tachiyomi/source/online/HttpSource.kt#L172-L184

stevenyomi: I moved it to `toSMangaDetails()` in `mangaDetailsParse()`.

PR #12413, opened 2022-07-02
Add 6Manhua

stevenyomi:
Responses from this site's server cause glitch in the emulator's socket and make okhttp throw `unexpected end of stream`. It works correctly on devices.

Checklist:


src/zh/sixmh/build.gradle
```diff
@@ -0,0 +1,16 @@
...
+    extName = '6Manhua'
+    pkgNameSuffix = 'zh.sixmh'
+    extClass = '.SixMH'
+    extVersionCode = 1
+}
+
+apply from: "$rootDir/common.gradle"
+
+dependencies {
+    implementation 'com.github.stevenyomi:unpacker:1.0'
```
arkon, line 15: Can you point to a specific SHA or publish to Maven Central? Replacing a tag release is possible on JitPack, so I'd like to at least avoid that possibility whenever possible.

stevenyomi: Done. It's worth mentioning that you didn't let me do this in https://github.com/tachiyomiorg/tachiyomi-extensions/pull/12237#discussion_r901148661 (although I happened to change it to maven in #12254).

PR #12423, opened 2022-07-03
SinMH: cleanup and extract lib

stevenyomi:
Checklist:


multisrc/build.gradle.kts
```diff
@@ -28,6 +28,7 @@ configurations {
 
 dependencies {
     compileOnly(libs.bundles.common)
+    compileOnly("com.github.stevenyomi:unpacker:12a09e3c1a") // 1.1, for SinMH
```
vetleledaal, line 31: don't think this is desirable, especially given that it applies to all multisrc extensions

stevenyomi: This is only used by the theme generators. The generated sources/gradle files will not contain this.

alessandrojean: Would be better to use the additional Gradle files, I also think it's not necessary to add it as a project dependency, even if it's only used by the generators.

stevenyomi: The additional gradle files are placed in `overrides` folder and will only take effect in `generated-src`. I'm using the library in the theme base class, and not adding this as project dependency will prevent the generators from working (the compiler will complain that it can't find the library).

alessandrojean: That's weird, it should work fine as the files are just copied to the `generated-src` and aren't executed IIRC. Something to be investigated definitely.

@arkon Should be fine to merge, right?

stevenyomi: This is because currently these base classes are also considered source files of the generator module. Probably need to exclude them but retain the theme generator classes as sources.

alessandrojean: Definitely weird. Anyway, something that could be improved further as it seems then.

alessandrojean: Can you at least add the library in the Dependency Catalog TOML file as the other dependencies?

stevenyomi: There's already an extension using an old version of this so I don't know if it will cause conflict. Also I'm not sure if arkon would be happy with this small library being added, see https://github.com/tachiyomiorg/tachiyomi-extensions/pull/12413#discussion_r912422271.

alessandrojean: It shouldn't cause any conflict I think. But let's wait for the arkon opinion on this.

arkon: It's definitely an unfortunate gap that you can't really add dependencies for the main generators without leaking into all of the other ones. I assume it'd be unreasonable to put any usage of the library in the overrides?

stevenyomi: I removed the denpendency from the generator class. It's a pity to have to copy the same piece of code here and there though.

PR #12470, opened 2022-07-07
Bato.to: add mirrors

vetleledaal:
Only bato.to and batoo.com resolve to the same IP, others do not. mycdhands.com was removed as it doesn't resolve to an IP.

Checklist:


stevenyomi:
Version code should be 1.3.x since 1.3.24 in the changelog.

PR #12595, opened 2022-07-16
Add "Hdtoday.cc" extension:))) thanks

IanSerrano1:
Add "Hdtoday.cc" extension:))) thanks

for movies cuz the other for movies sucks


loocool2:
We're not a video player.

PR #12676, opened 2022-07-21
Madara: Update User-Agent string

toqueteos:
Firefox 78 was released 2020-06-30.
If we keep the User-Agent string somewhat updated requests should be harder to detect & block by CloudFlare.

The randomizer seems to be unnessary nowadays.
Firefox doesn't bother to send the patch version back anyway.

It's easy to check with HTTPBin for example.
Navigating to https://httpbin.org/headers with Firefox v102.0.1 on Windows returns:
```json
{
  "headers": {
    "Accept": "application/json",
    "Accept-Encoding": "gzip, deflate, br",
    "Accept-Language": "es-ES,es;q=0.8,en-US;q=0.5,en;q=0.3",
    "Host": "httpbin.org",
    "Referer": "https://httpbin.org/",
    "Sec-Fetch-Dest": "empty",
    "Sec-Fetch-Mode": "cors",
    "Sec-Fetch-Site": "same-origin",
    "User-Agent": "Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:102.0) Gecko/20100101 Firefox/102.0",
    "X-Amzn-Trace-Id": "Root=1-62d985e1-2b51fc6437de34c433e1b902"
  }
}
```

Checklist:


arkon:
I think the randomization was in place because some sites liked to just block valid UA agents too as their janky anti-scraping mechanism.

toqueteos:
> I think the randomization was in place because some sites liked to just block valid UA agents too as their janky anti-scraping mechanism.

Good point, I'll leave that in.

PR #12678, opened 2022-07-21
[skip ci] Add source language to issue template

stevenyomi:
This will make it easier to search for issues.


Riztard:
maybe also add is NSFW?

arkon:
Probably not that necessary, unless contributors have felt a need for that.

PR #12775, opened 2022-07-29
HentaiVN: update to new base domain

mechaman122:
* Update base domain
* Update extVersionCode to 13


hoang-himself:
Duplicate of #12760

PR #12030, opened 2022-06-01
[RU]Multi Lib (MangaLib and HentaiLib)

e-shl:
Checklist:

- [x] Updated `extVersionCode` value in `build.gradle` for individual extensions
- [x] Updated `overrideVersionCode` or `baseVersionCode` as needed for all multisrc extensions
- [x] Referenced all related issues in the PR body (e.g. "Closes #xyz")
- [x] Added the `isNsfw = true` flag in `build.gradle` when appropriate
- [x] Have not changed source names (*partially, doesn't seem to affect)
- [x] Have explicitly kept the `id` if a source's name or language were changed
- [x] Have tested the modifications by compiling and running the extension through Android Studio

PR #12806, opened 2022-07-30
[RU]GroupLe tags & filters

e-shl:
Checklist:


arkon:
Feels like you could move a lot of the duplicated things into the parent class at some point.

e-shl:
@arkon filter lists are different, and some sources do not have them at all. But problem with rewriting/overrides filters wrote here https://github.com/tachiyomiorg/tachiyomi-extensions/pull/12030

> How to make work filters from the overrides? Filters are replaced / built into the list, but requests from them do not go(Only from overrides).

PR #12838, opened 2022-08-01
Remove Bainian Manhua: site is down

stevenyomi:
Will wait for a week to see if it's completely dead.

URLs to check: https://www.bnman.net/ and https://www.mhqj.com/


alessandrojean:
Merging, but can be readded later if their website goes online again. The same applies to the other PR.

PR #12911, opened 2022-08-08
Remove Koushoku

MatchaSoba:
Refer to [site announcement](https://koushoku.org/2022.08.07.txt). Site will no longer be updated, and owner wants people to use the site directly.
Users should seek alternative sources and uninstall this extension.

Checklist:


arkon:
We can remove it once it's dead. It's still functional afaik.

chauuun:
Hmm frankly I don't quite agree. With how severely rate limited the extension is, and how strict the Cloudflare on the site has become, on the rare occasions that the extension works, its frequently buggy and way too slow to use comfortably. I'd say its quite pointless to send additional traffic from Tachiyomi to their servers when the end result for users is so suboptimal. Not to mention increasing their server costs like this will likely cause them to shut down earlier.

MatchaSoba:
Hmm frankly I'm not sure if it is even functional. With how severely rate limited the extension is, and how strict the Cloudflare on the site is, on the rare occasions where the extension works, its frequently buggy and way too slow to use comfortably. I'd say its pointless to send traffic from Tachiyomi to their servers if the end result for users is so suboptimal. Not to mention adding to their server costs like this will likely cause them to shut down earlier.

PR #13083, opened 2022-08-19
Bilibili (Multisrc): Cleanup

AntsyLich:
Checklist:


multisrc/overrides/bilibili/bilibilicomics/src/BilibiliComicsFactory.kt
```diff
@@ -45,13 +45,13 @@ abstract class BilibiliComics(lang: String) : Bilibili(
 ) {
 
     override val client: OkHttpClient = super.client.newBuilder()
-        .addInterceptor(::signedInIntercept)
-        .addInterceptor(::expiredTokenIntercept)
-        .rateLimitHost(baseUrl.toHttpUrl(), 1)
-        .rateLimitHost(CDN_URL.toHttpUrl(), 2)
-        .rateLimitHost(COVER_CDN_URL.toHttpUrl(), 2)
+        .also { it.interceptors().add(0, Interceptor { chain -> signedInIntercept(chain) }) }
```
stevenyomi, line 48:
```suggestion
        .apply { interceptors().add(0, Interceptor { chain -> signedInIntercept(chain) }) }
```
Better use `apply` because you are changing that thing. `also` is best used to log things.

PR #13130, opened 2022-08-22
NHentai: set User-Agent to get past Cloudflare

### LINKED ISSUE #12784: Nhentai Failed to bypass Cloudfare

Nhentai 1.3.35

### Steps to reproduce

1.Failed to Bypass Cloudfare 
2.Webview doesn't work properly (all white and letters) 

### Expected behavior

It should open properly and webview is looks exactly as it should be on website 

### Actual behavior

It didn't open properly and fail and I've also tried all sorts of things to fix it but it still didn't work 

### Tachiyomi version

0.13.5

### Android version

Android 6.0.1

### Other details

Screenshot_20220729_220243

Screenshot_20220729_214223

### Acknowledgements

- [X] I have updated the app to version **[0.13.5](https://github.com/tachiyomiorg/tachiyomi/releases/latest)**.

vetleledaal:
Closes #12784

Checklist:


juniorjacob:
I don't think I am the only one who believes this to be a terrible idea but I want to ensure the note is put here. This would make their job of cat and mouse even more manageable and they have proven to watch the default user agents so why would this be any different?! Just let users change their user agent to a non-default option so that this issue is manageable. Otherwise, you would be better off changing this to utilize a list of updated and popular user agents and taking a random agent to use when accessing NHentai so that this specific extension reaps the benefits of a well-thought-out solution.

vetleledaal:
I don't see much of any back and forth in the git history in regards to NHentai, just regular changes to logic that is to be expected. If they wanted to target Tachiyomi specifically there are better ways to do it than through the User-Agent.

Changing the User-Agent is just intended to get past Cloudflare's IUAM. I do however agree that this will overwrite any custom User-Agent set by the user, but there's no reasonable way for extensions to know if it has been modified.

PR #13153, opened 2022-08-23
New source: Central Novel

Claudemirovsky:
Checklist:


alessandrojean:
Tachiyomi is meant to be a manga reader, you would be better handling this by submitting a PR to add this source to a light novel reader app like Shosetsu or LN Reader instead of handling the image creation as this source is purely based on novels only, so denying this. Also pretty much of the extension code is duplicate as the website uses the MangaThemesia theme which we have a multisrc generator.

We appreciate your time though.

PR #13163, opened 2022-08-24
Animepahe Audio types

Kidinu:
allow Animepahe to show language audio type.
For example:
Naruto episode 1[sub][dub].


CarlosEsco:
This isn't aniyomi

PR #13240, opened 2022-08-27
Manga18fx: add support for genre filter

AriaMoradi:
Checklist:


multisrc/overrides/madara/manga18fx/src/Manga18fx.kt
```diff
@@ -53,8 +58,16 @@ class Manga18fx : Madara(
...
-        GET("$baseUrl/search?q=$query&page=$page", headers)
+    override fun searchMangaRequest(page: Int, query: String, filters: FilterList): Request {
+        if (query.isEmpty()) {
+            filters.forEach { filter ->
+                if (filter is GenreFilter)
+                    return GET(filter.vals[filter.state].second, headers)
+            }
+        }
+
+        return GET("$baseUrl/search?q=$query&page=$page", headers)
```
alessandrojean, line 71: This could lead to some issues with special characters, so maybe it's better to build the URL with a proper class that handles these special cases.

```kotlin
val url = "$baseUrl/search".toHttpUrl().newBuilder()
    .addQueryParameter("q", query)
    .addQueryParameter("page", page.toString())
    .toString()

return GET(url, headers)
```

Issue #13106, opened 2022-08-20
Komik Cast: Pagination broken


Subadrunz:
### Source information and language

Komik Cast 1.3.29 (Indonesia)

### Steps to reproduce

1. Click update in komikcast extension.
2. Scroll down.
3. And its get looping 

### Expected behavior

normal as usual 

### Actual behavior

Extension is looping, and idk What to do. 

### Tachiyomi version

0.13.6

### Android version

Android 11

### Other details

![IMG_20220820_192426](https://user-images.githubusercontent.com/107401290/185746173-088d9937-2176-4c2e-88b2-73cec0d3b15a.jpg)
![IMG_20220820_192530](https://user-images.githubusercontent.com/107401290/185746175-09d200a8-6d6e-48fb-9009-026dcc0c046d.jpg)
![IMG_20220820_192703](https://user-images.githubusercontent.com/107401290/185746177-660fa9c4-32cf-4a45-83c4-e55b134e05b4.jpg)

### Acknowledgements

- [X] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open issue.
- [X] I have written a short but informative title.
- [X] I have updated the app to version **[0.13.6](https://github.com/tachiyomiorg/tachiyomi/releases/latest)**.
- [X] I have updated all installed extensions.
- [X] I have tried the [troubleshooting guide](https://tachiyomi.org/help/guides/troubleshooting/).
- [X] If this is an issue with the app itself, I should be opening an issue in the [app repository](https://github.com/tachiyomiorg/tachiyomi/issues/new/choose).
- [X] I will fill out all of the requested information in this form.

happywillow0:
Notes for whoever works on this:
Komik Cast uses the /page/# system rather than the ?page=# system on the search / filter pages.

arispati:
PR for this has been merged, @Subadrunz please try the latest version. if solved, please close this issue.

Subadrunz:
Thank you :)

PR #13280, opened 2022-08-31
Komikcast: fix broken pagination

arispati:
FIX: latest update pagination and popular page #13106 

Checklist:


multisrc/overrides/mangathemesia/komikcast/src/KomikCast.kt
```diff
@@ -45,6 +46,16 @@ class KomikCast : MangaThemesia(
...
 
+    override fun popularMangaRequest(page: Int) = customPageRequest(page, "orderby", OrderByFilter("popular"))
+    override fun latestUpdatesRequest(page: Int) = customPageRequest(page, "sortby", OrderByFilter("update"))
+
+    private fun customPageRequest(page: Int, filterKey: String, filter: OrderByFilter): Request {
+        val url = baseUrl.toHttpUrl().newBuilder().addPathSegment(mangaUrlDirectory.substring(1))
+        val onPage = if (page > 1) "page/$page/" else ""
+        url.addPathSegments(onPage).addQueryParameter(filterKey, filter.selectedValue())
+        return GET(url.toString())
+    }
```
alessandrojean, line 57: You can simplify the code by using Strings directly here.

Also, there's no need to use `toHttpUrl` as the values are simple and doesn't depend on encoding.

```suggestion
    override fun popularMangaRequest(page: Int) = customPageRequest(page, "orderby", "popular")
    override fun latestUpdatesRequest(page: Int) = customPageRequest(page, "sortby", "update")

    private fun customPageRequest(page: Int, filterKey: String, filterValue: String): Request {
        val pagePath = if (page > 1) "/page/$page/" else ""
        
        return GET("$baseUrl$mangaUrlDirectory/$pagePath?$filterKey=$filterValue", headers)
    }
```

multisrc/overrides/mangathemesia/komikcast/src/KomikCast.kt
```diff
@@ -8,6 +8,7 @@ import kotlinx.serialization.decodeFromString
 import kotlinx.serialization.json.jsonArray
 import kotlinx.serialization.json.jsonObject
 import okhttp3.Headers
+import okhttp3.HttpUrl.Companion.toHttpUrl
```
alessandrojean, line 11:
```suggestion
```

Can be removed as it's not used.

alessandrojean (changes requested):

alessandrojean (approved):

PR #13394, opened 2022-09-08
Remove all custom rate limit interceptors

stevenyomi:
Removed all custom rate limit interceptors since the app provided one can already handle canceled requests (https://github.com/tachiyomiorg/tachiyomi/pull/7389) since v0.13.5.

MangaDex's custom interceptor is also removed due to this (and the cover pattern has changed, too).

A rewritten one (to be merged into a future release) will handle cached requests. (https://github.com/tachiyomiorg/tachiyomi/pull/7889)

Checklist:


src/all/mangadex/src/eu/kanade/tachiyomi/extension/all/mangadex/MdAtHomeReportInterceptor.kt
```diff
@@ -13,18 +13,6 @@ import okhttp3.RequestBody.Companion.toRequestBody
 import okhttp3.Response
 import uy.kohesive.injekt.injectLazy
 
-/**
- * Rate limit requests ignore covers though
```
vetleledaal, line 17: Am I reading this wrong, or does this change add rate limiting for covers? They were probably excluded for a reason

stevenyomi: Reasons for this:
- The cover URL pattern has already changed (so the code is not working as expected) for quite a while and nothing has been wrong with the extension.
- The app provided interceptor can handle canceled requests and won't let cover requests jam it when opening manga detail screen.

PR #13446, opened 2022-09-12
fix welovemanga

2Loong6:
Checklist:


multisrc/overrides/fmreader/rawlh/src/WeLoveManga.kt
```diff
@@ -30,4 +36,16 @@ class WeLoveManga : FMReader("WeLoveManga", "https://weloma.art", "ja") {
     }
     // Referer needs to be chapter URL
     override fun imageRequest(page: Page): Request = GET(page.imageUrl!!, headersBuilder().set("Referer", page.url).build())
+
+    override fun popularMangaFromElement(element: Element): SManga {
+        val id = element.select(popularMangaSelector()).select(".thumb-wrapper")
+        val document = client.newCall(GET("$popUrl?action=pop&id=${id.attr("data-id")}")).execute().asJsoup()
```
alessandrojean, line 42: You're doing an extra network call for each entry in the popular list. This is not a good approach. If the website doesn't provide all information needed in their current popular listing, it's better to just use some other listing instead of popular.

2Loong6: The website just provide information on mouse stay, i will take another way to resolve

PR #13552, opened 2022-09-22
Add new niceoppai extention

ponekub083:
Checklist:


src/th/niceoppai/build.gradle
```diff
@@ -0,0 +1,12 @@
+apply plugin: 'com.android.application'
+apply plugin: 'kotlin-android'
+
+ext {
+    extName = 'Niceoppai'
+    pkgNameSuffix = 'th.niceoppai'
+    extClass = '.Niceoppai'
+    extVersionCode = 1
```
vetleledaal, line 8:
```suggestion
    extVersionCode = 25
```
... to continue from the Madara version

vetleledaal (changes requested):
Since you're moving from Madara you should remove the other resources (`./multisrc/overrides/madara/niceoppai/`), and the line in `MadaraGenerator.kt`

ponekub083:
@alessandrojean please review code and build extention for release to main apps.

src/th/niceoppai/build.gradle
alessandrojean: It's the same package name and classname so it should appear as an update. Please apply the changes suggested by Vetle.

alessandrojean:
The code is fine, just need to apply the changes suggested by Vetle so the new extension appears as an update to existing users.

src/th/niceoppai/build.gradle
ponekub083: @alessandrojean @vetleledaal  thank you. it applied.

vetleledaal (approved):

PR #13562, opened 2022-09-24
BatoTo: Replace CryptoJS use with javax.crypto

pratyush3757:
Removed quickJS dependency which causes the extension to not run on Tachidesk-Server.
Tested viewing and downloads with https://bato.to/chapter/2057246 and various other chapters. 

Checklist:


stevenyomi:
Can this be applied to Comico and MangaPark too?

https://github.com/tachiyomiorg/tachiyomi-extensions/blob/4b956148e1da2b61205e3f205c84d2a23ab984b5/src/all/comico/src/eu/kanade/tachiyomi/extension/all/comico/Comico.kt#L169-L176

https://github.com/tachiyomiorg/tachiyomi-extensions/blob/e96fd22d307d5cc46cbf7e3ecbbbffb644f0fcda/src/all/mangapark/src/eu/kanade/tachiyomi/extension/all/mangapark/MangaPark.kt#L263-L266

pratyush3757:
> Can this be applied to Comico and MangaPark too?

I'm planning to apply this to mangapark after this in a seperate PR, but haven't checked Comico. 
> https://github.com/tachiyomiorg/tachiyomi-extensions/blob/4b956148e1da2b61205e3f205c84d2a23ab984b5/src/all/comico/src/eu/kanade/tachiyomi/extension/all/comico/Comico.kt#L169-L171

This says empty IVs aren't supported, I have to try something to bypass that. 
I think CryptoJS uses 0x00 bytes as IV under the hood, not an empty array. Becasue AES needs an IV to function.

pratyush3757:
Or should I include the changes for Mangapark in this PR?

stevenyomi:
Would be best if the crypto code can be extracted to a separate lib (like what I did for the code to decode packer-encoded snippets).

@arkon Can we just add these libs to `lib/` in this repo? I can move my Unpacker code to this repo too.

arkon:
> Can we just add these libs to lib/ in this repo? I can move my Unpacker code to this repo too.

We could, although it'd introduce some redundant module building per run. It'd probably be ideal to actually move them to another repo.

stevenyomi:
I've opened a PR to move the unpacker script. @pratyush3757 You can copy the dataimg lib's structure to make a new module. You also need to add it to `settings.gradle.kts`.

pratyush3757:
Uh, I've migrated it to a module but I'm not familiar enough with kotlin and gradle.
Although it's building and running correctly, can you review the code and build files for any problems?
I'm not sure if the build process has any redundancy.

stevenyomi:
Looks good. Just that the deobfuscator should only be in the lib if another source also uses it. If it's exclusive to Bato.to, you should move it inside the source.

pratyush3757:
The deobfuscator will be used by mangapark too.

stevenyomi:
Also I think it's okay to include other sources in this PR if the replacement is the only change.

pratyush3757:
> Also I think it's okay to include other sources in this PR if the replacement is the only change.

Yeah, I'm almost done changing mangapark. Will probably take some time on Comico.

PR #13600, opened 2022-09-27
MangaDex: use `matchEntire` to match UUID

stevenyomi:
An addition to #13597.


alessandrojean:
It isn't a major change that would necessarily require a version bump, so merging as it is.

Issue #13444, opened 2022-09-11
Doujindesu: Change website appearance


TheKingTermux:
### Source information and language

Doujindesu 1.2.2 (ID)

### Steps to reproduce

Add Doujin to library
Refresh all Comic in Tachi
Error `Attempt to invoke virtual method 'org.jsoup.select.Elements org.jsoup.nodes.Element.select(java.lang.String)' on a null object referen

### Expected behavior

Doujin must be refresh

### Actual behavior

Error `Attempt to invoke virtual method 'org.jsoup.select.Elements org.jsoup.nodes.Element.select(java.lang.String)' on a null object referen
![Screenshot_2022-09-12-04-48-52-61_40deb401b9ffe8e1df2f1cc5ba480b12](https://user-images.githubusercontent.com/50316075/189550443-5d84ac1b-12eb-449a-9025-7dbba68f712b.jpg)

### Latest Information 
![IMG_20220912_180111](https://user-images.githubusercontent.com/50316075/189638062-c226f817-29ab-4635-9c8b-3c3af6006dce.jpg)

The Doujindesu website is changes in appearance and add addition of new features

### Doujindesu domain
Still can be accessible with https://doujindesu.xxx (by using a VPN)
IP raw domain https://212.32.226.234 (By not using a VPN & absolutely can't be used in tachi extension due non-secure measure)

### Latest update
Main page is can be accessible with my local coding, _description (it appears now)_, _author (it appears now but still confusing)_, _stats (it appears now)_

![Screenshot_2022-09-26-18-46-25-24_afd66b7916d19a60815f56273408b73e.jpg](https://user-images.githubusercontent.com/50316075/192268519-04876a2b-a315-4ba7-8163-a910369ea77a.jpg)

![Screenshot_2022-09-26-18-46-27-99_afd66b7916d19a60815f56273408b73e.jpg](https://user-images.githubusercontent.com/50316075/192268537-e6ccc666-1609-41c6-a3ee-f275eee10265.jpg)

![Screenshot_2022-09-28-16-36-59-98_afd66b7916d19a60815f56273408b73e.jpg](https://user-images.githubusercontent.com/50316075/192753670-3c69e8db-bc5f-40a2-8bbf-24e285d57af0.jpg)

### Tachiyomi version

0.13.6

### Android version

Android 11

### Acknowledgements

- [X] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open or closed issue.
[...]

Issue #12637, opened 2022-07-19
Doujindesu: Missing genres


AETHER0009:
### Describe your suggested feature

I hope the developer can complement the genres in the doujindesu extension.In tachiyomi there are only genres from age regression to muscle. Whereas in the web version doujindesu there is more than that. Please help perhaps more genres can be added. Thanks

### Acknowledgements

- [X] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open issue.
- [X] I have written a short but informative title.
- [X] If this is an issue with the app itself, I should be opening an issue in the [app repository](https://github.com/tachiyomiorg/tachiyomi/issues/new/choose).
- [X] I have updated the app to version **[0.13.5](https://github.com/tachiyomiorg/tachiyomi/releases/latest)**.
- [X] I will fill out all of the requested information in this form.

AETHER0009:
Oke gan makasih

TheKingTermux:
![IMG_20220802_165921.jpg](https://user-images.githubusercontent.com/50316075/182347886-f6f84357-3a8e-45f3-b0d9-7cb90d25d554.jpg)
Sudah dicoba gan
Mungkin ntar up klo Doujindesu dah bikin domain baru.
Untuk thumbnail mungkin masih belum bisa muncul

PR #13605, opened 2022-09-27
Doujindesu: Change website appearance 

TheKingTermux:
### Status
- Search `Tested`
- Multi Genre `Actually the website only takes comics with the chosen genre, but not all of them, only comics that contain the selected genres` [Read This for Further Information](https://github.com/tachiyomiorg/tachiyomi-extensions/pull/13605#issuecomment-1267891653)
- Full Genre `Yes`
- Better Filter `Tested`

### Manga Details
- Thumbnail `Didn't appear`
- Description `It appear now`
- Genres `It appear now`
- Author `It appear now`
- Artist / Character `It appear now`
- Status `It appear now`

### Referenced Issues
Closes [#13444 ](https://github.com/tachiyomiorg/tachiyomi-extensions/issues/13444) `Doujindesu: Change website appearance`
Closes [#12637](https://github.com/tachiyomiorg/tachiyomi-extensions/issues/12637) `Doujindesu: Missing genres`

### Checklist:


src/id/doujindesu/src/eu/kanade/tachiyomi/extension/id/doujindesu/DoujinDesu.kt
```diff
@@ -271,23 +323,29 @@ class DoujinDesu : ParsedHttpSource() {
...
+            val genre = element.text()
+            genres.add(genre)
         }
-        manga.author = document.select("div.infox > div.spe > span:nth-child(5)").text()
-        manga.genre = document.select("div.genre-info > a[itemprop=genre]").joinToString { it.text() }
-        manga.status = parseStatus(document.select("div.infox > div.spe > span:nth-child(1)").text())
-        manga.thumbnail_url = document.select("div.thumb > img").attr("src")
-        manga.artist = document.select("div.infox > div.spe > span:nth-child(6)").text()
+        manga.author = document.select("section.metadata > table > tbody > tr.pages:nth-child(4) > td:nth-child(2) > a").first().text()
+        manga.genre = genres.joinToString(", ")
```
alessandrojean, line 345: You can do this in a better way:

```kotlin
manga.genre = infoElement.select("div.tags > a[itemprop=tags]").joinToString { it.text() }
```

alessandrojean (changes requested):

src/id/doujindesu/src/eu/kanade/tachiyomi/extension/id/doujindesu/DoujinDesu.kt
TheKingTermux: Ok

TheKingTermux:
IMG_20220928_162745
IMG_20220928_162944
IMG_20220928_163902
IMG_20220928_164255

That is very inversely proportional to Manhwa. At Manhwa I got an author in Child (4) and an artist in Child (5), while in Manga I got an author in Child (5) and an artist in Child (4). After trying debug I was a little confused because what appeared on the author was the artist, while the artist's part was the author. Can this be tricked using certain codes? Or just let it go?

TheKingTermux:
> `td:contains(Status) + td`.
Then like this ?
manga.author = document.select("section.metadata > table:nth-child(2) > tbody > tr.pages:nth-child(4) > td:contains(Artist) + td:nth-child(3) > td:nth-child(2) > a").first().text()

td:nth-child(2) : Author
td:nth-child(3) : Artist

alessandrojean:
No, the selector I did as an example is all you probably need to select the author, no need for a complex an gigantic selector like the one you provided. You may want to do some tests on the online tool provided by Jsoup: https://try.jsoup.org

TheKingTermux:
> 

Hhmmm..... 🤔
Then like this ?

td:contains(Author) + td > a
td:contains(Artist) + td > a

TheKingTermux:
The genre in extension and in the web is entering the url ?

Like in the extension is
Genre("Mind Control", "mind-control")
With "genre[]"

In the website is
Mind Control 
But in url like this 
https://doujindesu.xxx/manga/?title=&author=&character=&statusx=&typex=&order=title&genre%5B%5D=Mind+Control

("mind-control") and =Mind+Control must same ?

And for multi genre search like this
Genre("Mind Control", "mind-control")
Genre("Mind Break", "mind-break")
With "genre[]"

But in the website is
https://212.32.226.234/manga/?title=&author=&character=&statusx=&typex=&order=title&genre%5B%5D=Mind+Break&genre%5B%5D=Mind+Control

I can't access it
I mean I can't access the genres with 2 words or more
What should I do?

vetleledaal:
You can modify the code like so:
```kotlin
Genre("Mind Control")
Genre("Mind Break")
// ...
```
OkHttp will properly\* escape it for you in the search query.


\*There might be a minor difference where OkHttp encodes spaces as `%20` (per RFC 3986) and not `+` (because OkHttp doesn't know that it's a `application/x-www-form-urlencoded` form). Ultimately this doesn't matter as the server interprets the request the same.

TheKingTermux:
> You can modify the code like so:
> ```kotlin
> Genre("Mind Control")
> Genre("Mind Break")
> // ...
> ```

This is applied for all genres or just genres with 2 words or more?

Like
```kotlin
Genre("Ahegao")
Genre("Amprit")
Genre("Mind Control")
Genre("Mind Break")
// ...
```

Or like this
```kotlin
Genre("Ahegao", "ahegao")
Genre("Amprit", "amprit")
Genre("Mind Control")
Genre("Mind Break")
// ...
```

vetleledaal:
All genres, so your first example would be correct 
```kotlin
Genre("Ahegao")
Genre("Amprit")
Genre("Mind Control")
Genre("Mind Break")
// ...
```

TheKingTermux:
> All genres, so your first example would be correct 

Alright thanks 👍🏻

TheKingTermux:
> ```kotlin
> Genre("Ahegao")
> Genre("Amprit")
> Genre("Mind Control")
> Genre("Mind Break")
> // ...
> ```

And now how to use this code to search multiple genres at once ?

I already check it with the

> Genre("Mind Control")
> Genre("Mind Break")

And that only brings up comics with one of those genres, not both

TheKingTermux:
I have error because the manga is not contains Author and Character, the impact is the genres and the stats can't be loaded





```kotlin
tr.pages > td:contains(Author) + td:nth-child(2) > a
tr.pages > td:contains(Character) + td:nth-child(2) > a
```

vetleledaal:
since `document.select("...").first()` can return null, and it's Java code (read: no type hints) you need to tell Kotlin that it can be null.
so something like this is fine:
```kotlin
manga.author = document.select("...").first()?.text()
manga.artist = document.select("...").first()?.text()
// or ...
manga.author = document.selectFirst("...")?.text()
manga.artist = document.selectFirst("...")?.text()
// or for multiple matches ...
manga.author = document.select("...")?.joinToString { it.text() }
manga.artist = document.select("...")?.joinToString { it.text() }
```

For the genre filter I don't think you can use a TriState, as the site doesn't support excluding genres from search. To only send the selected genres you can do something like in Madara:
https://github.com/tachiyomiorg/tachiyomi-extensions/blob/d0793a1741a80b22537c77327810d4fc2185baa1/multisrc/src/main/java/eu/kanade/tachiyomi/multisrc/madara/Madara.kt#L211-L217

TheKingTermux:
>```kotlin
> // or for multiple matches ...
> manga.author = document.select("...").joinToString { it.text() }
> manga.artist = document.select("...").joinToString { it.text() }
> ```

How to use with this configuration?
If it still fail I will change to this one

> ```kotlin
> manga.author = document.select("...").first()?.text()
> manga.artist = document.select("...").first()?.text()
> // or ...
> manga.author = document.selectFirst("...")?.text()
> manga.artist = document.selectFirst("...")?.text()
> ```

> For the genre filter I don't think you can use a TriState, as the site doesn't support excluding genres from search. To only send the selected genres you can do something like in Madara:
> https://github.com/tachiyomiorg/tachiyomi-extensions/blob/d0793a1741a80b22537c77327810d4fc2185baa1/multisrc/src/main/java/eu/kanade/tachiyomi/multisrc/madara/Madara.kt#L211-L217

Ok then I just copy and paste like Madara in line 211 - 217 to my configuration

vetleledaal:
> > ```kotlin
> > // or for multiple matches ...
> > manga.author = document.select("...").joinToString { it.text() }
> > manga.artist = document.select("...").joinToString { it.text() }
> > ```
> 
> How to use with this configuration? If it still fail I will change to this one

This is only if the site supports multiple artists/authors, the selector needs to pick up multiple elements and they are concatenated with a ", " so the final string would look like "artist1, artist2, artist3" and so on

TheKingTermux:
> This is only if the site supports multiple artists/authors,  a ", "  "artist1, artist2, artist3" 

Actually it have a manga have 2-3 author
But I will use with this instead

> ```kotlin
> manga.author = document.select("...").first()?.text()
> manga.artist = document.select("...").first()?.text()
> // or ...
> manga.author = document.selectFirst("...")?.text()
> manga.artist = document.selectFirst("...")?.text()
> ```

TheKingTermux:
How to display the thumbnail?

The thumbnail source is in

```kotlin
#archives > div.entries > article:nth-child(1) > a > figure.thumbnail > img
```

Then in my configuration like this

```kotlin
private fun basicInformationFromElement(element: Element): SManga {
        val manga = SManga.create()
        element.select("a").let {
            manga.title = it.attr("title")     manga.setUrlWithoutDomain(it.attr("href"))
            manga.thumbnail_url = it.select("figure.thumbnail > img").attr("abs:src")
        }
```

And in the mangaDetailParse

```kotlin
#archive > div > aside > figure.thumbnail > a > img
```

```kotlin
manga.thumbnail_url = document.select("figure.thumbnail > img").attr("src")
```

But the thumbnail is didn't appear

Is it because the BaseUrl from the domain becomes the IP Address or something else?
Can it be fixed with code?

vetleledaal:
Nothing wrong with the code, the site is just broken. See [example (NSFW)](https://doujindesu.xxx/manga/?title=&author=&character=&statusx=&typex=&order=title&genre%5B%5D=Mind+Break&genre%5B%5D=Mind+Control), the site only looks at the first genre

TheKingTermux:
> Nothing wrong with the code, the site is just ~~broken~~ looking for any of the genres. See [example (NSFW)](https://doujindesu.xxx/manga/?title=&author=&character=&statusx=&typex=&order=title&genre%5B%5D=Mind+Break&genre%5B%5D=Mind+Control)
> 
> "Academi Girls" in the results use both genres. I don't think you can change the behavior so it only includes results where all the genres match

Basically the web is the main problem at it 
Now I got it .... (⁠─⁠.⁠─⁠|⁠|⁠）

I already use the Madara GenreList
Did this change to ?

https://github.com/tachiyomiorg/tachiyomi-extensions/blob/4d01b25466fa190f21e8f0919c13d96def3cfc14/src/id/doujindesu/src/eu/kanade/tachiyomi/extension/id/doujindesu/DoujinDesu.kt#L44-L47

vetleledaal:
You are returning `name`, which is not defined, should be either `title` or `id`. I think Madara uses `id`, but not sure

TheKingTermux:
> You are returning `name`, which is not defined, should be either `title` or `id`. I think Madara uses `id`, but not sure

Okay
I will use `id` instead and checking it later

TheKingTermux:
> You are returning `name`, which is not defined, should be either `title` or `id`. I think Madara uses `id`, but not sure

https://github.com/tachiyomiorg/tachiyomi-extensions/blob/d0793a1741a80b22537c77327810d4fc2185baa1/multisrc/src/main/java/eu/kanade/tachiyomi/multisrc/madara/Madara.kt#L208-L209

https://github.com/tachiyomiorg/tachiyomi-extensions/blob/d0793a1741a80b22537c77327810d4fc2185baa1/multisrc/src/main/java/eu/kanade/tachiyomi/multisrc/madara/Madara.kt#L360-L365

https://github.com/tachiyomiorg/tachiyomi-extensions/blob/d0793a1741a80b22537c77327810d4fc2185baa1/multisrc/src/main/java/eu/kanade/tachiyomi/multisrc/madara/Madara.kt#L367-L372

This used for ?

vetleledaal:
Madara fetches the genre list automatically. Not really a concern here since it's just for a single site, unless they updte genres often.

TheKingTermux:
> Madara fetches the genre list automatically. Not really a concern here since it's just for a single site, unless they updte genres often.

I see
Thanks 👍🏻

src/id/doujindesu/src/eu/kanade/tachiyomi/extension/id/doujindesu/DoujinDesu.kt
```diff
@@ -30,8 +30,8 @@ class DoujinDesu : ParsedHttpSource() {
     }
 
     private fun parseStatus(status: String) = when {
-        status.lowercase(Locale.US).contains("finished") -> SManga.ONGOING
-        status.lowercase(Locale.US).contains("publishing") -> SManga.COMPLETED
```
MoonLGH, line 34: wait what, I'm so sorry i wrote this wron

TheKingTermux: Yeah no problem
I've fixed it

MoonLGH:
good, after all the above fixed its an LGTM approved

TheKingTermux:
> good, after all the above fixed its an LGTM approved

I'm still testing how to show the thumbnail, and for the code I still haven't copied and pasted it from my Android Studio to Github so the code here isn't up to date, maybe I'll try again later and updating the Github, because I'm a little dizzy today

TheKingTermux:
Well I still confusing about the thumbnail
I thought I had written the code correctly, but there was no progress for it

Or maybe the thumbnail can't be taken because the `BaseUrl` is different? because here I have to use the IP Address to access it

MoonLGH:
fair enough, LGTM but sadly i cant approved a PR

TheKingTermux:
> fair enough, LGTM but sadly i cant approved a PR

Sorry wrong press

alessandrojean (approved):
Some selectors probably can be simplified, but if it works for now it's fine.

PR #13627, opened 2022-09-29
MyHentaiGallery: Fix broken thumbnail_url in mangaDetailsParse

SuperCoolGuy855:
Checklist:


src/en/myhentaigallery/src/eu/kanade/tachiyomi/extension/en/myhentaigallery/MyHentaiGallery.kt
```diff
@@ -111,7 +111,6 @@ class MyHentaiGallery : ParsedHttpSource() {
                 title = info.select("h1").text()
                 genre = info.select("div:containsOwn(categories) a").joinToString { it.text() }
                 artist = info.select("div:containsOwn(artists) a").text()
-                thumbnail_url = info.select("div.comic-cover img").attr("abs:src")
```
alessandrojean, line 114: Why not just get the first image from the album instead, then?

```kotlin
thumbnail_url = document.selectFirst(".comic-listing .comic-inner img")?.attr("src")
```

It's not good to not set the `thumbnail_url` at the `mangaDetailsParse` as if they ever change the thumbnail, it will never get updated in the app.

SuperCoolGuy855: I didn't think of that.

alessandrojean (approved):

PR #13668, opened 2022-10-01
9Hentai: Fix download page number broken

AntsyLich:
Checklist:


CarlosEsco:
Maybe @arkon disagrees but I don't think we should hiding issues with extensions that are broken and need fixed.

PR #13686, opened 2022-10-02
ComickFun: Fixed WebView not working correctly

### LINKED ISSUE #13676: Comick.fun webview lead to API page instead site.

Comick.fun 1.3.14

### Steps to reproduce

1. Go to manga details
2. Click webview 

### Expected behavior

Go to manga page on site

### Actual behavior

JSON from API shows.

### Tachiyomi version

0.13.6

### Android version

A 12.1

### Other details

Ok, so mangadex also use API, but when we click webview it's lead us to manga page, maybe I'm wrong, but it's should be the same here

### Acknowledgements

- [X] I have updated the app to version **[0.13.6](https://github.com/tachiyomiorg/tachiyomi/releases/latest)**.

sergiomalagon:
Closes #13676

WebView now shows the manga page instead of JSON

Checklist:


vetleledaal:
You should override `fetchMangaDetails()` so you don't have to perform extra requests. Like in MangaDex:

https://github.com/tachiyomiorg/tachiyomi-extensions/blob/289f5db72362b74dda7aa3aec58cf1739c6c3b51/src/all/mangadex/src/eu/kanade/tachiyomi/extension/all/mangadex/MangaDex.kt#L437-L457

sergiomalagon:
> You should override `fetchMangaDetails()` so you don't have to perform extra requests. Like in MangaDex:
> 
> https://github.com/tachiyomiorg/tachiyomi-extensions/blob/289f5db72362b74dda7aa3aec58cf1739c6c3b51/src/all/mangadex/src/eu/kanade/tachiyomi/extension/all/mangadex/MangaDex.kt#L437-L457

But if I use fetchMangaDetails() is still 2 request...

vetleledaal:
The WebView button gets the URL directly from `mangaDetailsRequest()`
By default `fetchMangaDetails()` makes a call to `mangaDetailsRequest()`
In MangaDex `fetchMangaDetails()` was modified to make a call to `apiMangaDetailsRequest()`, which is 1 request in total

In this PR you first make a call to the manga URL, then the API

Consider:
```kotlin
/** Manga Details **/
// Shenanigans to allow "open in webview" to show a webpage instead of JSON
override fun fetchMangaDetails(manga: SManga): Observable<SManga> {
    return client.newCall(apiMangaDetailsRequest(manga))
        .asObservableSuccess()
        .map { response ->
            mangaDetailsParse(response).apply { initialized = true }
        }
}

override fun mangaDetailsRequest(manga: SManga): Request {
    return GET("$baseUrl${manga.url}".toHttpUrl().toString())
}

private fun apiMangaDetailsRequest(manga: SManga): Request {
    return GET(
        "$API_BASE${manga.url}".toHttpUrl().newBuilder().apply {
            addQueryParameter("tachiyomi", "true")
        }.toString(),
        headers
    )
}

override fun mangaDetailsParse(response: Response): SManga {
    val mangaData = json.decodeFromString<MangaDetails>(response.body!!.string())
    return SManga.create().apply {
        url = "$baseUrl/comic/${mangaData.comic.slug}"
        // ...
    }
    // ...
}
```

src/all/comickfun/src/eu/kanade/tachiyomi/extension/all/comickfun/ComickFun.kt
```diff
@@ -193,17 +193,24 @@ abstract class ComickFun(override val lang: String, private val comickFunLang: S
...
-                addQueryParameter("tachiyomi", "true")
-            }.toString(),
-            headers
-        )
+        return GET("$baseUrl${manga.url}".toHttpUrl().toString())
     }
 
     override fun mangaDetailsParse(response: Response): SManga {
-        val mangaData = json.decodeFromString<MangaDetails>(response.body!!.string())
+        val mangaSlug = response.request.url.toString().substringAfterLast("/")
```
alessandrojean, line 200: In that way you're doing one additional call that could be avoided. Take a look on how extensions like Mangadex or MangaPlus which also uses an API does and you will find a better way of fixing that.

https://github.com/tachiyomiorg/tachiyomi-extensions/blob/master/src/all/mangadex/src/eu/kanade/tachiyomi/extension/all/mangadex/MangaDex.kt#L441-L473

Also, no need to define the mangaDetailsRequest as it is right now as it's exactly like the super method, so it can be removed.

alessandrojean (changes requested):

src/all/comickfun/src/eu/kanade/tachiyomi/extension/all/comickfun/ComickFun.kt
```diff
@@ -192,23 +194,24 @@ abstract class ComickFun(override val lang: String, private val comickFunLang: S
...
                 }.toString(),
                 headers
             )
-        ).execute()
-        val mangaData = json.decodeFromString<MangaDetails>(mangaRequest.body!!.string())
+        ).asObservableSuccess()
+            .map { response -> mangaDetailsParse(response).apply { initialized = true } }
+    }
+
+    override fun mangaDetailsRequest(manga: SManga): Request {
```
alessandrojean, line 209: You can also remove this one now as it's exactly like the superclass, so no override is needed.

It's just a suggestion though, not really mandatory.

alessandrojean (approved):

src/all/comickfun/src/eu/kanade/tachiyomi/extension/all/comickfun/ComickFun.kt
sergiomalagon: It's not the same, `fetchMangaDetails()` use API_BASE and `mangaDetailsRequest()` use baseUrl

AntsyLich: Your implementation of `mangaDetailsRequest` is same as that in `HttpSource` so it's pretty much redundant.

sergiomalagon: I prefer to leave it there, performance/request wise it's the same, but it makes it easy to understand where the WebView url it's coming from. At least that's how I see it.

AntsyLich: ...it does not lol. You gotta see app code to understand it.

sergiomalagon: It's possible, please tell me then with it does matter

alessandrojean: You can leave like it's right now, it was just a suggestion, not really mandatory.

sergiomalagon: I know I know, I wasn't being ironic, I just want to know why it matters

PR #13304, opened 2022-09-01
Fix Webtoons & Tapas Author's Notes

altaccosc:
The [`placehold.jp`](https://placehold.jp/en.html) API I was using before went under, so now I'm using a different one: [`fakeimg.pl`](https://fakeimg.pl).
I don't like this one as much, but it'll have to do. I tested soo many other APIs like this and this one was the best one out of all of those.

Checklist:

- [x] Updated `extVersionCode` value in `build.gradle` for individual extensions
- [x] Updated `overrideVersionCode` or `baseVersionCode` as needed for all multisrc extensions
- [x] Referenced all related issues in the PR body (e.g. "Closes #xyz")
- [x] Added the `isNsfw = true` flag in `build.gradle` when appropriate
- [x] Have not changed source names
- [x] Have explicitly kept the `id` if a source's name or language were changed
- [x] Have tested the modifications by compiling and running the extension through Android Studio

Closes #13279

PR #13859, opened 2022-10-14
New Lib: Text Interceptor

curche:
(Please read the individual commit messages)

for now, moved the current two usages of TextInterceptor to new lib: Tapastic, and Webtoons

Webtoons one was bit more tricky and if needed, I can make it a separate PR so that its bit more easier to review

Also planning to use the interceptor for xkcd alt text and patch friday alt text but those PR won't be anytime soon. gotta learn the interceptor a bit more lul

Checklist:


curche:
cc previous contributors @altaccosc and @stevenyomi from #13304. Just wanted to confirm one thing: Is the author notes setting meant for just Webtoons extension or all sources in the Webtoon multisrc family (webtoons, webtoonstranslate, and dongmanmanhua)

lib/textinterceptor/src/main/java/eu/kanade/tachiyomi/lib/textinterceptor/TextInterceptor.kt
```diff
@@ -0,0 +1,111 @@
...
+        return Response.Builder()
+            .request(request)
+            .protocol(Protocol.HTTP_1_1)
+            .code(200)
+            .message("OK")
+            .body(responseBody)
+            .build()
+    }
+
+    private fun textFixer(t: String): String {
```
vetleledaal, line 95: Why not use [`Html.fromHtml()`](https://stackoverflow.com/a/40871051)? or preferably handle this in the individual extensions.

lib/textinterceptor/src/main/java/eu/kanade/tachiyomi/lib/textinterceptor/TextInterceptor.kt
```diff
@@ -0,0 +1,111 @@
...
+        private const val WIDTH: Int = 1000
+        private const val X_PADDING: Float = 50f
+        private const val Y_PADDING: Float = 25f
+        private const val HEADING_FONT_SIZE: Float = 36f
+        private const val BODY_FONT_SIZE: Float = 30f
+        private const val SPACING_MULT: Float = 1.1f
+        private const val SPACING_ADD: Float = 2f
+
+        // No need to touch this one:
+        private const val HOST = "note"
```
vetleledaal, line 33: this is technically a valid domain name. maybe use `note.invalid`, as the TLD is one of the suggested ones by [RFC 2606](https://www.rfc-editor.org/rfc/rfc2606)

altaccosc:
> cc previous contributors @altaccosc and @stevenyomi from #13304. Just wanted to confirm one thing: Is the author notes setting meant for just Webtoons extension or all sources in the Webtoon multisrc family (webtoons, webtoonstranslate, and dongmanmanhua)

I only use the normal WebToons, so I added it to that one not bothering about the others.
If you want to and they could use it, by all means, do add it to those too.

lib/textinterceptor/src/main/java/eu/kanade/tachiyomi/lib/textinterceptor/TextInterceptor.kt
curche: I just copied over what the previous implementation did as a start. I'll have a look

lib/textinterceptor/src/main/java/eu/kanade/tachiyomi/lib/textinterceptor/TextInterceptor.kt
curche: > valid domain name

I believe you meant *invalid* 🤔 

Out of the options .test, .example, .invalid, .localhost, and their defined use cases, localhost werks ig

> custom scheme

what could we have? like a uri? tachiyomi::/lib-text something?

curche:
> > cc previous contributors @altaccosc and @stevenyomi from #13304. Just wanted to confirm one thing: Is the author notes setting meant for just Webtoons extension or all sources in the Webtoon multisrc family (webtoons, webtoonstranslate, and dongmanmanhua)
> 
> I only use the normal WebToons, so I added it to that one, while not bothering about the others. If you want to and they could use it, by all means, do add it to those too.

I prefer keeping them in normal webtoons only and that's a lot easier to do (c.f. #13860) so I'll stick with that. thanks!

lib/textinterceptor/src/main/java/eu/kanade/tachiyomi/lib/textinterceptor/TextInterceptor.kt
vetleledaal: > > valid domain name
> 
> I believe you meant _invalid_ 🤔
> 
Domains without TLDs are valid.

> 
> > custom scheme
> 
> what could we have? like a uri? tachiyomi::/lib-text something?
>
This could work, although an URI only requres one colon, like `tachiyomi:lib-text`. Additional data could be passed in any manner, but sticking to norm here would be best `tachiyomi:lib-text/data1/data2`

lib/textinterceptor/src/main/java/eu/kanade/tachiyomi/lib/textinterceptor/TextInterceptor.kt
vetleledaal: Fair, but once it becomes a library you'd expect less dependency on assumptions made for the first site. If you wanted to display `&amp;` you would have to send `&amp;amp;`. Contrived example, I know, but double decoding is rarely a desirable feature.

altaccosc: My implementation could indeed be considered "hacky", so if there are improvements for my code, please do out those in :D

altaccosc:
> > > cc previous contributors @altaccosc and @stevenyomi from #13304. Just wanted to confirm one thing: Is the author notes setting meant for just Webtoons extension or all sources in the Webtoon multisrc family (webtoons, webtoonstranslate, and dongmanmanhua)
> > 
> > 
> > I only use the normal WebToons, so I added it to that one, while not bothering about the others. If you want to and they could use it, by all means, do add it to those too.
> 
> I prefer keeping them in normal webtoons only and that's a lot easier to do (c.f. #13860) so I'll stick with that. thanks!

Fine by me! :smile::+1:

lib/textinterceptor/src/main/java/eu/kanade/tachiyomi/lib/textinterceptor/TextInterceptor.kt
ArjixWasTaken: @curche


> > valid domain name
> 
> I believe you meant *invalid* 🤔 

Nah, https://pn/ is a valid site.

curche:
Long time no see! >.>
Rebased the branch but I've made note of the previous suggestions
 - use in-built Html.fromHtml() to unescape html instead of current implementation
 - collect the HOST name into the lib (kind of?) and redefine it as `tachiyomi-lib-interceptor`

tested on the following:
 - webtooons - check the Author Note chapter on title "GOTCHA"
 - tapas - check the first chapter on title "The Art in Heartbeat", an author's first title and its first chapter

vetleledaal (approved):

src/en/tapastic/src/eu/kanade/tachiyomi/extension/en/tapastic/Tapastic.kt
```diff
@@ -360,7 +361,7 @@ class Tapastic : ConfigurableSource, ParsedHttpSource() {
 
                 pages = pages + Page(
                     pages.size, "",
-                    "http://note/" + Uri.encode(creator) + "/" + Uri.encode(episodeStory)
```
altaccosc, line 375: Why not just make a new (static) function in the TextInterceptor that returns a new Page from the two input strings: `author` and `note`?

lib/textinterceptor/src/main/java/eu/kanade/tachiyomi/lib/textinterceptor/TextInterceptor.kt
```diff
@@ -0,0 +1,115 @@
...
+import okhttp3.ResponseBody.Companion.toResponseBody
+import java.io.ByteArrayOutputStream
+
+class TextInterceptor : Interceptor {
+    // With help from:
+    // https://github.com/tachiyomiorg/tachiyomi-extensions/pull/13304#issuecomment-1234532897
+    // https://medium.com/over-engineering/drawing-multiline-text-to-canvas-on-android-9b98f0bfa16a
+
+    companion object {
+        // Designer values:
```
altaccosc, line 25: Maybe another fun thing would be to allow these to be overridden by giving them into the constructor.
This would allow multiple apps to implement their own look and style, while still depending on this same library code.

lib/textinterceptor/src/main/java/eu/kanade/tachiyomi/lib/textinterceptor/TextInterceptor.kt
```diff
@@ -0,0 +1,115 @@
...
+        private const val WIDTH: Int = 1000
+        private const val X_PADDING: Float = 50f
+        private const val Y_PADDING: Float = 25f
+        private const val HEADING_FONT_SIZE: Float = 36f
+        private const val BODY_FONT_SIZE: Float = 30f
+        private const val SPACING_MULT: Float = 1.1f
+        private const val SPACING_ADD: Float = 2f
+
```
altaccosc, line 33: Ooh maybe add font customisation to it too

multisrc/src/main/java/eu/kanade/tachiyomi/multisrc/webtoons/Webtoons.kt
```diff
@@ -1,19 +1,6 @@
 package eu.kanade.tachiyomi.multisrc.webtoons
```
altaccosc, line 1: Why does this file still exist now that you seem to have replaced it with `WebtoonSrc`?
I don't think this file is used anywhere anymore?

ArjixWasTaken:
Am I the only one who got this in their notifications for no reason?

TechnicJelle:
> Am I the only one who got this in their notifications for no reason?

I got it, but there's a reason for me

You can unsubscribe if you don't want that, btw

lib/textinterceptor/src/main/java/eu/kanade/tachiyomi/lib/textinterceptor/TextInterceptor.kt
vetleledaal: Steps into feature creep area, don't think this is a good fit. Would in any case be limited to pre-installed fonts

altaccosc: I'll admit I do love me some great many features.
I believe the XKCD source also has text to display, but it uses a different font than Roboto, which is the default Android font.
So if we want to replace that too with this new TextInterceptor, the font would change dramatically, which imo doesn't necessarily have to happen if we were to add different fonts to this.

altaccosc:
> Am I the only one who got this in their notifications for no reason?

You [replied here once](https://github.com/tachiyomiorg/tachiyomi-extensions/pull/13859#discussion_r999667122), so you're subscribed to this PR (which you can unsub from though)

ArjixWasTaken:
> > Am I the only one who got this in their notifications for no reason?
> 
> You [replied here once](https://github.com/tachiyomiorg/tachiyomi-extensions/pull/13859#discussion_r999667122), so you're subscribed to this PR (which you can unsub from though)

Wow, my memory sucks

lib/textinterceptor/src/main/java/eu/kanade/tachiyomi/lib/textinterceptor/TextInterceptor.kt
curche: Yea I'll agree with vetle here and not add this

lib/textinterceptor/src/main/java/eu/kanade/tachiyomi/lib/textinterceptor/TextInterceptor.kt
curche: Maybe some time later, I feel. Currently, let's stick with this

lib/textinterceptor/src/main/java/eu/kanade/tachiyomi/lib/textinterceptor/TextInterceptor.kt
curche: Yea I'll agree with vetle here and not add this

multisrc/src/main/java/eu/kanade/tachiyomi/multisrc/webtoons/Webtoons.kt
curche: this is still inherited by WebtoonsTranslate

src/en/tapastic/src/eu/kanade/tachiyomi/extension/en/tapastic/Tapastic.kt
curche: done

Issue #13430, opened 2022-09-11
request for Aresmanga's source removal from tachiyomi app


Volver555:
### Source link

https://aresmanga.com/

### Other details

this an ownership proofk if it's enough: 
https://aresmanga.com/owner.html

it has my discord account if there is anything else required.
please contact me there, with my sincere thank!

### Requirements

- [X] Proof of ownership/intent to remove sent to a Tachiyomi Discord server mod via DM
- [X] Site only hosts content scanlated by the group and not stolen from other scanlators or official releases (i.e., not an aggregator site)
- [X] Site is not infested with user-hostile features (e.g., invasive or malicious ads)

loocool2:
Proof of ownership has been DMed to me

PR #13888, opened 2022-10-17
[skip ci] Move missing sources list to repo file

stevenyomi:
This moves the "Missing sources" list in #3475 to a separate markdown file in the repo, so it can be easily updated when extensions are removed.


mobi2002:
Should add #13430 to the list as well.

PR #13926, opened 2022-10-19
[RU]NudeMoon some fixes

abubaca4:
Load cover instead of first page
Sort chapters by name(server return sorted by date)
Fix some warnings
Fix date parse in rare situations
replace client.newCall to chapterListParse

Checklist:

I also planed to write cookie nm_mobile 1 to webview storage(to open mobile site instead of PC version) but I think in can take more time(find one example [here](https://github.com/tachiyomiorg/tachiyomi-extensions/blob/master/src/ru/comx/src/eu/kanade/tachiyomi/extension/ru/comx/ComX.kt#L52))


g4merka:
THANK YOU VERY MUCH!!!

PR #14031, opened 2022-10-29
Add option to attempt to use first volume cover on MangaDex

alessandrojean:
Changelog:

- Add option to attempt to use first volume covers (from the entry original language) when available;
- Mark one-shots with one volume and one chapter that have the "Oneshot" tag (and no "Anthology" tag) as complete;
- Query text on search is now trimmed.

Checklist:


src/all/mangadex/src/eu/kanade/tachiyomi/extension/all/mangadex/MangaDex.kt
```diff
@@ -365,33 +370,74 @@ abstract class MangaDex(final override val lang: String, private val dexLang: St
...
+    /**
+     * Attempt to get the first volume cover if the setting is enabled.
+     * Uses the 'covers' endpoint.
+     *
+     * @see CoverListDto
+     */
+    private fun fetchFirstVolumeCovers(mangaList: List<MangaDataDto>): Map<String, String>? {
+        if (!preferences.tryUsingFirstVolumeCover) {
+            return null
+        }
```
curche, line 413: instead of having this inside the function
wouldn't it be better to check the preference before the function is called

eg: https://github.com/tachiyomiorg/tachiyomi-extensions/blob/3dc9db2cc6f3b39b4be82dcd69f3c3f18fa23619/src/all/mangadex/src/eu/kanade/tachiyomi/extension/all/mangadex/MangaDex.kt#L489

src/all/mangadex/src/eu/kanade/tachiyomi/extension/all/mangadex/MangaDexIntl.kt
```diff
@@ -265,6 +265,20 @@ class MangaDexIntl(lang: String) {
                 "Enter as a Comma-separated list of uploader UUIDs"
     }
 
+    val tryUsingFirstVolumeCover: String = when (availableLang) {
+        BRAZILIAN_PORTUGUESE, PORTUGUESE -> "Tentar usar a capa do primeiro volume"
+        else -> "Attempt to use the first volume cover"
```
curche, line 270: Sentence seems incomplete
> Attempt to use First Volume Cover "as Cover"?

src/all/mangadex/src/eu/kanade/tachiyomi/extension/all/mangadex/MangaDexIntl.kt
```diff
@@ -265,6 +265,20 @@ class MangaDexIntl(lang: String) {
...
+    }
+
+    val tryUsingFirstVolumeCoverSummary: String = when (availableLang) {
+        BRAZILIAN_PORTUGUESE, PORTUGUESE ->
+            "Pode ser necessário atualizar manualmente os itens na biblioteca e limpar o " +
+                "banco de dados para as novas capas serem mostradas nas listas."
+        else ->
+            "You may need to manually refresh your library entries to take effect and " +
+                "clear the database items to new covers to be shown on listings."
```
curche, line 279: May need to manually refresh entries already in Library. Otherwise, Clear Database to have new covers show up

curche (approved):
Tested on local. 
Cache takes a while to renew
Might take a while to load Browse because it tries to find covers ig but its only slight in the first time. Otherwise, haven't found any incorrect covers in Latest/Browse

curche:
Er Might need to handle the Status of Anthology made up of Oneshot different, maybe?

src/all/mangadex/src/eu/kanade/tachiyomi/extension/all/mangadex/MangaDex.kt
curche: ye that was my dilemma as well
its fine then

PR #14038, opened 2022-10-30
Add extension for Randowiz

ericlp:
Created my first ever extension 🎉, which is for [Randowiz](https://randowis.com/). Has 3 mangas, his webcomic, short comics and illustrations.

Checklist:


src/en/randowiz/src/eu/kanade/tachiyomi/extension/en/randowiz/Randowiz.kt
```diff
@@ -0,0 +1,154 @@
...
+        if (next.isNotEmpty()) client.newCall(GET(next))
+            .asObservableSuccess()
+            .map { newWebpage ->
+                chapterListParse(newWebpage)
+            }.forEach { a -> chapters.addAll(a) }
+        return chapters.mapIndexed {
+            i, ch ->
+            ch.apply { chapter_number = chapters.size.toFloat() - i }
+        }
+    }
```
alessandrojean, line 96: You don't need to deal with observables and neither recursivity if you do like this:

```suggestion
    override fun chapterListParse(response: Response): List<SChapter> {
        val document = response.asJsoup()
        val chapters = document.select(chapterListSelector())
            .map { chapterFromElement(it) }
            .toMutableList()
        var next = document.selectFirst(".next")?.attr("href") ?: ""
        
        while (next.isNotEmpty()) {
            val nextDocument = client.newCall(GET(next, headers)).execute().asJsoup()
            
            chapters += nextDocument.select(chapterListSelector())
                .map{ chapterFromElement(it) }
            next = nextDocument.selectFirst(".next")?.attr("href") ?: ""
        }
        
        return chapters.mapIndexed { i, chapter ->
            chapter.apply { chapter_number = chapters.size.toFloat() - i }
        }
    }
```

Please test the code, as I wrote it directly on the GitHub review.

src/en/randowiz/src/eu/kanade/tachiyomi/extension/en/randowiz/Randowiz.kt
```diff
@@ -0,0 +1,154 @@
...
+                false
+            )
+        )
+    }
+
+    override fun fetchSearchManga(
+        page: Int,
+        query: String,
+        filters: FilterList
+    ): Observable<MangasPage> = Observable.just(MangasPage(emptyList(), false))
```
alessandrojean, line 77: Can't you implement a basic search? Even if it has only three hardcoded titles, it's good to have in case someone eventually wants to migrate to this extension, for example. To do that, you can get the list calling `fetchPopularManga` and map the result to filter by the `title.contains(query, ignoreCase = true)`.

alessandrojean (changes requested):
Overall, it seems good for a first extension 👍🏻.

Just suggested a few things to make the extension code and usability better.

src/en/randowiz/src/eu/kanade/tachiyomi/extension/en/randowiz/Randowiz.kt
ericlp: Neat, first time using kotlin and I was thinking about doing this but couldn't quite figure out how. Your suggestions compiled and worked perfectly! 👌

src/en/randowiz/src/eu/kanade/tachiyomi/extension/en/randowiz/Randowiz.kt
ericlp: It is now implemented, don't know where thwy would migrate from though, these don't exist in any other extension. 🤷

alessandrojean: It was just an example, haha.

alessandrojean (approved):

src/en/randowiz/src/eu/kanade/tachiyomi/extension/en/randowiz/Randowiz.kt
```diff
@@ -0,0 +1,166 @@
...
+        while (next.isNotEmpty()) {
+            val nextDocument = client.newCall(GET(next, headers)).execute().asJsoup()
+
+            chapters += nextDocument.select(chapterListSelector())
+                .map { chapterFromElement(it) }
+            next = nextDocument.selectFirst(".next")?.attr("href") ?: ""
+        }
+
+        return chapters.mapIndexed { i, chapter ->
+            chapter.apply { chapter_number = chapters.size.toFloat() - i }
```
AntsyLich, line 106: We should avoid this cause the author might put some X.5 chapters

src/en/randowiz/src/eu/kanade/tachiyomi/extension/en/randowiz/Randowiz.kt
```diff
@@ -0,0 +1,166 @@
...
+
+    override fun chapterListSelector() = ".has-post-thumbnail"
+
+    override fun chapterFromElement(element: Element): SChapter {
+        val linkTag = element.select(".elementor-post__title a").first()
+        val chapter = SChapter.create()
+        chapter.name = linkTag.text()
+        chapter.url = linkTag.attr("href").split(".com")[1]
+        chapter.date_upload = parseDate(element.select(".elementor-post-date").text())
+        return chapter
```
AntsyLich, line 118:
```suggestion
        return SChapter.create().apply {
            name = linkTag.text()
            setUrlWithoutDomain(linkTag.attr("href"))
            date_upload = parseDate(element.select(".elementor-post-date").text())
        }
```

AntsyLich (changes requested):
Already been merged but oh well

src/en/randowiz/src/eu/kanade/tachiyomi/extension/en/randowiz/Randowiz.kt
alessandrojean: Btw, is the chapter_number set like this to reverse the list? If so, can just use `chapters.reversed()`.

alessandrojean:
Lich suggestions are doable, if you eventually have time and can open a future PR we appreciate it.

And sorry @AntsyLich, merged the PR too quickly 😆.

Issue #13749, opened 2022-10-07
Komga: Add ability to map existing titles to a new base url.


THElegend5:
### Source name and language

Komga (Multi)

### Describe your suggested feature

Add an option in the extension settings in which you can add a new base url and it basically migrates all existing titles to the new url and makes the new url the default for komga. 

### Other details

I use komga on the local network which results in the base url (the internal ip) changing once a month or so and the options are to either migrate everything manually to the new url which is not really feasible anymore or to just leave them be and migrate as needed and potentially add duplicates to library if by chance I change the name of the entry in komga and forget about it, so this would be pretty nice to have feature.

### Acknowledgements

- [X] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open or closed issue.
- [X] I have written a short but informative title.
- [X] If this is an issue with the app itself, I should be opening an issue in the [app repository](https://github.com/tachiyomiorg/tachiyomi/issues/new/choose).
- [X] I have updated the app to version **[0.13.6](https://github.com/tachiyomiorg/tachiyomi/releases/latest)**.
- [X] I will fill out all of the requested information in this form.

curche:
Relevant code: https://github.com/tachiyomiorg/tachiyomi-extensions/blob/master/src/all/komga/src/eu/kanade/tachiyomi/extension/all/komga/Komga.kt#L274

Maybe it can be done similar to how Manatoki ext does it
Manatoki (& co) is a source which is infamous for changing URLs so often that it has an extension setting to just change base url while not needing to migrate (atleast afaik)
 - setting manga.url using getUrlPath : https://github.com/tachiyomiorg/tachiyomi-extensions/blob/db71211623b7b2b707d13bf861068fdb2bc7ee1f/src/ko/newtoki/src/eu/kanade/tachiyomi/extension/ko/newtoki/NewToki.kt#L57
 - getUrlPath: https://github.com/tachiyomiorg/tachiyomi-extensions/blob/db71211623b7b2b707d13bf861068fdb2bc7ee1f/src/ko/newtoki/src/eu/kanade/tachiyomi/extension/ko/newtoki/NewToki.kt#L265-L269

stevenyomi:
> Maybe it can be done similar to how Manatoki ext does it
> Manatoki (& co) is a source which is infamous for changing URLs so often that it has an extension setting to just change base url while not needing to migrate (atleast afaik)

[...]

PR #14126, opened 2022-11-04
Refactor the Komga code

alessandrojean:
Changelog:
- App restart not needed any more when preferences are changed
- Fix empty filter groups if the fetching failed
- Fix outdated filter values if preferences are changed
- Extension code refactored to make it easier to change in new updates

Checklist:


alessandrojean:
In my tests it seems to be working as expected, but any additional test by someone that uses more often the extension would be appreciated. Also, @gotson, the people from Tachi's Discord told me it would be good to also mark you, so you can point if I missed something during the refactor.

src/all/komga/src/eu/kanade/tachiyomi/extension/all/komga/KomgaDto.kt
```diff
@@ -0,0 +1,237 @@
...
+    val id: String,
+    val name: String,
+    val summary: String,
+    val bookIds: List<String>,
+    val createdDate: String,
+    val lastModifiedDate: String,
+    val filtered: Boolean
+) {
+
+    fun toSManga(baseUrl: String): SManga = SManga.create().apply {
```
gotson, line 230: I find it bad practice to put the mapping logic inside the DTO. DTO are purely for data transfer, and should not hold any logic whatsoever.

src/all/komga/src/eu/kanade/tachiyomi/extension/all/komga/KomgaDto.kt
```diff
@@ -0,0 +1,237 @@
+package eu.kanade.tachiyomi.extension.all.komga
```
gotson, line 1: why move the DTO classes into the main folder and not keep them where they were?

src/all/komga/src/eu/kanade/tachiyomi/extension/all/komga/KomgaDto.kt
```diff
@@ -0,0 +1,237 @@
...
+    val name: String,
+    val created: String?,
+    val lastModified: String?,
+    val fileLastModified: String,
+    val booksCount: Int,
+    val metadata: SeriesMetadataDto,
+    val booksMetadata: BookMetadataAggregationDto
+) {
+
+    fun toSManga(baseUrl: String): SManga = SManga.create().apply {
```
gotson, line 41: I find it bad practice to put the mapping logic inside the DTO. DTO are purely for data transfer, and should not hold any logic whatsoever.

src/all/komga/src/eu/kanade/tachiyomi/extension/all/komga/KomgaDto.kt
```diff
@@ -0,0 +1,237 @@
...
+        }
+
+        private val DATE_TIME_ISO_8601_FORMATTER by lazy {
+            SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss", Locale.US)
+        }
+
+        private val DATE_TIME_FULL_ISO_8601_FORMATTER by lazy {
+            SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss.S", Locale.US)
+        }
+    }
```
gotson, line 169: Similar comment here, DTOs should not hold any business logic

src/all/komga/src/eu/kanade/tachiyomi/extension/all/komga/KomgaDto.kt
alessandrojean: It's a common pattern some extensions are using, so I just replicated it. But sure, I can move these to a separated file and make them extension functions, in that way they don't make the main class with a huge line count like it was before.

In relation to the `dto` package as stated in the next comment, I can probably also move them back there, but separate into some files by some common groups like the MangaDex extension does ([reference](https://github.com/tachiyomiorg/tachiyomi-extensions/tree/master/src/all/mangadex/src/eu/kanade/tachiyomi/extension/all/mangadex/dto)). In that way, we can even have the extension function that maps to Tachi's models in the same file to easy make any modifications in the future.

Alternatively, I can also do the MangaDex way and create a `KomgaHelper` that does these mappings too ([reference](https://github.com/tachiyomiorg/tachiyomi-extensions/blob/master/src/all/mangadex/src/eu/kanade/tachiyomi/extension/all/mangadex/MangaDexHelper.kt#L255-L284)). Which one do you prefer?

gotson: > In relation to the `dto` package as stated in the next comment, I can probably also move them back there, but separate into some files by some common groups like the MangaDex extension does ([reference](https://github.com/tachiyomiorg/tachiyomi-extensions/tree/master/src/all/mangadex/src/eu/kanade/tachiyomi/extension/all/mangadex/dto)).

No worries, when i have a few objects i tend to keep them in a single file, since Kotlin doesn't enforce having a class per file like Java does. That works well when classes have relationships, so you don't have to jump between files to know how that works, but i'm fine with one class per file.

> In that way, we can even have the extension function that maps to Tachi's models in the same file to easy make any modifications in the future.

For me the extension functions are a convenience, but are not part of the DTO, because they are ultimately part of the business logic. 

> Alternatively, I can also do the MangaDex way and create a `KomgaHelper` that does these mappings too ([reference](https://github.com/tachiyomiorg/tachiyomi-extensions/blob/master/src/all/mangadex/src/eu/kanade/tachiyomi/extension/all/mangadex/MangaDexHelper.kt#L255-L284)). Which one do you prefer? In a second thought, I think the Helper alternative would be better.

I honestly don't have a problem with files with huge line count. IDEs provide many ways of navigating such files. You can separate the extension function for mapping in a separate file, as Kotlin doesn't need a class to define functions.

alessandrojean: > You can separate the extension function for mapping in a separate file

I ended up in the end creating a helper object, but it's working the same as before in the end.

src/all/komga/src/eu/kanade/tachiyomi/extension/all/komga/KomgaDto.kt
alessandrojean: Moved back to a `dto` package now.

gotson:
> Why when creating the manga does it have an absolute url? Unless I am missing something I don' see the benefit. Downsides for it is if you don't have a static ip and your instance ip changes it force you to migrate all your manga.

It was done in 3cd9bb0abbbf33794f553e372ba4a91ef609deba because else backups could not be restored.

It's also required for the Komga tracker. As there is a single Komga tracker, but there can be up to 3 Komga extensions, the tracker needs to know the full URL to know where to track.

It's probably worth adding a comment in the extension code to explain why we store the full URL.

CarlosEsco:
" Because of that, restoring from a backup would generate some errors because URL would be missing http or https"


99% of extension urls are relative so this explanation on that commit doesn't make sense.  So I wonder if something else was going on.

The url really should have had a placeholder then relative path. Then the placeholder could be replaced with the actual preferences base url upon loading. 1/api/v1..

The tracker could have then done the same thing and looked up the preference from the placeholder.

gotson:
> The tracker could have then done the same thing and looked up the preference from the placeholder.

From what i know, the tracker doesn't know which extension "instance" the Manga is from, only that it's from the Komga extension. So it cannot retrieve the baseUrl for that particular instance.

gotson:
> That's why it should parse the manga url and pull out the placeholder. Then using the placeholder look up the actual base url. Same with the track entry. It should be creating a track db record with the placeholder in the URL.

i don't understand what you mean, could you provide an example ?

Let's say there are 2 Komga instances configured in the extension, with:
- https://komga.domain1.com
- https://komga.me.com

What would be in the Manga URL ?

And how can the tracker find the real URL from the Manga URL?

CarlosEsco:
> > That's why it should parse the manga url and pull out the placeholder. Then using the placeholder look up the actual base url. Same with the track entry. It should be creating a track db record with the placeholder in the URL.
> 
> i don't understand what you mean, could you provide an example ?
> 
> Let's say there are 2 Komga instances configured in the extension, with:
> - https://komga.domain1.com
> - https://komga.me.com
> 
> What would be in the Manga URL ?
> 
> And how can the tracker find the real URL from the Manga URL?

Sure on mobile so gonna use pseudo code.

So assuming the first url is on komga 1 and second komga 2

Manga url for instance 1->  /API/v1/rest-of-manga-url

Manga url for instance 2->  2/API/v1/rest-of-manga-url


Then in the extension you lookup the URL anytime you need to connect

If manga.url starts with / -> preference.getKomgaBaseUrl()

If manga.url starts with 2 -> preference.getKomgaBaseUrl(2)

Then you just concat manga.url with the retrieved base url.( Removing placeholders).  Do whatever action that was needed and you are done. 

When a user links the tracker to an entry 
The track.url would be the same as the manga.url and do the same lookup before each action.

gotson:
That's quite a big change, and probably outside the scope of this PR. It would also require some changes in the Tracker in the main repo, so there's a need to synchronize the 2 changes too.

src/all/komga/src/eu/kanade/tachiyomi/extension/all/komga/KomgaFilters.kt
```diff
@@ -0,0 +1,375 @@
+package eu.kanade.tachiyomi.extension.all.komga
+
+import android.content.SharedPreferences
+import eu.kanade.tachiyomi.network.GET
```
curche, line 4:
```suggestion
import android.content.SharedPreferences
import eu.kanade.tachiyomi.extension.all.komga.dto.AuthorDto
import eu.kanade.tachiyomi.extension.all.komga.dto.CollectionDto
import eu.kanade.tachiyomi.extension.all.komga.dto.LibraryDto
import eu.kanade.tachiyomi.network.GET
```

the build was [failing](https://github.com/tachiyomiorg/tachiyomi-extensions/actions/runs/3407399293/jobs/5667024436#step:4:877) because of the missing imports after moving DTOs

curche:
Tested this locally, found two things
 - failed build because of missing imports (which I fixed locally easily)
 - Collections don't seem to be loading correctly in Filter Menu

In my testing, everything else was working fine

curche:
> > That's quite a big change, and probably outside the scope of this PR. It would also require some changes in the Tracker in the main repo, so there's a need to synchronize the 2 changes too.
> 
> Definitely outside the scope was more confused a few weeks ago when someone on discord mentioned they had to migrate their own stuff on komga, when it shouldn't be the case

FYI @CarlosEsco @gotson there's an open issue for this: https://github.com/tachiyomiorg/tachiyomi-extensions/issues/13749
(I actually told them to create this after they talked abt this in Discord)

src/all/komga/src/eu/kanade/tachiyomi/extension/all/komga/KomgaFilters.kt
```diff
@@ -0,0 +1,378 @@
...
+            onError("error while loading libraries for filters", librariesResult.exceptionOrNull())
+        }
+
+        libraries = librariesResult.getOrDefault(emptyList())
+    }
+
+    private fun fetchCollections() {
+        val collectionsResult = runCatching {
+            val request = GET("$baseUrl/api/v1/collections?unpaged=true", headers)
+            client.newCall(request).execute().parseAs<List<CollectionDto>>()
```
curche, line 105:
```suggestion
            client.newCall(request).execute().parseAs<PaginatedResponseDto<CollectionDto>>().content
```

this fixes the collection filter
type-safety kinda didn't help here. It got type cast? 🤔 
anyways, should be PaginatedResponse (was previously [PageWrapperDto](https://github.com/tachiyomiorg/tachiyomi-extensions/blob/master/src/all/komga/src/eu/kanade/tachiyomi/extension/all/komga/Komga.kt#L542))

could make it slightly more readable by having a separate variable (eg: outputList.content)
also don't forget: import PaginatedResponse from dto package

alessandrojean: Thanks for the heads-up, will fix soon 👍🏻

alessandrojean:
Closing this for now as I don't have time to update some missing stuff and test it again for now.

Will likely revisit this in the future, though.

PR #14149, opened 2022-11-06
Use schedule instead of last added in Latest on Bilibili

alessandrojean:
Bilibili Manga is still using the search request as they do have a sort by update time.

Checklist:


multisrc/src/main/java/eu/kanade/tachiyomi/multisrc/bilibili/Bilibili.kt
```diff
@@ -85,15 +89,40 @@ abstract class Bilibili(
...
 
-    override fun latestUpdatesRequest(page: Int): Request = searchMangaRequest(
-        page = page,
-        query = "",
-        filters = FilterList(
-            SortFilter("", getAllSortOptions(), defaultLatestSort)
-        )
-    )
+    override fun latestUpdatesRequest(page: Int): Request {
+        val jsonPayload = buildJsonObject { put("day", dayOfWeek) }
```
AntsyLich, line 93: Imo we should overide this in bilibilicomics since it's missing the last updated sort

alessandrojean: #14150 likely.

PR #14161, opened 2022-11-07
[RU]Newbie cloudflareClient (fix 403)

e-shl:
Checklist:


src/ru/newbie/src/eu/kanade/tachiyomi/extension/ru/newbie/Newbie.kt
```diff
@@ -66,8 +68,10 @@ class Newbie : ConfigurableSource, HttpSource() {
 
     private var branches = mutableMapOf<String, List<BranchesDto>>()
 
+    private val userAgentRandomizer = "${Random.nextInt().absoluteValue}"
+
     override fun headersBuilder(): Headers.Builder = Headers.Builder()
-        .add("User-Agent", "Tachiyomi " + System.getProperty("http.agent"))
+        .add("User-Agent", "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/100.0.4896.127 Safari/537.36 Edg/100.0.$userAgentRandomizer")
```
alessandrojean, line 74: Isn't it better to just not set it and let users change through the new app setting?

e-shl: > Isn't it better to just not set it and let users change through the new app setting?

it's easier

alessandrojean: Adding a explicit user-agent (even with the random part) will allow the site owners to easily block on the Cloudflare panel though. Specially if they already have enabled the protection, it's very likely they will do that.

Not setting will make the extension use the value from the app preferences so users can change to whatever user agent string they want and the site owners can't block a specific one.

e-shl: They do not specifically block the tachiemi user-agent, it just started blocking the client for some, that is, it is necessary to periodically enter a captcha. I'm changing user-agent for a working bundle as on remanga.

alessandrojean: Fine, sure... but if they start blocking in the future, it's better to remove it next time and let the users choose one on the app.

PR #14322, opened 2022-11-20
Kavita: Bugfix/expected url scheme

ThePromidius:
This PR fixes an error when downloading that was apparently brought up in the latest preview version (r5254 and previous).

Latest preview version does now call `pageListRequest`  which from my understanding it shouldn't if `fetchPageList` is overriden.
I updated the method to return a dummy request to the chapter page since this call apparently does nothing in this use case.

I'm open to discussion if an issue should be raised in the main repo. Meanwhile here's valid fix for now (works with latest stable as well).

Note that this error only happens when downloading chapters.

Checklist:

### DIFF
--- a/src/all/kavita/CHANGELOG.md
+++ b/src/all/kavita/CHANGELOG.md
@@ -1,3 +1,9 @@
+## 1.3.8
+
+### Fix
+
+* Fixed `Expected URL scheme 'http' or 'https` when downloading
+
 ## 1.3.7
 
 ### Features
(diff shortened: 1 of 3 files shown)


stevenyomi:
The cause is the app using `getChapterUrl()` to get an URL for ComicInfo.xml generation. That method has a default implementation of calling `pageListRequest()` and can't be overriden until we get to extlib 1.4.

https://github.com/tachiyomiorg/tachiyomi/blob/e22896a9569b3dc8d8bc087a4765f8de2391a69b/app/src/main/java/eu/kanade/tachiyomi/data/download/Downloader.kt#L527

ThePromidius:
I see... Makes sense. Thanks for answering.
I'm guessing my fix is valid then for now.

PR #14382, opened 2022-11-25
update Madara User-Agent string

toqueteos:
Continuation of #12676

Should help with CloudFlare detection.

Firefox 107 released November 15, 2022
https://www.mozilla.org/en-US/firefox/107.0/releasenotes/

Checklist:


Seishirou101:
it may be better to just remove the user agent from madara tbh

alessandrojean:
What Seishirou said, it would be actually better to remove the randomization thing from the base class and overrides and let users use their own custom strings of the app settings if they're facing issues.

CarlosEsco:
Agreed

PR #14394, opened 2022-11-28
FlameScans: Fix permanent url logic

h-hyuuga:
Fixes #14216

Checklist:

### DIFF
--- a/multisrc/overrides/mangathemesia/flamescans/src/FlameScans.kt
+++ b/multisrc/overrides/mangathemesia/flamescans/src/FlameScans.kt
@@ -230,7 +230,7 @@ open class FlameScans(
 
         private const val PREF_PERM_CHAPTER_URL_KEY_PREFIX = "pref_permanent_chapter_url"
         private const val PREF_PERM_CHAPTER_URL_TITLE = "Permanent Chapter URL"
-        private const val PREF_PERM_CHAPTER_URL_SUMMARY = "Turns all chapter urls into permanent one."
+        private const val PREF_PERM_CHAPTER_URL_SUMMARY = "Turns all chapter urls into permanent ones."
 
         /**
          *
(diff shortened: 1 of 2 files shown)


multisrc/overrides/mangathemesia/flamescans/src/FlameScans.kt
```diff
@@ -243,7 +233,18 @@ open class FlameScans(
...
+         * De-obfuscates the slug of a series or chapter to the permanent slug
+         *   * For a series: "12345678-this-is-a-series" -> "this-is-a-series"
+         *   * For a chapter: "12345678-this-is-a-series-chapter-1" -> "this-is-a-series-chapter-1"
+         *
+         * @param obfuscated_slug the obfuscated slug of a series or chapter
+         *
+         * @return
+         */
+        private fun deobfuscateSlug(obfuscated_slug: String) = obfuscated_slug
+            .replaceFirst(Regex("""^\d+-"""), "")
```
AntsyLich, line 247: This'll have issues if series name starts with a number and url isn't temporary one

h-hyuuga: At the moment, that doesn't appear to be an issue: They only have one series that starts with a number and the slug uses the word (fourty-eight) instead of the decimal number.

In the event this changes in the future, this'll cause a 404 or an error and can be dealt with then.

multisrc/overrides/mangathemesia/flamescans/src/FlameScans.kt
```diff
@@ -243,7 +232,18 @@ open class FlameScans(
         private const val PREF_PERM_CHAPTER_URL_TITLE = "Permanent Chapter URL"
         private const val PREF_PERM_CHAPTER_URL_SUMMARY = "Turns all chapter urls into permanent one."
```
FlaminSarge, line 233: while we're at it, can fixup this line's grammar to "Turns all chapter urls into permanent ones."

h-hyuuga: Resolved by https://github.com/tachiyomiorg/tachiyomi-extensions/pull/14394/commits/669633b60aeb9d53cd6079fc3a76e2077330356f

PR #14633, opened 2022-12-22
MangaHub Factory: get Edo Tensei

Riztard:
- back to scrapping
- add user agent randomizer(latest chrome)
- make css selector more general
- use new pageListParse method
- add back some source: MangaHub.io, MangaPanda.onl, MangaReader.site

Checklist:


vetleledaal:
Does switching back to scraping fix the cookie issue, which prompted the removal in the first place? Test with some single chapter mange like `Tomo-chan wa Onnanoko` and confirm you can load 11 chapters within ~20 minutes.

Also update the issue moderator and removed sources list as appropriate.

Riztard:
nvm i just realize i have another issue with the chapter page only loading max 6

Riztard:
> Does switching back to scraping fix the cookie issue, which prompted the removal in the first place? Test with some single chapter mange like `Tomo-chan wa Onnanoko` and confirm you can load 11 chapters within ~20 minutes.

not sure how you test that. read 1 by 1 very slow?

> Also update the issue moderator and removed sources list as appropriate.

just leave it first just in case they broke it again intentionaly

vetleledaal:
>> confirm you can load 11 chapters within ~20 minutes.
>
> not sure how you test that. read 1 by 1 very slow?

Opposite, read 11 chapters fast. Within (less than) 20 minutes so you hit any rate limiting they put in. Should be doable in 1-2 minutes with manga that has a single page for each chapter.

multisrc/src/main/java/eu/kanade/tachiyomi/multisrc/mangahub/MangaHub.kt
```diff
@@ -251,11 +251,82 @@ abstract class MangaHub(
...
+
+            for (page in 1..maxPage) {
+                val url = urlTemplate.replaceAfterLast("/", "$page.$extension")
+                val pageObject = Page(page - 1, "", url)
+                pages.add(pageObject)
+            }
+        } else {
+
+            // if there is no max pages string make some calls to check if the pages exist
+            // increase or decreasing in big number to small to reduce call[ 50 25 10 4 1] if maxPages is 20+ ~200+
```
vetleledaal: The pattern you're looking for is binary search. That gives you the answer in O(log n) requests on average

Riztard:
> > not sure how you test that. read 1 by 1 very slow?
> 
> Opposite, read 11 chapters fast. Within (less than) 20 minutes so you hit any rate limiting they put in. Should be doable in 1-2 minutes with manga that has a single page for each chapter.

then it works been through 30 ch. maybe the rate limit  only for API

multisrc/src/main/java/eu/kanade/tachiyomi/multisrc/mangahub/MangaHub.kt
```diff
@@ -268,54 +292,10 @@ abstract class MangaHub(
...
-                            }
-                        }
-                    }
-                    break
-                }
-            }
+            // if there is no max pages string make some calls to check if the pages exist using findPageCount()
+            // increase or decreasing by using binary search algorithm
+            val headers = headersBuilder().build()
+            val maxPage = findPageCount(urlTemplate, extension, client, headers)
```
vetleledaal, line 298: you don't need to send in `client` or `headers`. should already be set in the class

multisrc/src/main/java/eu/kanade/tachiyomi/multisrc/mangahub/MangaHub.kt
```diff
@@ -251,6 +251,30 @@ abstract class MangaHub(
...
+
+        while (lowerBound <= upperBound) {
+            val midpoint = lowerBound + (upperBound - lowerBound) / 2
+            val url = urlTemplate.replaceAfterLast("/", "$midpoint.$extension")
+            val request = Request.Builder()
+                .url(url)
+                .headers(headers)
+                .head()
+                .build()
+            val response = client.newCall(request).execute()
```
vetleledaal, line 266: maybe wrap the call in a try/catch. So you can either show a better error to the user, or on error assume the image doesn't exist.

vetleledaal:
```suggestion
            val response = try {
                client.newCall(request).execute()
            } catch (e: IOException) {
                throw Exception("Failed to fetch $url")
            }
```

multisrc/src/main/java/eu/kanade/tachiyomi/multisrc/mangahub/MangaHub.kt
```diff
@@ -0,0 +1,416 @@
+package eu.kanade.tachiyomi.multisrc.mangahub
+
+import android.util.Log
```
vetleledaal, line 3: Unused import 😤

```suggestion
```

Riztard: ah forgot to remove that lol

alessandrojean:
Is this done for merge?

Riztard:
> Is this done for merge?

it should be, unless i/we miss something. atleast it works

PR #14688, opened 2022-12-27
Madara Custom and Random Latest User-Agent

Riztard:
- removing old hardcoded user agent randomizer
- add random latest user agent(off by default)
- add custom user agent(off by default)

Checklist:


vetleledaal:
> * need better way to check if `override val client` has `.addInterceptor(::uaIntercept)` or not (some source override this for rate limiting)

Was looking into something similar, but found another solution that works. Was something like:
```kt
.addInterceptor(network.cloudflareClient.interceptors.first { it.javaClass.simpleName == "CloudflareInterceptor" })
```

multisrc/src/main/java/eu/kanade/tachiyomi/multisrc/madara/Madara.kt
```diff
@@ -11,44 +17,104 @@ import eu.kanade.tachiyomi.source.model.SChapter
...
+     *      listOf("108")
+     */
+    protected open val filterIncludeUserAgent: List<String> = listOf()
+    protected open val filterExcludeUserAgent: List<String> = listOf()
+
+    private var userAgent: String? = null
+    private var checkedUa = false
+    private val hasUaIntercept = client.interceptors.any { interceptor -> interceptor.toString().contains(::uaIntercept.name) }
+
+    protected open fun uaIntercept(chain: Interceptor.Chain): Response {
```
alessandrojean, line 80: Better to also wrap in a `try-catch` in case the JSON format changes again.

```kotlin
try {
    // ...
} catch (e: Exception) {
    throw IOException(e.message)
}
```

multisrc/src/main/java/eu/kanade/tachiyomi/multisrc/madara/Madara.kt
```diff
@@ -11,44 +17,104 @@ import eu.kanade.tachiyomi.source.model.SChapter
...
+     *      listOf("chrome")
+     *      listOf("linux", "windows")
+     *      listOf("108")
+     */
+    protected open val filterIncludeUserAgent: List<String> = listOf()
+    protected open val filterExcludeUserAgent: List<String> = listOf()
+
+    private var userAgent: String? = null
+    private var checkedUa = false
+    private val hasUaIntercept = client.interceptors.any { interceptor -> interceptor.toString().contains(::uaIntercept.name) }
```
alessandrojean, line 78: Why not add it always and just check in the interceptor code if `useRandomUserAgentByDefault && userAgent == null && preference`?

I think there's no need to check if the interceptor was added. If there's a reason, I didn't understand.

Riztard: > I think there's no need to check if the interceptor was added. If there's a reason, I didn't understand.

sometimes `val client` got override to add rate limiting or other

alessandrojean: I see, makes sense.

Riztard:
> Was looking into something similar, but found another solution that works. Was something like:
> 
> ```kotlin
> .addInterceptor(network.cloudflareClient.interceptors.first { it.javaClass.simpleName == "CloudflareInterceptor" })
> ```
 found another way by making it into an object so the function name will appear in the string

alessandrojean (approved):

PR #14710, opened 2022-12-28
MangaHub: fix random user agent

mobi2002:
Fixes the json error recently introduced by the changes in user-agents.json 

Not sure if using `!!` is a good approach here as it will throw an `NullPointerException`  when `uaList["desktop"]` is null.

Checklist:


AntsyLich:
Rest assured `uaList["desktop"]` won't be null unless the format is changed again

And I fat fingered it

AntsyLich:
Btw update the comment to say chromium browsers since all chromium browsers has chrome in their ua

mobi2002:
should I also modify the REMOVED_SOURCES.md?

vetleledaal:
Go for it, was left out of the previous PR by mistake

PR #14784, opened 2023-01-03
Add team x back

CarlosEsco:
Had to rewrite everything since they changed the site since it was last here


src/ar/teamx/src/eu/kanade/tachiyomi/extension/ar/teamx/TeamX.kt
```diff
@@ -0,0 +1,192 @@
...
+
+            date_upload = parseChapterDate(element.select("div.epl-date").text())
+
+            setUrlWithoutDomain(element.attr("href"))
+        }
+    }
+
+    private fun parseChapterDate(date: String): Long {
+        return kotlin.runCatching {
+            SimpleDateFormat("yyyy-MM-dd hh:mm:ss", Locale.getDefault()).parse(date)?.time
```
AntsyLich, line 174: This needs to be a constant or it will be recreated for every chapter

PR #14828, opened 2023-01-06
Rename Tachidesk to Suwayomi

AriaMoradi:
Checklist:


src/all/suwayomi/build.gradle
```diff
@@ -3,10 +3,10 @@ apply plugin: 'kotlin-android'
...
 
 ext {
-    extName = 'Tachidesk'
-    pkgNameSuffix = 'all.tachidesk'
-    extClass = '.Tachidesk'
-    extVersionCode = 7
+    extName = 'Suwayomi'
+    pkgNameSuffix = 'all.suwayomi'
+    extClass = '.Suwayomi'
+    extVersionCode = 8
```
AntsyLich, line 9: Might as well start from 1 since you're changing the package name

src/all/suwayomi/build.gradle
```diff
@@ -3,10 +3,10 @@ apply plugin: 'kotlin-android'
 apply plugin: 'kotlinx-serialization'
 
 ext {
-    extName = 'Tachidesk'
-    pkgNameSuffix = 'all.tachidesk'
-    extClass = '.Tachidesk'
-    extVersionCode = 7
+    extName = 'Suwayomi'
+    pkgNameSuffix = 'all.suwayomi'
```
alessandrojean, line 7: Need to keep the same pkgName, otherwise users will not receive an update in their apps.

src/all/suwayomi/src/eu/kanade/tachiyomi/extension/all/suwayomi/Dto.kt
```diff
@@ -1,4 +1,4 @@
-package eu.kanade.tachiyomi.extension.all.tachidesk
+package eu.kanade.tachiyomi.extension.all.suwayomi
```
alessandrojean, line 1: Same here.

src/all/suwayomi/build.gradle
AriaMoradi: I'm in a pinch here, as a sovereign citizen, I can't name stuff Tachidesk

alessandrojean: Users will not see the package name so it can be kept.

PR #14843, opened 2023-01-08
HeanCms: Fix Search

seew3l:
Checklist:


multisrc/src/main/java/eu/kanade/tachiyomi/multisrc/heancms/HeanCms.kt
```diff
@@ -122,49 +122,84 @@ abstract class HeanCms(
...
+            .addQueryParameter("q", "filter")
             .toString()
 
         return POST(apiUrl, apiHeaders, payload)
     }
 
     override fun searchMangaParse(response: Response): MangasPage {
         val json = response.body?.string().orEmpty()
-        val query = response.request.url.queryParameter("q").orEmpty()
+        val searchType = response.request.url.queryParameter("q").orEmpty()
```
alessandrojean, line 153: It's better if you avoid passing that `q=filter` when you're not using the search. You can determine the type of handling by getting the last path on `response.request.url`.

multisrc/src/main/java/eu/kanade/tachiyomi/multisrc/heancms/HeanCms.kt
```diff
@@ -122,49 +122,84 @@ abstract class HeanCms(
...
-            mangaList = mangaList.filter { it.title.contains(query, ignoreCase = true) }
-        }
-
         fetchAllTitles()
 
         return MangasPage(mangaList, hasNextPage = false)
     }
 
+    private fun getSearchMangas(vSlug: String, vTitle: String): SManga {
+        val vThumbnail = seriesSlugMap?.get(vSlug)?.get(1) ?: ""
```
alessandrojean, line 186: Using a list makes things difficult to read, it's better if you create a `data class` to hold that information, which can later be incremented with more attributes if needed.

```kotlin
data class HeanCmsTitle(val slug: String, val thumbnailUrl: String, val status: String)
```

multisrc/src/main/java/eu/kanade/tachiyomi/multisrc/heancms/HeanCms.kt
```diff
@@ -122,49 +122,84 @@ abstract class HeanCms(
...
-        if (query.isNotBlank()) {
-            mangaList = mangaList.filter { it.title.contains(query, ignoreCase = true) }
-        }
-
         fetchAllTitles()
 
         return MangasPage(mangaList, hasNextPage = false)
     }
 
+    private fun getSearchMangas(vSlug: String, vTitle: String): SManga {
```
alessandrojean, line 185: Make it have a more descriptive name and name the parameters better as well.

```suggestion
    private fun createSearchResult(slug: String, title: String): SManga {
```

alessandrojean (changes requested):
Good approach, similar to what I would have done. Here are some requests to improve the code readability and maintainability.

alessandrojean:
I also forgot to add in the review, but better to add a warning in the filters mentioning the filters will be ignored if the query is not empty.

seew3l:
> I also forgot to add in the review, but better to add a warning in the filters mentioning the filters will be ignored if the query is not empty. In that way users will not think the filters are broken when trying to use them together with a query search.


How can I do that?

alessandrojean:
Use a `Filter.Header` and create a entry in the `intl` class so it can be translatable as well.

https://github.com/tachiyomiorg/tachiyomi-extensions/blob/master/CONTRIBUTING.md#filters

multisrc/src/main/java/eu/kanade/tachiyomi/multisrc/heancms/HeanCms.kt
```diff
@@ -136,28 +138,25 @@ abstract class HeanCms(
...
 
         return POST(apiUrl, apiHeaders, payload)
     }
 
     override fun searchMangaParse(response: Response): MangasPage {
         val json = response.body?.string().orEmpty()
-        val searchType = response.request.url.queryParameter("q").orEmpty()
 
-        if (searchType == "search") {
+        if (Uri.parse(response.request.url.toString()).lastPathSegment == "search") {
```
alessandrojean, line 154: You can get the pathSegments directly from the `url`, no need to parse it again through Uri.

https://square.github.io/okhttp/4.x/okhttp/okhttp3/-http-url/path-segments/

```kotlin
response.request.url.pathSegments.last()
```

multisrc/src/main/java/eu/kanade/tachiyomi/multisrc/heancms/HeanCms.kt
```diff
@@ -182,9 +181,9 @@ abstract class HeanCms(
         return MangasPage(mangaList, hasNextPage = false)
     }
 
-    private fun getSearchMangas(vSlug: String, vTitle: String): SManga {
-        val vThumbnail = seriesSlugMap?.get(vSlug)?.get(1) ?: ""
-        val vStatus = seriesSlugMap?.get(vSlug)?.get(2) ?: ""
+    private fun createSearchResult(vSlug: String, vTitle: String): SManga {
+        val vThumbnail = seriesSlugMap?.get(vSlug)?.thumbnailUrl ?: ""
```
alessandrojean: You're accessing the map twice, you can create a variable to access it once and then use it to get the thumbnail and status directly when creating the SManga, so no need to `vThumbnail` and `vStatus` in the end.

```kotlin
val theSeries = seriesSlugMap?.get(slug)
```

multisrc/src/main/java/eu/kanade/tachiyomi/multisrc/heancms/HeanCmsDto.kt
```diff
@@ -121,3 +121,10 @@ data class HeanCmsQuerySearchPayloadDto(
 data class HeanCmsSearchPayloadDto(
     val term: String,
 )
+
+@Serializable
```
alessandrojean, line 125: No need to be serializable, and it's better you add it inside the `HeanCms` class as this is technically not a DTO class.

multisrc/src/main/java/eu/kanade/tachiyomi/multisrc/heancms/HeanCmsIntl.kt
```diff
@@ -64,6 +64,12 @@ class HeanCmsIntl(lang: String) {
         else -> "Recently added"
     }
 
+    val filterWarning: String = when (availableLang) {
+        BRAZILIAN_PORTUGUESE -> "Os filtros serão ignorados se a consulta não estiver vazia."
```
alessandrojean, line 68: "Os filtros serão ignorados se a busca não estiver vazia."

multisrc/src/main/java/eu/kanade/tachiyomi/multisrc/heancms/HeanCmsIntl.kt
```diff
@@ -64,6 +64,12 @@ class HeanCmsIntl(lang: String) {
...
+        SPANISH -> "Los filtros se ignorarán si la consulta no está vacía."
```
alessandrojean, line 69: Change query to search while translating to Spanish and it should be good.

multisrc/src/main/java/eu/kanade/tachiyomi/multisrc/heancms/HeanCms.kt
```diff
@@ -182,9 +181,9 @@ abstract class HeanCms(
         return MangasPage(mangaList, hasNextPage = false)
     }
 
-    private fun getSearchMangas(vSlug: String, vTitle: String): SManga {
-        val vThumbnail = seriesSlugMap?.get(vSlug)?.get(1) ?: ""
-        val vStatus = seriesSlugMap?.get(vSlug)?.get(2) ?: ""
+    private fun createSearchResult(vSlug: String, vTitle: String): SManga {
```
alessandrojean: Can remove the `v` from the parameter names.

```suggestion
    private fun createSearchResult(slug: String, title: String): SManga {
```

alessandrojean (changes requested):
Just a minor tweaks this time and it should be ready.

multisrc/src/main/java/eu/kanade/tachiyomi/multisrc/heancms/HeanCms.kt
seew3l: I forgot and changed it in the other commit

multisrc/src/main/java/eu/kanade/tachiyomi/multisrc/heancms/HeanCms.kt
```diff
@@ -182,14 +181,13 @@ abstract class HeanCms(
     }
 
     private fun createSearchResult(slug: String, title: String): SManga {
-        val thumbnailUrl = seriesSlugMap?.get(slug)?.thumbnailUrl ?: ""
-        val status = seriesSlugMap?.get(title)?.status ?: ""
+        val theSeries = seriesSlugMap?.get(slug)
 
         return SManga.create().apply {
-            url = "/series/$slug"
+            url = "/series/${theSeries?.slug}"
```
alessandrojean, line 187: Just in case the title is not found (and can happen sometimes since their All series pages doesn't include some titles for some unknown reason), it's better to include the old slug instead of letting it empty.

```suggestion
            url = "/series/${theSeries?.slug ?: slug}"
```

Rest seems good now, just this minor one that needs to be fixed 👍🏻.

alessandrojean (changes requested):

multisrc/src/main/java/eu/kanade/tachiyomi/multisrc/heancms/HeanCms.kt
alessandrojean: Actually need to always be the `slug` since it doesn't include the timestamp that changes, otherwise it will make some titles not showing the "In library" badge if the users already have it on their libraries.

The slug in the map is used to create the requests only, but it should not be stored in the Tachiyomi database.

alessandrojean: Just fix that one mistake I suggested and it will be good to merge, sorry about it, but that's the last one now 👍🏻.

seew3l: So It should be
` url = "/series/$slug"` ?

alessandrojean (approved):

PR #14869, opened 2023-01-09
CuuTruyen: General improvements

beerpiss:
Mainly improvements in the unscrambling interceptor:
- Switched to a more performant and smaller `TextEncoder`/`TextDecoder` JS polyfill
- Pre-compiled the descrambler to QuickJS bytecode
- Transformations for unscrambling are casted directly instead of converting to and from JSON

Improvements to the source:
- `fetchSearchManga` adds `SManga.url` for id searches
- Activity handler now uses `CuuTruyen.PREFIX_ID_SEARCH` instead of hardcoding

Checklist:


src/vi/cuutruyen/src/eu/kanade/tachiyomi/extension/vi/cuutruyen/CuuTruyen.kt
```diff
@@ -43,8 +42,11 @@ class CuuTruyen : HttpSource(), ConfigurableSource {
         Injekt.get<Application>().getSharedPreferences("source_$id", 0x0000)
     }
 
-    override fun headersBuilder() = Headers.Builder()
-        .add("Referer", "$baseUrl/")
+    private val json = Json {
```
alessandrojean, line 45: The Json instance from the app already uses `ignoreUnknownKeys` so you can reuse it.

```kotlin
private val json: Json by injectLazy()
```

alessandrojean (changes requested):

PR #14898, opened 2023-01-11
HeanCms: Use source's timezone

seew3l:
Checklist:


multisrc/src/main/java/eu/kanade/tachiyomi/multisrc/heancms/HeanCmsDto.kt
```diff
@@ -94,9 +95,9 @@ data class HeanCmsChapterDto(
     @SerialName("created_at") val createdAt: String,
 ) {
 
-    fun toSChapter(seriesSlug: String): SChapter = SChapter.create().apply {
+    fun toSChapter(seriesSlug: String, timeZone: TimeZone): SChapter = SChapter.create().apply {
         name = this@HeanCmsChapterDto.name.trim()
-        date_upload = runCatching { DATE_FORMAT.parse(createdAt.substringBefore("."))?.time }
+        date_upload = runCatching { DATE_FORMAT.apply { this.timeZone = timeZone }.parse(createdAt.substringBefore("."))?.time }
```
alessandrojean, line 100: Would be better instead making the date format a class attribute instead of only the TimeZone. In that way the overrides can still set it and it will only be done once (instead of doing for every chapter when parsing).

alessandrojean (changes requested):

multisrc/src/main/java/eu/kanade/tachiyomi/multisrc/heancms/HeanCmsDto.kt
alessandrojean: Also need to add the `Z` to the date format pattern as well.

seew3l: > Also need to add the `Z` to the date format pattern as well.
But the sources don't use 'Z'

multisrc/src/main/java/eu/kanade/tachiyomi/multisrc/heancms/HeanCms.kt
```diff
@@ -381,12 +384,26 @@ abstract class HeanCms(
      */
     data class HeanCmsTitle(val slug: String, val thumbnailFileName: String, val status: Int)
 
+    class DateFormatter(timeZone: TimeZone) {
```
alessandrojean, line 387: That's not what I meant. I said to remove the timezone attribute and move the `DATE_FORMAT` to a class attribute (renamed to `dateFormat`) and make it `protected open val`. This class is useless.

alessandrojean (changes requested):

multisrc/src/main/java/eu/kanade/tachiyomi/multisrc/heancms/HeanCmsDto.kt
alessandrojean: Z represents the Timezone in the date string, at least Reaper uses that (it have a `"GMT+01"` in the end of the string.

seew3l: Ohhhhh I see. I ignored the part after `.` in e.g. `2023-01-10T18:58:59.738+01:00` because previously a substring is being made.

Should it be?
```Kotlin
SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss.SSSz", Locale.US)
```
Some `create_at` use +01:00 and others use +02:00
In that case can i just use?
```Kotlin
protected open val dateFormat: DateFormat = SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss.SSS'Z'", Locale.US)
```
and pass it to sChapter?

alessandrojean:
Closing in favour of #14900.

PR #14996, opened 2023-01-17
[YuriNeko] Fix error interceptor and URL intent filter

beerpiss:
Checklist:


src/vi/yurineko/src/eu/kanade/tachiyomi/extension/vi/yurineko/YuriNeko.kt
```diff
@@ -70,6 +62,18 @@ class YuriNeko : HttpSource() {
         }.build()
         return chain.proceed(authRequest)
     }
+    private fun errorIntercept(chain: Interceptor.Chain): Response {
+        val response = chain.proceed(chain.request())
+
+        if (response.code >= 400 && response.body != null) {
+            val error = kotlin.runCatching {
+                response.parseAs<ErrorResponseDto>()
+            }.getOrNull() ?: return response
```
stevenyomi, line 71:
```suggestion
            val error = try {
                response.parseAs<ErrorResponseDto>()
            } catch (_: Throwable) {
                return response
            }
```

stevenyomi: Also doesn't this consume the response body? If it's not an error this might break the response.

PR #15195, opened 2023-01-31
[skip ci] Ignore issue moderator and md changes in PR CI

scb261:
- Add `.github/` to the list of ignored changes. This won't start PR build at all if changes are only in `.github` or `*.md` files. Example of PR where all extensions were build even though it wasn't needed: https://github.com/tachiyomiorg/tachiyomi-extensions/pull/15193.
- Explicitly ignore `.md` files and `.github` folder for cases when issue moderator or a markdown file is edited along with code changes, for example when removing a source. Example of PR where removing an individual extension caused to build all extensions: https://github.com/tachiyomiorg/tachiyomi-extensions/pull/14974. I also made a test PR with old workflow that shows that ignored path is still included in the list of changed files: https://github.com/scb261/tachiyomi-extensions/pull/7.

Justification for ignoring the `.github` folder:
- issue and PR templates, readme images and renovate config aren't used in builds
- ~~changing workflow files in PR doesn't affect the actual workflow until it's merged~~ this is wrong, read comment below

Couple of test PRs:
- https://github.com/scb261/tachiyomi-extensions/pull/5
- https://github.com/scb261/tachiyomi-extensions/pull/6

In case someone is wondering - `true` in bash is equivalent to `pass` in Python, i.e. do nothing.


scb261:
Hm, actually I think it used my PR's workflow to ignore the PR itself. It wasn't [skip ci] because it must be in the commit message, not PR title ([docs](https://docs.github.com/en/actions/managing-workflow-runs/skipping-workflow-runs), examples where [skip ci] didn't skip: https://github.com/scb261/tachiyomi-extensions/pull/12, https://github.com/tachiyomiorg/tachiyomi-extensions/pull/15193). Guess I'll just exclude issue moderator workflow only, since it's the most updated file there that should be ignored.

scb261:
Welp, it used `ignore-paths` from my PR, but still runs the repo's workflow (probably to avoid malicious code in PRs that change workflows). I'll wait for someone to tell if I should ignore the whole `.github` folder (since it mostly doesn't affect PR builds), or only issue moderator (since changing `ignore-paths` does affect PR build check).

stevenyomi:
Probably just the issue moderator file. We may still want to check if the CI passes when actions dependencies are updated.

But most of the time that file is updated together with the removal (or re-addition) of an extension so CI is still triggered.

scb261:
> But most of the time that file is updated together with the removal (or re-addition) of an extension so CI is still triggered.

Yes, but it's counted as a root change and triggers both multisrc and individual extensions build.

PR #15303, opened 2023-02-10
Hiperdex baseUrl option

### LINKED ISSUE #15302: Hiperdex: Add domain switcher
### Source name and language

Hiperdex (English)

Hiperdex (from Madara multisrc) frequently switches back and forth between domains (https://hiperdex.com and https://1sthiperdex.com). A domain switcher would allow users to change it from within the app without needing an extension update

Empire Webtoon from Madara also has this https://github.com/tachiyomiorg/tachiyomi-extensions/pull/12887/commits/bc42f4070c46a0fae0558bf33fab94fd56b6a5f7

### Other details

_No response_

### Acknowledgements

Basara-Hatake:
Closes #15302

Checklist:


multisrc/overrides/madara/hiperdex/src/Hiperdex.kt
```diff
@@ -1,7 +1,56 @@
...
+    override val baseUrl by lazy { getPrefBaseUrl() }
+
+    private val preferences: SharedPreferences by lazy {
+        Injekt.get<Application>().getSharedPreferences("source_$id", 0x0000)
+    }
+
+    companion object {
+        private const val RESTART_TACHIYOMI = "Restart Tachiyomi to apply new setting."
+        private const val BASE_URL_PREF_TITLE = "Override BaseUrl"
+        private val BASE_URL_PREF = "overrideBaseUrl_v${AppInfo.getVersionName()}"
```
stevenyomi, line 27: This won't work as expected because `AppInfo.getVersionName()` returns the app's version name.

multisrc/overrides/madara/hiperdex/src/Hiperdex.kt
```diff
@@ -1,7 +1,56 @@
...
+        private const val BASE_URL_PREF_SUMMARY = "For temporary uses. Updating the extension will erase this setting."
+    }
+
+    override fun setupPreferenceScreen(screen: PreferenceScreen) {
+        val baseUrlPref = androidx.preference.EditTextPreference(screen.context).apply {
```
stevenyomi, line 32: Not sure if a list preference is better here.

multisrc/overrides/madara/hiperdex/src/Hiperdex.kt
```diff
@@ -1,7 +1,56 @@
...
+            key = BASE_URL_PREF_TITLE
```
stevenyomi, line 33:
```suggestion
            key = BASE_URL_PREF
```

multisrc/overrides/madara/hiperdex/src/Hiperdex.kt
```diff
@@ -1,7 +1,56 @@
...
+            title = BASE_URL_PREF_TITLE
+            summary = BASE_URL_PREF_SUMMARY
+            this.setDefaultValue(defaultBaseUrl)
+            dialogTitle = BASE_URL_PREF_TITLE
+
+            setOnPreferenceChangeListener { _, newValue ->
+                try {
+                    val res = preferences.edit().putString(BASE_URL_PREF, newValue as String).commit()
```
stevenyomi, line 41: This line is not needed. To eliminate the whole change listener thing, put the restart prompt in the summary.

multisrc/overrides/madara/hiperdex/src/Hiperdex.kt
```diff
@@ -1,7 +1,56 @@
...
+                    res
+                } catch (e: Exception) {
+                    e.printStackTrace()
+                    false
+                }
+            }
+        }
+        screen.addPreference(baseUrlPref)
+
+        super.setupPreferenceScreen(screen)
```
stevenyomi, line 52: I don't think there is a super implementation of this. Same applies to the edit to EmpireWebtoon.

stevenyomi (changes requested):
I know you're copying the code from another extension, but here are the issues. Either fix them for this extension, or fix for both/all extensions.

multisrc/overrides/madara/hiperdex/src/Hiperdex.kt
stevenyomi: Then remove `ConfigurableSource` from the class declaration, as Madara already implements it.

multisrc/overrides/madara/hiperdex/src/Hiperdex.kt
Basara-Hatake: how do i get the extension version code then

multisrc/overrides/madara/hiperdex/src/Hiperdex.kt
Basara-Hatake: list preference doesnt make any sense

multisrc/overrides/madara/hiperdex/src/Hiperdex.kt
Basara-Hatake: the restart toast is done everywhere like this

stevenyomi: I'm referring to the `edit()` line. That line is not needed.

Removing that line causes the change listener to only do the toast thing. TBH toasts are easy neglected, so putting the prompt into summary is a good idea.

stevenyomi:
According to the original issue, the URL of this site just jumps back and forth between two alternatives. I doubt if a list/switch preference to choose from, or even better, automatic detection without settings is better.

multisrc/overrides/madara/hiperdex/src/Hiperdex.kt
stevenyomi: Use `BuildConfig.VERSION_CODE` and make sure it's a `private const val`. Your IDE might complain not finding it, but it's generated on build and will pass the build.

stevenyomi: Also make sure your import line is `import eu.kanade.tachiyomi.extension.BuildConfig` otherwise it will be something else.

multisrc/overrides/madara/hiperdex/src/Hiperdex.kt
```diff
@@ -1,7 +1,55 @@
...
+        val baseUrlPref = androidx.preference.EditTextPreference(screen.context).apply {
+            key = BASE_URL_PREF
+            title = BASE_URL_PREF_TITLE
+            summary = BASE_URL_PREF_SUMMARY
+            this.setDefaultValue(defaultBaseUrl)
+            dialogTitle = BASE_URL_PREF_TITLE
+
+            setOnPreferenceChangeListener { _, newValue ->
+                try {
+                    val res = preferences.edit().putString(BASE_URL_PREF, newValue as String).commit()
```
stevenyomi, line 40:
```suggestion
```

This line still needs to be removed.

PR #15312, opened 2023-02-11
[skip ci] Bump dependencies

arkon:
Skipping OkHttp and JSoup. These match what's in v0.14.3.


alessandrojean:
The Kotlin update also made using the `else` branch on `when` mandatory. Some extensions don't have it while looping through the search filters, so need to add to these as well to avoid build errors.

arkon:
Yeah, I'm working through those locally.

PR #15366, opened 2023-02-15
ReaderFront: Fix crash when reading a chapter

### LINKED ISSUE #15373: Scylla Scan's Chapters Crashes the App

Scylla Scans 1.4.8

### Steps to reproduce

1. Open the App
2. Select a manga/manhwa from Scylla scans
3. It can be downloaded or not
4. App immediately crashes

### Expected behavior

Chapters should open normally without any issues 

### Actual behavior

App crashes 

### Tachiyomi version

Stable 0.14.4

### Android version

Android 11

### Other details

https://user-images.githubusercontent.com/87140863/219410306-0718a0f5-d729-414f-b99a-0e1398c0495e.mp4

### Acknowledgements

seew3l:
**Users will probably lose read chapters**

Checklist:

Also implement a way to get the url of the chapter

Closes #15373


multisrc/src/main/java/eu/kanade/tachiyomi/multisrc/readerfront/ReaderFront.kt
```diff
@@ -88,15 +95,24 @@ abstract class ReaderFront(
...
         }
 
-    override fun chapterListParse(response: Response) =
-        response.parse<List<Release>>("chaptersByWork").map {
+    override fun chapterListParse(response: Response): List<SChapter> {
+        val stub = response.request.url.fragment ?: ""
+        return response.parse<List<Release>>("chaptersByWork").map {
             SChapter.create().apply {
-                url = it.id.toString()
+                val jsonObject = JsonObject(
```
alessandrojean, line 102: Use `buildJsonObject`

multisrc/src/main/java/eu/kanade/tachiyomi/multisrc/readerfront/ReaderFront.kt
```diff
@@ -42,10 +46,13 @@ abstract class ReaderFront(
         GET("$apiUrl?query=${work(i18n.id, manga.url)}", headers)
 
     override fun chapterListRequest(manga: SManga) =
-        GET("$apiUrl?query=${chaptersByWork(i18n.id, manga.url)}", headers)
+        GET("$apiUrl?query=${chaptersByWork(i18n.id, manga.url)}#${manga.url}", headers)
 
-    override fun pageListRequest(chapter: SChapter) =
-        GET("$apiUrl?query=${chapterById(chapter.url.toInt())}", headers)
+    override fun pageListRequest(chapter: SChapter): Request {
+        val jsonObj = json.parseToJsonElement(chapter.url) as JsonObject
```
alessandrojean, line 52: Should do like this instead:

```kotlin
val jsonObj = json.parseToJsomElement(chapter.url).jsonObject
val id = jsonObj["id"]!!.jsonPrimitive.content
```

multisrc/src/main/java/eu/kanade/tachiyomi/multisrc/readerfront/ReaderFront.kt
```diff
@@ -107,8 +123,20 @@ abstract class ReaderFront(
 
     override fun getMangaUrl(manga: SManga) = "$baseUrl/work/$lang/${manga.url}"
 
-    override fun getChapterUrl(chapter: SChapter) =
-        throw UnsupportedOperationException("Not implemented!")
+    override fun getChapterUrl(chapter: SChapter): String {
+        val jsonObj = json.parseToJsonElement(chapter.url) as JsonObject
```
alessandrojean, line 127: Same here, use the `.jsonObject` attribute instead of casting.

multisrc/src/main/java/eu/kanade/tachiyomi/multisrc/readerfront/ReaderFront.kt
```diff
@@ -107,8 +123,20 @@ abstract class ReaderFront(
...
+        return "$baseUrl/read/${jsonObj["stub"].toString().removeSurrounding("\"")}/$lang/${jsonObj["volume"]}/${formatNumber(chapter.chapter_number)}"
+    }
+
+    private fun formatNumber(num: Float): String {
```
alessandrojean, line 131: Why not use String.format?

alessandrojean (changes requested):

multisrc/src/main/java/eu/kanade/tachiyomi/multisrc/readerfront/ReaderFront.kt
seew3l: I thought of using
```Kotlin
String.format("%.1f", num)
```
But what would happen if it has more than one decimal?

alessandrojean: The .1f part makes sure it will always shows just as one decimal even if it has more.

seew3l: I mean, it may be the case that the page uses 2 decimal places (for the chapter url they concatenate "chapter.subchapter"). Otherwise I can send the chapter and subchapter by the json and not use chapter_number

alessandrojean: Ah I see the issue. Should be fine to let it as if is for now then.

ObserverOfTime (commented):
I can't reproduce the crash. Does it happen on any particular chapter?

Also, if you change the chapter URLs, you ought to bump the `versionId` (I think).

alessandrojean:
> Also, if you change the chapter URLs, you ought to bump the versionId (I think).

Not really needed in this case, it's better to bump if the manga URLs are changed to force users to migrate, but on the chapter list only it's not needed.

alessandrojean (approved):

seew3l:
> I can't reproduce the crash. Does it happen on any particular chapter?

The crash happens on `Throw UnsupportedOperationException("Not implemented!")` in `getChapterUrl`

The function is called when the chapter opens and apparently the app does not handle an exception in that place. 

*Discord users reported the bug

ObserverOfTime:
> The crash happens on `Throw UnsupportedOperationException("Not implemented!")` in `getChapterUrl`.

I thought you meant reading the chapter. Opening it in the browser won't work indeed.

ObserverOfTime:
Anyway, you don't need to use json for this. I'd separate the id from the path with `|` and take the corresponding substring. Should be more performant.

PR #15384, opened 2023-02-17
Guya: remove proxy support and updates for extensions-lib 1.4

mobi2002:
The Proxy features on [**Guya** are simply redirects to **cubari**](https://guya.cubari.moe/proxy/gist/OPM/).

And they only work on **Guya** itself, other sources in the mutlisrc like **Hachirumi** and [**Danke** return 500](https://danke.moe/proxy/gist/OPM/) for them.

Checklist:


multisrc/src/main/java/eu/kanade/tachiyomi/multisrc/guya/Guya.kt
```diff
@@ -74,62 +74,40 @@ abstract class Guya(
...
-                        proxyMangaDetailsParse(response, manga)
-                    }
-            }
-            else -> {
-                client.newCall(chapterListRequest(manga))
-                    .asObservableSuccess()
-                    .map { response ->
-                        mangaDetailsParse(response, manga)
-                    }
+        return client.newCall(chapterListRequest(manga))
```
ObserverOfTime, line 85:
```suggestion
        return client.newCall(mangaDetailsRequest(manga))
```

PR #15466, opened 2023-02-24
Tachidesk (Suwayomi): Implement search

### LINKED ISSUE #14167: Tachidesk: Search functionality
### Source name and language

Tachidesk (All)

Currently, attempting to search for a manga causes the extension to display "Only Empty search is supported!", meaning that manga has to be found by manual scrolling, which is inefficient when dealing with hundreds of entries. Search should instead work like it does on the WebUI i.e. via the search endpoint exposed by Tachidesk-Server (or via client-side filtering, if viable).

### Acknowledgements

- [x] I have written a short but informative title.
- [X] I have updated the app to version **[0.14.2](https://github.com/tachiyomiorg/tachiyomi/releases/latest)**.

Sirver51:
Closes #14167

Checklist:

### DIFF
--- a/src/all/tachidesk/src/eu/kanade/tachiyomi/extension/all/tachidesk/Tachidesk.kt
+++ b/src/all/tachidesk/src/eu/kanade/tachiyomi/extension/all/tachidesk/Tachidesk.kt
@@ -122,14 +122,23 @@ class Tachidesk : ConfigurableSource, UnmeteredSource, HttpSource() {
 
     // ------------- Filters & Search -------------
 
+    private var categoryList: List<CategoryDataClass> = emptyList()
+
+    private val defaultCategoryId: Int
+        get() = categoryList.firstOrNull()?.id ?: 0
+
+    class CategorySelect(categoryList: List<CategoryDataClass>) :
+        Filter.Select<String>("Category", categoryList.map { it.name }.toTypedArray())
+
+    class DisableGlobalSearch() :
+        Filter.CheckBox("Search only current category", false)
+
     override fun getFilterList(): FilterList =
         FilterList(
             CategorySelect(refreshCategoryList(baseUrl).let { categoryList }),
             Filter.Header("Press reset to attempt to fetch categories"),
             DisableGlobalSearch(),
         )
-
-    private var categoryList: List<CategoryDataClass> = emptyList()
 
     private fun refreshCategoryList(baseUrl: String) {
         Single.fromCallable {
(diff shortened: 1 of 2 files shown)


jobobby04:
Why are you filtering in searchMangaRequest and not searchMangaParse? You could add the query into the request as a query parameter and grab it in the parse, and do the parsing in the background instead of blocking the UI by doing it in the request.

alessandrojean:
Yeah, don't do any network calls on the `request` methods, they're supposed to only return the Request that will be done by the `fetch` methods.

Sirver51:
Welp, I feel silly now; doing what you said reduces the amount of data being passed in the request and obviates the need for overriding fetchSearchManga. Should have caught that, my bad.

src/all/tachidesk/src/eu/kanade/tachiyomi/extension/all/tachidesk/Tachidesk.kt
```diff
@@ -191,11 +202,87 @@ class Tachidesk : ConfigurableSource, UnmeteredSource, HttpSource() {
...
+                    GET(categoryUrl, headers)
+                val categoryMangaListResponse =
+                    client.newCall(categoryMangaListRequest).execute()
+                val categoryMangaListJson = if ("gzip".equals(
+                        categoryMangaListResponse.header("Content-Encoding"),
+                        ignoreCase = true,
+                    )
+                ) {
+                    val categoryMangaListJsonGzip =
+                        GZIPInputStream(categoryMangaListResponse.body.byteStream())
```
alessandrojean, line 235: Can't you just use `response.body.string()`? I think OkHttp already handles Gzip encoding automatically.

src/all/tachidesk/src/eu/kanade/tachiyomi/extension/all/tachidesk/Tachidesk.kt
```diff
@@ -191,11 +202,87 @@ class Tachidesk : ConfigurableSource, UnmeteredSource, HttpSource() {
...
+                        }
+                        else -> {
+                            false
+                        }
+                    }
+                }
+            }.distinct()
+
+            // Construct new response with search results
+            val jsonString = json.encodeToString(
```
alessandrojean, line 268: Use the other method

```kotlin
json.encodeToString<List<MangaDataClass>>(searchResults)
```

I could be wrong and this could not exist at all, but if it does it requires an extra import IIRC. Alternatively it could be like this, but test as well:

```kotlin
json.encodeToString(searchResults)
```

src/all/tachidesk/src/eu/kanade/tachiyomi/extension/all/tachidesk/Tachidesk.kt
```diff
@@ -176,8 +180,15 @@ class Tachidesk : ConfigurableSource, UnmeteredSource, HttpSource() {
         Filter.Select<String>("Category", categoryList.map { it.name }.toTypedArray())
 
     override fun searchMangaRequest(page: Int, query: String, filters: FilterList): Request {
+        val request: Request
         if (query.isNotEmpty()) {
-            throw RuntimeException("Only Empty search is supported!")
+            // Embed search query for processing in searchMangaParse
+            request = Request.Builder()
```
alessandrojean, line 186: Could just return straightaway instead, no need for the variable.

You can also make it more optimized by sending the query through the fragment of the URL. By definition fragments are not sent to the server but they will be available in the `response.request.url.fragment` later.

By using the fragment you don't even need to use the Request builder, could just use the provided `GET` method by the extensions-lib directly.

src/all/tachidesk/src/eu/kanade/tachiyomi/extension/all/tachidesk/Tachidesk.kt
```diff
@@ -191,11 +202,87 @@ class Tachidesk : ConfigurableSource, UnmeteredSource, HttpSource() {
...
+        return request
     }
 
-    override fun searchMangaParse(response: Response): MangasPage = popularMangaParse(response)
+    override fun searchMangaParse(response: Response): MangasPage {
+        val request = response.request
+        val newResponse: Response
+        // Check if searchQuery header exists
+        val searchQuery = request.headers["searchQuery"]
+        if (searchQuery != null && searchQuery.isNotEmpty()) {
```
alessandrojean, line 215:
```suggestion
        if (!searchQuery.isNullOrEmpty()) {
```

src/all/tachidesk/src/eu/kanade/tachiyomi/extension/all/tachidesk/Tachidesk.kt
```diff
@@ -191,11 +202,87 @@ class Tachidesk : ConfigurableSource, UnmeteredSource, HttpSource() {
...
+                }
+                val categoryMangaList =
+                    json.decodeFromString<List<MangaDataClass>>(categoryMangaListJson)
+                mangaList.addAll(categoryMangaList)
+            }
+
+            // Filter according to search terms.
+            // Basic substring search, room for improvement.
+            val searchResults = mangaList.filter { mangaData ->
+                val fieldsToCheck = listOf<String?>(
```
alessandrojean, line 248: Can use `listOfNotNull`, so it will build a list without nullable strings and you can skip the null check later.

alessandrojean (changes requested):

src/all/tachidesk/src/eu/kanade/tachiyomi/extension/all/tachidesk/Tachidesk.kt
Sirver51: I did that because it wasn't working without it, but I may have confused it with another error - I'll check it out again.

src/all/tachidesk/src/eu/kanade/tachiyomi/extension/all/tachidesk/Tachidesk.kt
Sirver51: Android Studio marks both of these as a type mismatch on my machine; the second one is actually what I tried first. I will check to see if it builds in spite of the linter error, but any ideas on what's causing that?

src/all/tachidesk/src/eu/kanade/tachiyomi/extension/all/tachidesk/Tachidesk.kt
Sirver51: > Could just return straightaway instead, no need for the variable.

Since the request to be returned depends on the conditional, I figured it would be better to just have one return statement rather than one in each in the if and else. This is how I usually handle such a situation, but if what you said is more conformant with the expected style for this repo, I can change that.

> You can also make it more optimized by sending the query through the fragment of the URL. By definition fragments are not sent to the server but they will be available in the `response.request.url.fragment` later.

That makes sense, I'll make that change. Thanks for the info about fragments not going to the server, I wasn't aware of that.

src/all/tachidesk/src/eu/kanade/tachiyomi/extension/all/tachidesk/Tachidesk.kt
Sirver51: Sorry, not familiar with Kotlin; will make that change.

src/all/tachidesk/src/eu/kanade/tachiyomi/extension/all/tachidesk/Tachidesk.kt
Sirver51: Didn't know you could do that, will incorporate, thanks.

src/all/tachidesk/src/eu/kanade/tachiyomi/extension/all/tachidesk/Tachidesk.kt
jobobby04: That specific function needs to be imported, use the IDE autocomplete to find and select that function.

Sirver51:
I implemented the suggested changes, as well as an option to search a specific category rather than the whole library, and did some cleanup. I'm also halfway through an implementation for pagination (extension right now has trouble with large numbers of entries), but I have some questions that need answering, for which I will create a new issue.

Issue #15468, opened 2023-02-25
Asura Scans "No pages found"


mdpolson:
### Source information and language

Asura Scans v1.4.42

### Steps to reproduce

1. Install Asura Scans extension if not already installed
2. Select any series from Asura Scans
3. Attempt to download any chapter
4. Download will fail with error message "No pages found"

### Expected behavior

The download should succeed.

### Actual behavior

The download fails with error message "No pages found"

### Tachiyomi version

0.14.5

### Android version

Android 13 

### Acknowledgements

- [X] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open or closed issue.
- [X] I have written a short but informative title.
- [X] I have updated the app to version **[0.14.5](https://github.com/tachiyomiorg/tachiyomi/releases/latest)**.
- [X] I have updated all installed extensions.
- [X] I have tried the [troubleshooting guide](https://tachiyomi.org/help/guides/troubleshooting/).
- [X] If this is an issue with the app itself, I should be opening an issue in the [app repository](https://github.com/tachiyomiorg/tachiyomi/issues/new/choose).
- [X] I will fill out all of the requested information in this form.

Mondotosz:
I'm using TachiyomiJ2K v1.6.2 and also getting this issue with asura scans and reaper scans.
Tried it on TachiyomiSY v1.9.0 and it doesn't occur there.

asdkant:
Can confirm this happens with TachiJ2K v1.6.2 + asura v1.4.42 with both downloading and reading non-downloaded

leehwongxing:
Reading directly on Asura IS slow atm, but Reaper is still fast, but from the extension, Asura return no page and Reaper return extra yet non existing page.

rytt0001:
Had this problem on TachiyomiSy 1.9.0 with asurascan 1.4.42
Fixed it by resetting the User agent string in 
parameter ->advanced
If you want to know your user agent of your browser :
https://www.whatismybrowser.com/detect/what-is-my-user-agent/

leehwongxing:
> Had this problem on TachiyomiSy 1.9.0 with asurascan 1.4.42 Fixed it by resetting the User agent string in parameter ->advanced If you want to know your user agent of your browser : https://www.whatismybrowser.com/detect/what-is-my-user-agent/

unfortunately, this doesnt work on latest j2k build as i set the user-agent globally, and 1.4.* remove the ability to change user-agent on each extension

[...]

PR #15512, opened 2023-02-27
MangaThemesia: Add support for custom and random user agent

shabnix:
This PR will mitigate some Cloudflare bypass issues.

Most of the code from this PR is from: #14688
There isn't a toast when the cloudflareClient doesn't have the "uaIntercept" Interceptor, which is one difference from #14688. Not providing the option to set a custom or random user agent is (in my opinion) sufficient to indicate that custom and random user agent settings are not supported.

When the setting Use Random Latest User-Agent is enabled it will fix:
- #15468 

Checklist:


alessandrojean:
I don't think it's needed to add to every MangaThemesia source, you can add to only the ones that will benefit from it, like it was done in #14688.

shabnix:
I will update this pr tomorrow. Asura scans and Realm scans are two sources I know will benefit. Are there any other sources that will benefit as well?

alessandrojean:
I don't think so, at least for now. Should be fine to use these two as an initial test.

PR #15632, opened 2023-03-08
MangaDex: When opening a chapter link, also display a series consisting of only that chapter

lord-ne:
Currently, when opening a MangaDex chapter link in Tachiyomi, it just opens the series page. This is usually fine, however sometimes someone will post a link to a chapter without specifying which chapter it is, and if you have MangaDex set to open links in Tachiyomi by default, it can be annoying to try and figure out what chapter they were actually referring to.

This PR makes it so that when opening a chapter link in Tachiyomi, the app displays two different manga as search results. The first manga is the actual series that the chapter came from, as normal. The second manga contains only one chapter, which is the single chapter that the link was referring to.

I took inspiration from how the Dynasty Scans source has their "Dynasty Series" and "Dynasty Chapters" categories, and clicking a chapter link will display both the series and the chapter.

I've attached a demo to help clarify what I mean. I click on the link for Chapter 155 of the series, and two results come up: The actual manga that the chapter comes from, and another entry which contains only the relevant chapter.

https://user-images.githubusercontent.com/34069872/223622029-933044b7-9c74-4b6f-b557-665d6d914ba2.mp4

&nbsp;
If this PR is received positively, I intend to make something similar to open a specific single page of a manga.

Checklist:

- [x] **(n/a)** Updated `overrideVersionCode` or `baseVersionCode` as needed for all multisrc extensions
- [x] **(n/a)** Referenced all related issues in the PR body (e.g. "Closes #xyz")
- [x] **(n/a)** Added the `isNsfw = true` flag in `build.gradle` when appropriate
- [x] **(n/a)** Have explicitly kept the `id` if a source's name or language were changed

### DIFF
--- a/src/all/mangadex/build.gradle
+++ b/src/all/mangadex/build.gradle
@@ -6,7 +6,7 @@ ext {
     extName = 'MangaDex'
     pkgNameSuffix = 'all.mangadex'
     extClass = '.MangaDexFactory'
-    extVersionCode = 177
+    extVersionCode = 178
     isNsfw = true
 }
 
--- a/src/all/mangadex/src/eu/kanade/tachiyomi/extension/all/mangadex/MDConstants.kt
+++ b/src/all/mangadex/src/eu/kanade/tachiyomi/extension/all/mangadex/MDConstants.kt
@@ -125,6 +125,11 @@ object MDConstants {
         return "${tryUsingFirstVolumeCoverPref}_$dexLang"
     }
 
+    private const val chapterLinkSeriesPref = "chapterLinkSeries"
+    fun getChapterLinkSeriesPrefKey(dexLang: String): String {
+        return "${chapterLinkSeriesPref}_$dexLang"
+    }
+
     private const val tagGroupContent = "content"
     private const val tagGroupFormat = "format"
     private const val tagGroupGenre = "genre"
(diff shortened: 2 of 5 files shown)


CarlosEsco:
Not really a fan of this. It makes it seem like a different source, and now you have weird states of people being able to favorite a chapter "manga".   dynasty makes sense since that's how the sources are.

lord-ne:
> Not really a fan of this. It makes it seem like a different source, and now you have weird states of people being able to favorite a chapter "manga". dynasty makes sense since that's how the sources are.

It's true that it doesn't really make sense to favorite a chapter of a manga (although that still doesn't really make sense for Dynasty either, even though the source is like that). And I can see how it would be confusing for people if these extra search results just started appearing with no context. 

How would you feel if we had it enabled by a preference option, default off? That way users won't see it unless they explicitly enable it, which means they'll be expecting it and will understand what the second series in the search results means.

lord-ne:
> Waiting for a next extensions-lib as suggested is indeed a better thing to do.

Fair enough. I do still feel like there's very little downside to just adding this temporarily, but it seems like I'm in the minority there so I'll respect that.

> Discussions are opened in the app repository, but alternatively you can also open a feature request there as well.

Got it. I'll look into opening a feature request (or commenting on an existing one) when I have time.

alessandrojean:
Closing as resolved for now, but this feature in Dex is open to be reviewed in a eventual new extensions-lib version that supports chapter deeplinking.

lord-ne:
Opened https://github.com/tachiyomiorg/tachiyomi/issues/9188

PR #15782, opened 2023-03-21
add bakahm.com to multisrc

michaelzeng145:
Checklist:


multisrc/overrides/madara/bakamh/res/mipmap-hdpi/ic_launcher.png
arkon, line 1: This is just the default Madara logo, so I'm not sure it's that useful as an overridden icon.

multisrc/overrides/madara/bakamh/src/bakamh.kt
```diff
@@ -0,0 +1,14 @@
...
+
+import eu.kanade.tachiyomi.multisrc.madara.Madara
+import java.text.SimpleDateFormat
+import java.util.Locale
+
+class bakamh : Madara(
+    "巴卡漫画",
+    "https://bakamh.com/",
+    "zh",
+    SimpleDateFormat("yyyy 年 M 月 d 日", Locale.ENGLISH),
```
alessandrojean, line 11: Isn't the locale wrong? Since it has characters in Japanese/Chinese, it should likely match the same locale.

multisrc/overrides/madara/bakamh/res/mipmap-hdpi/ic_launcher.png
michaelzeng145: Thanks for the feedback. New logos committed to reflect the source site.

multisrc/overrides/madara/bakamh/src/bakamh.kt
michaelzeng145: Updated locale to Chinese.

stevenyomi: English is fine. We're using arabic numbers anyway.

multisrc/src/main/java/eu/kanade/tachiyomi/multisrc/madara/MadaraGenerator.kt
```diff
@@ -459,6 +459,7 @@ class MadaraGenerator : ThemeSourceGenerator {
         SingleLang("مانجا ليك", "https://mangalek.com", "ar", className = "Mangalek", overrideVersionCode = 2),
         SingleLang("مانجا لينك", "https://mangalink.io", "ar", className = "MangaLinkio", overrideVersionCode = 3),
         SingleLang("Chunchunmaru Translations", "https://chunchunmarutl.com", "es", className = "ChunchunmaruTranslations"),
+        SingleLang("巴卡漫画", "https://bakamh.com/", "zh", isNsfw = true, className = "bakamh"),
```
stevenyomi, line 462:
```suggestion
        SingleLang("巴卡漫画", "https://bakamh.com", "zh", isNsfw = true, className = "bakamh"),
```

multisrc/overrides/madara/bakamh/src/bakamh.kt
stevenyomi: Just a note: you only need to specify the right locale if you need to match CJK characters like 星期二 and (火), which means Tuesday in Chinese and Japanese, respectively.

The following example is from an old version, where I used `(E)` and `Locale.JAPANESE` to match that (火), though I didn't really need to because I could have just ignored that part entirely.

https://github.com/tachiyomiorg/tachiyomi-extensions/blob/e472d7972d2c62833d8115d2c06de3427e51f59c/multisrc/src/main/java/eu/kanade/tachiyomi/multisrc/comicgamma/ComicGamma.kt#L96

multisrc/overrides/madara/bakamh/src/bakamh.kt
```diff
@@ -0,0 +1,14 @@
+package eu.kanade.tachiyomi.extension.zh.bakamh
+
+import eu.kanade.tachiyomi.multisrc.madara.Madara
+import java.text.SimpleDateFormat
+import java.util.Locale
+
+class bakamh : Madara(
```
arkon, line 7: Can you capitalize this to `Bakamh` to make it consistent with other classes?

michaelzeng145:
I mistakenly pulled in all the changes from master by rebasing. How can I clean this up? Thanks

stevenyomi:
@michaelzeng145 `git reset --hard eee20d5` then `git push -f`.

stevenyomi:
Then make the necessary changes. You don't need to rebase or merge master if there's no conflict.

stevenyomi:
You also need to remove the trailing slash `/` from the URL (both in the generator and override file).

PR #15828, opened 2023-03-24
Update URL

wwbfred:
Checklist:

### DIFF
--- a/src/zh/wnacg/src/eu/kanade/tachiyomi/extension/zh/wnacg/wnacg.kt
+++ b/src/zh/wnacg/src/eu/kanade/tachiyomi/extension/zh/wnacg/wnacg.kt
@@ -18,7 +18,7 @@ import rx.Observable
 // URL can be found at https://www.wnacglink.top/
 class wnacg : ParsedHttpSource() {
     override val name = "紳士漫畫"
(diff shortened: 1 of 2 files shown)


src/zh/wnacg/src/eu/kanade/tachiyomi/extension/zh/wnacg/wnacg.kt
```diff
@@ -18,7 +18,7 @@ import rx.Observable
 // URL can be found at https://www.wnacglink.top/
 class wnacg : ParsedHttpSource() {
     override val name = "紳士漫畫"
-    override val baseUrl = "https://www.htcomic.top"
+    override val baseUrl = "http://www.htmanga3.top"
```
alessandrojean, line 21: Is there any specific reason to hot use HTTPS? It's better to always use it when possible.

wwbfred: Because there are no https sources right now. All servers I can find either 443 port has been closed or certificate is not right. I think they abandoned to support https for some reasons. I'm also confused about that.

stevenyomi:
Superseded by #15832

PR #15914, opened 2023-03-31
[skip ci] Issue templates: add report dead source and reorder

stevenyomi:
Language information is made into a separate field because many people didn't fill it out.


alessandrojean:
Maybe put the label as "Source language" just to make it more clear?

Issue #15877, opened 2023-03-28
Com-X: broken chapter names


AllLife2la:
### Source information and language

Com-X 1.4.23 russian

### Steps to reproduce

1. Find Zero Hour: Crisis in Time (with 75 chapters) or any other comic with untranslated chapters.
2. Download 2 or more untranslated chapters (they all call same as [перевода не существует])

### Expected behavior

1. First of all, chapters [перевода не существует] (it meens translation does not exist) must have ther full names, like on site:

On site and in app:
  <p>
    <img src="https://user-images.githubusercontent.com/8400957/228316716-eb99cf22-5158-4076-9723-5a9af1bb7b65.jpg" alt="On site" width="200">
    <img src="https://user-images.githubusercontent.com/8400957/228317138-480d7694-1be4-4cfb-a91e-1694d5cbf375.jpg" alt="In App" width="200">
  </p>

2. 2 or more chapters with these names must be downloaded.

### Actual behavior

1. Symbol "/" breaks chapters names, it throws out everything before it.
2. After first [перевода не существует] chapter downloaded app marks as downloaded every other same name chapter:
<img src="https://user-images.githubusercontent.com/8400957/228321997-4e25b8a5-3136-4746-a22c-6cda9b2153b1.jpg" width="200">
That makes it impossible to download comic.

### Tachiyomi version

r5653

### Android version

MIUI 12.5.1 - Android 11 Phone / EMUI 11.0.0 - Android 10 Tablet

### Other details

Continuation of [issue #9267](https://github.com/tachiyomiorg/tachiyomi/issues/9267#issue-1641119105).
Also we need keep chapter names same as comic name in crossover-events like in "Zero Hour: Crisis in Time" example.

### Acknowledgements

- [X] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open or closed issue.
- [X] I have written a short but informative title.
- [X] I have updated the app to version **[0.14.5](https://github.com/tachiyomiorg/tachiyomi/releases/latest)**.
- [X] I have updated all installed extensions.
- [X] I have tried the [troubleshooting guide](https://tachiyomi.org/help/guides/troubleshooting/).
- [X] If this is an issue with the app itself, I should be opening an issue in the [app repository](https://github.com/tachiyomiorg/tachiyomi/issues/new/choose).
- [X] I will fill out all of the requested information in this form.

PR #15945, opened 2023-04-03
[RU]ComX: Crossover-events handle, chapter naming cleanup

AllLife2la:
Cleanup for
#15877

Check my stuff for errors, first time with kotlin, first time doing stuff on github, kindof scarry.

Checklist:

### DIFF
--- a/src/ru/comx/src/eu/kanade/tachiyomi/extension/ru/comx/ComX.kt
+++ b/src/ru/comx/src/eu/kanade/tachiyomi/extension/ru/comx/ComX.kt
@@ -298,26 +298,31 @@ class ComX : ParsedHttpSource() {
 
         val data = json.decodeFromString<JsonObject>(dataStr)
         val chaptersList = data["chapters"]?.jsonArray
-        var isCrossoverEvent = false
 
-        // TODO: Clean up "for loop - if statement" mess somehow
-        for (i in chaptersList!!.indices) {
-            // if there are different chapter names before the chapter number, then it means it is a crossover-event
-            if (chaptersList[0].jsonObject["title_en"]!!.jsonPrimitive.content.split("#")[0] !=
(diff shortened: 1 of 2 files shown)


src/ru/comx/src/eu/kanade/tachiyomi/extension/ru/comx/ComX.kt
```diff
@@ -298,18 +298,32 @@ class ComX : ParsedHttpSource() {
...
-            chapter.name = it.jsonObject["title_en"]!!.jsonPrimitive.content
-            if (chapter.name.isEmpty()) {
-                chapter.name = it.jsonObject["title"]!!.jsonPrimitive.content
-            }
-            if (chapter.name == "[перевода не существует]") {
-                chapter.name = it.jsonObject["title"]!!.jsonPrimitive.content + " " + "[перевода не существует]"
+        var isCrossoverEvent = false
+
+        // TODO: Clean up "for loop - if statement" mess somehow
+        for (i in chaptersList!!.indices) {
```
alessandrojean, line 304: You could use the Kotlin `any` method instead.

```kotlin
val isCrossoverEvent = chaptersList.orEmpty().any { /* ... */ }
```

Reference: https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/any.html

alessandrojean (changes requested):

AllLife2la:
I just noticed that sometimes reading order numbers appear in comics that aren't events because they don't follow their naming pattern. So i have 2 options:
1. Check if it's an event and then check if it's a crossover-event, but that will make it mess of complicated code.
2. Just add reading order numbers to all events, even if it is not crossover. That will make code easy and clean. But we will have some feeleng of inconsistency. (They are already inconsistent on ComX, so it will only get worse)

Help me choose the option that you would like to see as a solution to the problem. Or suggest something.

PR #15978, opened 2023-04-06
CI: check multisrc-related files

stevenyomi:
- When `CHECK_MULTISRC_FILES` environment variable is set, multisrc generator will check missing IntelliJ configs under `.run/` and orphaned overrides, annotate them on [run result](https://github.com/tachiyomiorg/tachiyomi-extensions/actions/runs/4636721079) and [changed files](https://web.archive.org/web/20230407083913/https://github.com/tachiyomiorg/tachiyomi-extensions/pull/15978/files), and exit with error.
  - Internally it's using `git ls-files` so it still works if you're working on a sparse checkout and don't have all the overrides locally. You only need to have all the generators.
- Rewrote the gradle task with `JavaExec`.
- ~~Cleaned up all the orphaned overrides present and fixed the config files.~~ (split into #15993)

### DIFF
--- a/.github/workflows/build_pull_request.yml
+++ b/.github/workflows/build_pull_request.yml
@@ -66,7 +66,7 @@ jobs:
 
       - name: Generate multisrc sources
         if: ${{ steps.parse-changed-files.outputs.isMultisrcChanged == '1' }}
-        uses: gradle/gradle-command-action@v2
+        uses: gradle/gradle-build-action@v2
         env:
           CHECK_MULTISRC_FILES: 1
         with:
@@ -116,7 +116,7 @@ jobs:
           distribution: adopt
 
       - name: Generate sources from the multi-source library
-        uses: gradle/gradle-command-action@v2
+        uses: gradle/gradle-build-action@v2
         env:
           CI_MODULE_GEN: "true"
         with:
@@ -124,7 +124,7 @@ jobs:
           cache-read-only: true
 
       - name: Build extensions (chunk ${{ matrix.chunk }})
-        uses: gradle/gradle-command-action@v2
+        uses: gradle/gradle-build-action@v2
         env:
           CI_MULTISRC: "true"
           CI_CHUNK_NUM: ${{ matrix.chunk }}
@@ -150,7 +150,7 @@ jobs:
           distribution: adopt
 
       - name: Build extensions (chunk ${{ matrix.chunk }})
-        uses: gradle/gradle-command-action@v2
+        uses: gradle/gradle-build-action@v2
         env:
           CI_MULTISRC: "false"
           CI_CHUNK_NUM: ${{ matrix.chunk }}
--- a/.github/workflows/build_pull_request.yml
+++ b/.github/workflows/build_pull_request.yml
@@ -67,6 +67,8 @@ jobs:
       - name: Generate multisrc sources
         if: ${{ steps.parse-changed-files.outputs.isMultisrcChanged == '1' }}
         uses: gradle/gradle-command-action@v2
+        env:
+          CHECK_MULTISRC_FILES: 1
         with:
           arguments: :multisrc:generateExtensions
 
--- a/multisrc/src/main/java/generator/GeneratorMain.kt
+++ b/multisrc/src/main/java/generator/GeneratorMain.kt
@@ -25,4 +25,6 @@ fun main(args: Array<String>) {
                 }
                 .forEach { it.invoke(null, emptyArray<String>()) }
         }
+
+    ThemeSourceGenerator.checkOrphanedOverrides(userDir)
 }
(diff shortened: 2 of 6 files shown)


.github/workflows/build_pull_request.yml
```diff
@@ -66,7 +66,9 @@ jobs:
 
       - name: Generate multisrc sources
         if: ${{ steps.parse-changed-files.outputs.isMultisrcChanged == '1' }}
-        uses: gradle/gradle-command-action@v2
+        uses: gradle/gradle-build-action@v2
+        env:
+          CHECK_MULTISRC_FILES: 1
```
arkon, line 71: Why not just rely on `CI`?

multisrc/src/main/java/eu/kanade/tachiyomi/multisrc/mmrcms/MMRCMSJsonGen.kt
```diff
@@ -214,7 +213,7 @@ class MMRCMSJsonGen {
     )
 
     companion object {
-        val sources = sourceList
+        val sources = MMRCMSGenerator().sources
```
arkon, line 216: Could you do a separate PR with all of the source changes first?

multisrc/src/main/java/generator/GeneratorMain.kt
```diff
@@ -25,4 +26,9 @@ fun main(args: Array<String>) {
                 }
                 .forEach { it.invoke(null, emptyArray<String>()) }
         }
+
+    ThemeSourceGenerator.checkOrphanedOverrides(userDir)
+    if (ThemeSourceGenerator.isError && System.getenv("GITHUB_EVENT_NAME") == "pull_request") {
```
arkon, line 31: We'd ideally just pass actual arguments rather than leaking GitHub-specific logic into the generator.

multisrc/src/main/java/generator/ThemeSourceGenerator.kt
```diff
@@ -35,10 +35,98 @@ interface ThemeSourceGenerator {
 
     fun createAll() {
         val userDir = System.getProperty("user.dir")!!
+        checkRelatedFiles(userDir)
         sources.forEach { createGradleProject(it, themePkg, themeClass, baseVersionCode, userDir) }
     }
 
+    /** Checks IntelliJ configuration files under `.run/` and orphaned override folders. */
+    fun checkRelatedFiles(userDir: String) {
```
arkon, line 43: Checking's fine, but I wonder if we should just do the things (generating config file, deleting orphaned files) as part of a build anyway, sort of like auto formatting?

.github/workflows/build_pull_request.yml
stevenyomi: Because this only need to be done once in the "Prepare job", and don't need to be done in the matrix chunks.

multisrc/src/main/java/eu/kanade/tachiyomi/multisrc/mmrcms/MMRCMSJsonGen.kt
stevenyomi: Sure.

.github/workflows/build_pull_request.yml
arkon: Ought to just be a separate gradle task then?

multisrc/src/main/java/generator/GeneratorMain.kt
stevenyomi: The `printError` function is also GitHub logic too. It can't be avoided. But I can gather them into a single file, like `GithubApi`.

.github/workflows/build_pull_request.yml
stevenyomi: I'll try doing that.

multisrc/src/main/java/generator/ThemeSourceGenerator.kt
stevenyomi: Generating config file is fine (since it's separate for each generator).

But deleting orphaned files is not (it's run as a whole), as some themes have multiple generators (Webtoons specifically). To fix this, I can reorganize that generator and merge them into one.

However orphaned folders actually need to be investigated before deleting, so I don't think it's ideal to delete them by default. You see, typos will lead to orphaned overrides, and they need to be renamed instead of deleted (this also applies to later checks on smaller chunks).

multisrc/src/main/java/eu/kanade/tachiyomi/multisrc/mmrcms/MMRCMSJsonGen.kt
stevenyomi: MMRCMS changes have been split out. Other cleanups will be submitted after that one gets merged.

stevenyomi: Other cleanups split into #15993.

PR #16011, opened 2023-04-09
NewToki: update domain number on build

stevenyomi:
This eliminates the need to guess domain number and update it in code. The domain number is fetched at build time and written into the source file. We could write a fake number like `0` or `999` but this way it shows in the extension index better, and is more friendly to first-install users.

I also set up a new repo to host and auto-update domains: https://github.com/stevenyomi/source-domains (might add some more frequent-updating sites to it later). This can be moved to organization if needed. 

Checklist:


src/ko/newtoki/src/eu/kanade/tachiyomi/extension/ko/newtoki/FallbackDomainNumber.kt
```diff
@@ -0,0 +1,10 @@
+package eu.kanade.tachiyomi.extension.ko.newtoki
+
+/**
+ * This value will be automatically overwritten when building the extension.
+ * DO NOT commit the automatic updated file with git.
```
arkon, line 5: Could add a comment in the autogenerated file to make it obvious too.

src/ko/newtoki/src/eu/kanade/tachiyomi/extension/ko/newtoki/DomainNumber.kt
```diff
@@ -55,13 +39,10 @@ object DomainInterceptor : Interceptor {
             Log.e("NewToki", "failed to fetch ${request.url}", e)
 
             val newDomainNumber = try {
-                val document = chain.proceed(GET("https://t.me/s/newtoki5")).asJsoup()
-                val description = document.select("a[href^=https://newtoki]").last()!!.attr("href")
-                numberRegex.find(description)!!.value
+                val domainNumberUrl = "https://stevenyomi.github.io/source-domains/newtoki.txt"
+                chain.proceed(GET(domainNumberUrl)).body.string().also { it.toInt() }
```
arkon, line 43: Should it even bother doing it at runtime if we can semi-reliably rely on extension updates to update it? Not sure if it's worth removing the logic to simplify things.

arkon (commented):
Don't feel strongly about moving the repo to the org, but we could.

src/ko/newtoki/src/eu/kanade/tachiyomi/extension/ko/newtoki/DomainNumber.kt
stevenyomi: There's no extension updates involved - the extension version stay unchanged. This build-time number is for first-install users. The in-extension check (and utilizing redirections) is for continuous users.

src/ko/newtoki/src/eu/kanade/tachiyomi/extension/ko/newtoki/FallbackDomainNumber.kt
stevenyomi: Well I managed to make Gradle restore this placeholder file after build, so this technically won't happen unless something bad happens or someone commits code at the exact building time. I can make it more obvious and change the wording. (I wrote this before making it restore)

src/ko/newtoki/build.gradle
```diff
@@ -5,8 +5,44 @@ ext {
...
+def backupFile = new File(domainNumberFileName + "_bak")
+
+task updateDomainNumber {
+    doLast {
+        def domainNumberUrl = "https://stevenyomi.github.io/source-domains/newtoki.txt"
+        def number = new URL(domainNumberUrl).withInputStream { it.readLines()[0] }
+        println("[NewToki] Updating domain number to $number")
+        domainNumberFile.renameTo(backupFile)
+        domainNumberFile.withPrintWriter {
+            it.println("// THIS FILE IS AUTO-GENERATED, DO NOT COMMIT")
```
alessandrojean, line 25: Could maybe add a local `.gitignore` file in the folder to prevent including this file as well?

stevenyomi: This temp file will be reverted after build so this prompt won't be seen. The placeholder file is put there to prevent AS from reporting errors.

PR #16108, opened 2023-04-19
HeanCms: Fix incorrect "Latest Manga" and some thumbnails not loading

seew3l:
Checklist:

Changes:

- Some thumbnails didn't load because they used an absolute url, now check if it's absolute or relative before call the API.
- The request to call the latest manga was wrong because `tags_ids` need to be sent in the payload.
- Remove `series_status` in `allTitlesRequest` because if I add `tags_ids` the request doesn't complete. (If `tags_ids` is added, the total number of manga in the response is higher, idk why)


multisrc/src/main/java/eu/kanade/tachiyomi/multisrc/heancms/HeanCmsDto.kt
```diff
@@ -116,9 +120,9 @@ data class HeanCmsQuerySearchPayloadDto(
     val order: String,
     val page: Int,
     @SerialName("order_by") val orderBy: String,
-    @SerialName("series_status") val status: String,
+    @SerialName("series_status") val status: String? = null,
     @SerialName("series_type") val type: String,
-    @SerialName("tags_ids") val tagIds: List<Int> = emptyList(),
+    @SerialName("tags_ids") val tagIds: List<Int>?,
```
alessandrojean, line 125: Could just let the default as it was before so it will use the default of `emptyList()` when not set, so you don't need to manually set it 3 times.

multisrc/src/main/java/eu/kanade/tachiyomi/multisrc/heancms/HeanCmsDto.kt
```diff
@@ -75,12 +75,16 @@ data class HeanCmsSeriesDto(
...
             .sortedBy(HeanCmsTagDto::name)
             .joinToString { it.name }
-        thumbnail_url = "$apiUrl/$coverPath$thumbnail"
+        thumbnail_url = getThumbnailUrl(apiUrl, coverPath, thumbnail)
         status = this@HeanCmsSeriesDto.status?.toStatus() ?: SManga.UNKNOWN
         url = "/series/${slug.replace(HeanCms.TIMESTAMP_REGEX, "")}"
     }
 }
 
+private fun getThumbnailUrl(apiUrl: String, coverPath: String, thumbnail: String): String {
```
alessandrojean, line 84:
```suggestion
private fun createThumbnailUrl(apiUrl: String, coverPath: String, thumbnail: String): String {
```

multisrc/src/main/java/eu/kanade/tachiyomi/multisrc/heancms/HeanCmsDto.kt
seew3l: if I leave it like this kotlinx.serialization interprets it as optional and doesn't add it to the payload

alessandrojean: Ah, I forgot the option to include the defaults isn't enabled in the `Json` instance injected from the app. No problem, leave it like this for now.

PR #16114, opened 2023-04-20
Undo the HeanCms refactor partially

alessandrojean:
Their API is bad designed so in `allTitlesRequested` if `status` is send with a `null` value or even an empty string, it will throw an error. Reverting this for now as the previous code was handling that already. Thanks to @seew3l for pointing this out.

Checklist:


alessandrojean:
Closing as `explicitNulls = false` is doing the trick of not sending attributes with a `null` value, as intended.

PR #16172, opened 2023-04-25
Toonily: fix search

mobi2002:
Checklist:


multisrc/overrides/madara/toonily/src/Toonily.kt
```diff
@@ -11,6 +19,71 @@ class Toonily : Madara(
     SimpleDateFormat("MMM d, yy", Locale.US),
 ) {
 
+    override val client: OkHttpClient = super.client.newBuilder()
+        .cookieJar(
+            object : CookieJar {
+                override fun loadForRequest(url: HttpUrl): List<Cookie> {
+                    return listOf(
+                        Cookie.Builder()
+                            .domain("toonily.com")
```
stevenyomi, line 28: Extract this to a constant and use this in base URL.

stevenyomi: Or just write a network interceptor to append this cookie instead. Replacing the cookie jar will disconnect the extension's requests from WebView login info.

PR #16252, opened 2023-04-30
Suwayomi (Tachidesk): Implement pagination, sorting, tag-based filtering

Sirver51:
QoL features for the Tachidesk extension. Pagination in particular drastically improves performance; app no longer slows down when fetching a library with hundreds of manga and/or high-res thumbnails.

Checklist:

### DIFF
--- a/src/all/tachidesk/src/eu/kanade/tachiyomi/extension/all/tachidesk/Dto.kt
+++ b/src/all/tachidesk/src/eu/kanade/tachiyomi/extension/all/tachidesk/Dto.kt
@@ -42,7 +42,6 @@ data class MangaDataClass(
     val inLibraryAt: Int = 0,
     val source: SourceDataClass? = null,
     val meta: Map<String, String> = emptyMap(),
-//    val unreadCount: Int? = 0,
     val chapterCount: Int? = 0,
 
     val realUrl: String? = null,
--- a/src/all/tachidesk/src/eu/kanade/tachiyomi/extension/all/tachidesk/Dto.kt
+++ b/src/all/tachidesk/src/eu/kanade/tachiyomi/extension/all/tachidesk/Dto.kt
@@ -39,8 +39,11 @@ data class MangaDataClass(
     val genre: List<String> = emptyList(),
     val status: String = "UNKNOWN",
     val inLibrary: Boolean = false,
+    val inLibraryAt: Int = 0,
     val source: SourceDataClass? = null,
     val meta: Map<String, String> = emptyMap(),
+//    val unreadCount: Int? = 0,
+    val chapterCount: Int? = 0,
 
     val realUrl: String? = null,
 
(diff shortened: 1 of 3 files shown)


Sirver51:
Did most of this a couple of months ago - didn't want to submit the PR until I finished tags, and didn't get around to that until now. Some implementation choices may be a bit odd (see: the tag filtering), mostly because I wanted to avoid code duplication if at all possible.

Also, I see that one of the automated Gradle checks has failed - sorry about that if it's my fault. I'm unsure as to what the issue is or how to fix it, since it works on my machine and I don't understand Gradle at the best of times.

alessandrojean:
The Action failed due to an external issue, I will trigger a rerun.

src/all/tachidesk/src/eu/kanade/tachiyomi/extension/all/tachidesk/Dto.kt
```diff
@@ -39,8 +39,11 @@ data class MangaDataClass(
     val genre: List<String> = emptyList(),
     val status: String = "UNKNOWN",
     val inLibrary: Boolean = false,
+    val inLibraryAt: Int = 0,
     val source: SourceDataClass? = null,
     val meta: Map<String, String> = emptyMap(),
+//    val unreadCount: Int? = 0,
```
alessandrojean, line 45: If you're not using anymore, just remove the line instead of letting it commented.

src/all/tachidesk/src/eu/kanade/tachiyomi/extension/all/tachidesk/Tachidesk.kt
```diff
@@ -123,23 +120,89 @@ class Tachidesk : ConfigurableSource, UnmeteredSource, HttpSource() {
...
-        Filter.CheckBox("Search only current category", false)
+    class ResultsPerPageSelect(options: List<Int>) :
+        Filter.Select<Int>("Results per page", options.toTypedArray())
 
-    override fun getFilterList(): FilterList =
-        FilterList(
-            CategorySelect(refreshCategoryList(baseUrl).let { categoryList }),
-            Filter.Header("Press reset to attempt to fetch categories"),
-            DisableGlobalSearch(),
+    class SortBy(options: List<Pair<String, KProperty1<MangaDataClass, Any?>>>) :
```
alessandrojean, line 154: I believe it's better to you manually map each attribute to the value later instead of using reflection, as it can easily cause issues.

alessandrojean (changes requested):

src/all/tachidesk/src/eu/kanade/tachiyomi/extension/all/tachidesk/Dto.kt
Sirver51: My bad, missed that during cleanup, will fix.

src/all/tachidesk/src/eu/kanade/tachiyomi/extension/all/tachidesk/Tachidesk.kt
Sirver51: Yeah OK, will change. Looking at it now I'm honestly not sure why I decided to do it this way.

alessandrojean (approved):

PR #16508, opened 2023-05-19
MangaTigre: Optimize mangaDetails parsing

bapeey:
Checklist:


src/es/mangatigre/src/eu/kanade/tachiyomi/extension/es/mangatigre/MangaTigre.kt
```diff
@@ -317,7 +322,11 @@ class MangaTigre : HttpSource() {
 
         val result = json.decodeFromString<ChapterDto>(jsonString)
         val slug = result.manga.slug
-        val number = result.number
+        val number = if (result.number % 1 == 0f) {
+            result.number.toInt().toString()
```
arkon, line 326: Could use `toIntOrNull()`

bapeey: I tried but it doesn't work as it should, `result.number` is Float, example `3.0`, and I need to convert it to Int when possible. `toIntOrNull()` returns  `null` instead `3`

PR #16512, opened 2023-05-20
Fixed Bentomanga 

Adam935:
Checklist:

That the third time cloudflare gets triggered by this source will be a result of how the website is coded tachi needs a lot of requests...
So here I am trying to avoid triggering cloudflare, this source might be slower but it will work.


RGFRv2:
Does it work for you?

On my side it still doesn't work, I can access Webview (which was already the case) but not in Tachiyomi itself.

Adam935:
> Does it work for you?
> 
> On my side it still doesn't work, I can access Webview (which was already the case) but not in Tachiyomi itself.

Well it's slower, but it works, maybe changing user agent

PR #16761, opened 2023-06-16
Add an UserAgent on headersBuilder() to bypass Cloudflare error

RGFRv2:
Checklist:

I don't know why, but adding a hard-coded User-Agent to the headersBuilder() function allows the extension to be used again.

For testing : 
All you have to do is open the page in Webview at least once to validate the Cloudflare captcha so you can use the application again.


RGFRv2:
I've added the ability to specify a User-Agent for this extension

And a little description in the popup 

image

image

image

image

PR #16939, opened 2023-06-30
Adding support for IsekaiScan.top

### LINKED ISSUE #15374: New Isekaiscan
### Source name

Isekaiscans

### Source link

http://isekaiscan.top/

### Language

English

### Other details

_No response_

### Acknowledgements

livinstun:
Closes #15374 

Checklist:


multisrc/src/main/java/eu/kanade/tachiyomi/multisrc/madara/Madara.kt
```diff
@@ -715,6 +722,14 @@ abstract class Madara(
         return POST("$mangaUrl/ajax/chapters", xhrHeaders)
     }
 
+    protected open fun newXhrChaptersRequestWithMangaId(mangaId: String): Request {
+        val xhrHeaders = headersBuilder()
+            .add("Referer", "$baseUrl/")
+            .add("X-Requested-With", "XMLHttpRequest")
+            .build()
+        return GET("$baseUrl/ajax-list-chapter?mangaID=$mangaId", xhrHeaders)
```
alessandrojean, line 730: Is there other Madara site using this as well? Looking at the URL it doesn't seem something native to Madara, but a custom endpoint implemented by the site owners. If it's the later, use the endpoint in the override instead of adding to the main theme class.

livinstun: I have removed the logic from the main theme and move to site-specific class

alessandrojean (approved):

multisrc/src/main/java/eu/kanade/tachiyomi/multisrc/madara/Madara.kt
```diff
@@ -836,6 +842,19 @@ abstract class Madara(
                 }
                     .let { dateFormat.tryParse(it.joinToString(" ")) }
             }
+
+            dateWithoutYear -> {
+                return try {
+                    val calendar = Calendar.getInstance()
+                    calendar.time = dateFormat.parse(date) ?: Date()
+                    // set year as current when year is missing in format
```
stevenyomi, line 850: I don't think this is desired. This will probably produce weird date patterns in the chapter list. Might as well just reset all upload dates to zero in the override.

livinstun: I have tried 3 options
1) give the format "MMM dd, HH:mm" in the constructor, since the year is not there it will show the year as 1970
2) don't pass any format so that time will be 0, then it will show chapter date as blank
3) keep the existing logic so that it takes the format without year and adds year as the current year

what do you suggest is the best approach?

stevenyomi: I suggest 2 (or you can manually override `fetchChapterList()` and reset the upload date to make sure it's zero).

If upload date is 0, main Tachi will use fetch date as the upload date while J2K shows it as blank. Either is okay.

Using current year will cause chapters uploaded in previous years (with month value greater than current month) appear as if it's in the future. The date sequence will also be very weird (this year's Jun..Jan, then again this year's Dec..Jan, etc).

livinstun: @stevenyomi 
_"This will probably produce weird date patterns in the chapter list"_
I didn't understand what you mean by this, as we are returning time as an integer, so the output date format will respect the user settings, how will it produce weird date patterns?

stevenyomi: I actually mean "date sequence/order" pattern.

livinstun: i have removed it now

stevenyomi (approved):

livinstun:
do I need to do anything else in this pr to get it to merge?

PR #16947, opened 2023-07-01
comick: improve performance

AwkwardPeak7:
_for real this time_

app takes too long adding manga objects into database, so have reduced the amount sent in one go (50 -> 20)

Checklist:


src/all/comickfun/src/eu/kanade/tachiyomi/extension/all/comickfun/ComickFun.kt
```diff
@@ -47,155 +40,179 @@ abstract class ComickFun(
...
+    private fun paginatedSearchPage(page: Int): MangasPage {
+        val mangas = mutableListOf<SManga>()
+        val endRange = ((page * limit) - 1).let { if (it <= searchResponse.lastIndex) it else searchResponse.lastIndex }
+
+        for (i in (((page - 1) * limit)..endRange)) {
+            mangas.add(
+                searchResponse[i].toSManga(),
+            )
         }
+        return MangasPage(mangas, endRange < searchResponse.lastIndex)
```
stevenyomi, line 127:
```suggestion
        val end = min(page * limit, searchResponse.size)
        val entries = searchResponse.subList((page - 1) * limit, end)
            .map(SearchManga::toSManga)
        return MangasPage(entries, end < searchResponse.size)
```

You'll also need to `import kotlin.math.min`.

stevenyomi (changes requested):

PR #16959, opened 2023-07-01
Fix: page ordering & broken chapter loading

THE-ORONCO:
An update to the comics page broke chapter loading, as the preview chapter did not have a link to use for the chapter overview.

I also saw that the ordering of the pages is wrong (which can manually be fixed by the end users, but that would have to be done for each new chapter).

Checklist:


src/en/aurora/src/eu/kanade/tachiyomi/extension/en/aurora/Aurora.kt
```diff
@@ -145,7 +145,8 @@ class Aurora : HttpSource() {
 
         val chapterOverviewDoc = client.newCall(GET(chapterArchiveUrl, headers)).execute().asJsoup()
         val chapterBlockElements = chapterOverviewDoc.select(".wp-block-image")
```
alessandrojean, line 147: Could just change the selector to `.wp-block-image:has(a)` instead.

THE-ORONCO: good idea!
I forgot that you could do that

Issue #16949, opened 2023-07-01
BacaKomik no pages found


FakeModz:
### Source information

BacaKomik 1.4.4

### Source language

Indonesia

### Steps to reproduce

1. Open any comic
2. Select chapter
3. Its says "no pages found"

### Expected behavior

The comic should be appears and you can read the comic

### Actual behavior

its says "no pages found"

### Tachiyomi version

0.14.6

### Android version

Android 13

### Other details

all comics can't be read because the chapter is "No pages found", it can only be opened using webview

### Acknowledgements

- [X] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open or closed issue.
- [X] I have written a short but informative title.
- [X] I have updated the app to version **[0.14.6](https://github.com/tachiyomiorg/tachiyomi/releases/latest)**.
- [X] I have updated all installed extensions.
- [X] I have tried the [troubleshooting guide](https://tachiyomi.org/help/guides/troubleshooting/).
- [X] If this is an issue with the app itself, I should be opening an issue in the [app repository](https://github.com/tachiyomiorg/tachiyomi/issues/new/choose).
- [X] I will fill out all of the requested information in this form.

PR #16977, opened 2023-07-02
Fix BacaKomik no pages found

friczky:
I aleready tested

Closes : #16949

Checklist:

### DIFF
--- a/src/id/bacakomik/src/eu/kanade/tachiyomi/extension/id/bacakomik/BacaKomik.kt
+++ b/src/id/bacakomik/src/eu/kanade/tachiyomi/extension/id/bacakomik/BacaKomik.kt
@@ -185,11 +185,14 @@ class BacaKomik : ParsedHttpSource() {
     override fun pageListParse(document: Document): List<Page> {
         val pages = mutableListOf<Page>()
         var i = 0
-        document.select("div.arkon-tachiyomi img").forEach { element ->
+        val cdnUrl = "https://ttl.bakul.buzz/" // change with correct CDN url
+        document.getElementsByTag("img").forEach { element ->
             val url = element.attr("onError").substringAfter("src='").substringBefore("';")
-            i++
-            if (url.isNotEmpty()) {
-                pages.add(Page(i, "", url))
+            if (url.startsWith(cdnUrl)) {
+                i++
+                if (url.isNotEmpty()) {
+                    pages.add(Page(i, "", url))
+                }
             }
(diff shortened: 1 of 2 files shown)


src/id/bacakomik/src/eu/kanade/tachiyomi/extension/id/bacakomik/BacaKomik.kt
```diff
@@ -185,7 +185,7 @@ class BacaKomik : ParsedHttpSource() {
     override fun pageListParse(document: Document): List<Page> {
         val pages = mutableListOf<Page>()
         var i = 0
-        document.select("div.imgch-auh img").forEach { element ->
+        document.select("div.arkon-tachiyomi img").forEach { element ->
```
alessandrojean, line 188: I feel like this is done on purpose and will be changed after the merge, isn't it possible to do another selector that doesn't depend on this?

friczky: hmm , I think so.
I will try with another selector

stevenyomi: You can try selecting img elements with attributes containing their CDN URL, which might be slightly more robust. Still breakable though.

friczky: @stevenyomi , sorry I miss understand about this method. 
but I try with this and still no pages found .
`    override fun pageListParse(document: Document): List<Page> {
        val pages = mutableListOf<Page>()
        var i = 0
        val cdnUrl = "https://ttl.bakul.buzz/" 
        document.select("img[src^=$cdnUrl]").forEach { element ->
            val url = element.attr("onError").substringAfter("src='").substringBefore("';")
            i++
            if (url.isNotEmpty()) {
                pages.add(Page(i, "", url))
            }
        }
        return pages
    }`

friczky: I got another selector and work

friczky: thankyou @stevenyomi , I already update with CDN URL @alessandrojean

PR #17004, opened 2023-07-04
YugenMangas: Disable fetchAllTitles

bapeey:
Yugen needs +100 requests for fetch all titles. Also they are not using the timestamp

Maybe we should make it an extension setting.

Checklist:


multisrc/overrides/heancms/yugenmangas/src/YugenMangas.kt
```diff
@@ -15,6 +15,8 @@ class YugenMangas : HeanCms(
     // Site changed from Madara to HeanCms.
     override val versionId = 2
 
+    override val fetchAllTitles = false
```
alessandrojean, line 18: Only Brazilian RS is using at the moment, so should be fine to make the default as `false` and make it opt-in now (by setting it to true on RS only and removing the overrides on the other extensions -- bumping the version for the others isn't needed as technically nothing will change in behavior).

alessandrojean (changes requested):

multisrc/overrides/heancms/yugenmangas/src/YugenMangas.kt
bapeey: Done

Striven100:
Hola, no se q problemas tenga con tachiyomi, pero ahora la pagina yugenmangas.com funciona bien. No se si fue un problema temporal u otra cosa. Porfavor, esta pagina es muy importante para las fuentes de lectores en español

PR #17032, opened 2023-07-06
YugenMangas: Add ratelimit and increment connection timeout

bapeey:
Checklist:


alessandrojean:
Is their site having troubles at the moment? Is it worth also adding a basic rate limit to the source of 1req/sec?

bapeey:
> Is their site having troubles at the moment? Is it worth also adding a basic rate limit to the source of 1req/sec?

Yep you're right. Done

multisrc/overrides/heancms/yugenmangas/src/YugenMangas.kt
```diff
@@ -19,6 +19,7 @@ class YugenMangas : HeanCms(
     override val client = super.client.newBuilder()
         .connectTimeout(60, TimeUnit.SECONDS)
         .readTimeout(90, TimeUnit.SECONDS)
+        .
         .build()
```
alessandrojean, line 23: Seems something is missing

bapeey: xD

Clouddark75:
Their site has its CDN partially dead, probably because of the high concurrent user wave, or in the worst-case scenario, a big data loss, and probably because of that the site is hanging out with the invalid requests.

PR #17181, opened 2023-07-19
update of the japscan extension - the complete reverse of the JS

zormy111:
I completely changed the way to find the keys allowing the decryption of the base64 which contains the information of the images to be recovered. Before there was a recovery by regex, but this was not reliable because the location of these strings changes places and the order is important because it allows to reconstitute the two decryption keys. To succeed in recovering them correctly and in the right order, I debugged the javascript, which allowed me to understand how obfuscation works.

 I understood that the base64 encryption file, the zjs, was not the same file when we had a mobile useragent and a PC useragent, it is however the same key, just that they are placed at different other place
at the end of the file we have the JS which assembles the keys directly as parameters of a function:

a0_0x39cacb('0x13d') + a0_0x39cacb('0x130') + a0_0x39cacb('0x118') + '6N', a0_0x39cacb('0x145') + a0_0x39cacb('0x116') + a0_0x39cacb('0x10f') + 'oe'

we have another function here:

return a0_0x46c1 = function(_0x46c1ab, _0x5a3fac) {
        _0x46c1ab = _0x46c1ab - 0x105;
        let _0x5c76b5 = _0xd80153[_0x46c1ab];
        return _0x5c76b5;
    }

which call string array which contains the key chunks

0 : "laretiLesrap"
1 : "gnp.600/sgmi/"
2 : "atad"
3 : "daol"
4 : "8ceABpX"
..

_0x46c1ab = _0x46c1ab - 0x105;

this operation allows to know the shift in the table, for example
By subtracting 0x105 from '0x13d', we get:

317 - 261 = 56

at location 56, so at index 55 of the array we have: "ohcVEMDTYO7FpBA20zRg"

so here: a0_0x39cacb('0x13d') + a0_0x39cacb('0x130') + a0_0x39cacb('0x118') + '6N'
we replace a0_0x39cacb('0x13d')

by this chain

"ohcVEMDTYO7FpBA20zRg" + a0_0x39cacb('0x130') + a0_0x39cacb('0x118') + '6N'

and that gives us the beginning of the key
and then reverse it
and that gives us the right key

Checklist:


oktapodia (commented):
Well done!

Krylanc3lo:
Thank you very much! I think I start to understand how it works, except maybe here: 
var zjscalc = zjs.split("/[A-Z0-9]/gi,")[1]

you are looking for any letter or digit 1 time ?

zormy111:
image

I cut the javascript with the split function here because I understood that it is just after this piece of string that the two keys are assembled, if I cut all the js from here and I recover the second part, the chain that I recover starts from the collection of the pieces of the keys. Then I just cut by comma, which gives me the first index of the cut the assembly of the first key, and the second index of the array the assembly of the second key

```
var zjscalc = zjs.split("/[A-Z0-9]/gi,")[1] -> the string retrieved in the red square
var calc1 = zjscalc.split(",")[0] -> the string retrieved in the blue square
var calc2 = zjscalc.split(",")[1] -> the string retrieved in the purple square
```


then I cut the assembly of the key a second time to make a table, and then I look for the functions to be replaced by the obfuscated strings

```
var calc1tab = calc1.split("+").toMutableList()
        for (i in 0 until calc1tab.size) {
            print("${calc1tab[i]} ")
            if (calc1tab[i].contains("0x")) {
                var decoupeHexa = calc1tab[i].split("('")[1]
                decoupeHexa = decoupeHexa.split("')")[0]
                var indexkey = Integer.decode(decoupeHexa)
                calc1tab[i] = listKey[indexkey - offsettab - 1]
            }
        }

```

overall I get all the parts of the js that interest me and I reinterpret them

Krylanc3lo:
OK thank you! it is very clear. Thank you for taking the time to explain. I will continue to practice then. I have a different results but it is maybe related to the mobile / desktop version of the js you mentionned.

zormy111:
Yes it's related to that, if you want to do tests on pc with the phone file you have to use a user agent spoof: https://chrome.google.com/webstore/detail/user-agent-switcher-for-c/djflhoibgkdhkhhcedjiklpkjnoahfmg

oktapodia:
@arkon can thus be merged please?

src/fr/japscan/src/eu/kanade/tachiyomi/extension/fr/japscan/Japscan.kt
```diff
@@ -278,32 +275,53 @@ class Japscan : ConfigurableSource, ParsedHttpSource() {
...
         }.attr("src")
         Log.d("japscan", "ZJS at $zjsurl")
-        val zjs = client.newCall(GET(baseUrl + zjsurl, headers)).execute().body.string()
-
-        // Japscan split its lookup table into 1 string of 2 chars + 2 * 3 strings of 20 chars
-        // Try to find the 2 set of 3 strings of 20 chars first
-        val rawLongStringLookupTables = longDecodingStringsRe.findAll(zjs).map {
-            it.groupValues[1]
-        }.toList()
+        var zjs = client.newCall(GET(baseUrl + zjsurl, headers)).execute().body.string()
```
arkon, line 278: Prefer `val` where possible.

src/fr/japscan/src/eu/kanade/tachiyomi/extension/fr/japscan/Japscan.kt
```diff
@@ -278,32 +275,53 @@ class Japscan : ConfigurableSource, ParsedHttpSource() {
...
+        decoupeOffset = "0x" + decoupeOffset.split(";")[0]
+        // on a récupérer l'offset de l'emplacement des clés
+        var offsettab = Integer.decode(decoupeOffset)
+
+        var zjscalc = zjs.split("/[A-Z0-9]/gi,")[1]
+
+        var calc1 = zjscalc.split(",")[0]
+        var calc1tab = calc1.split("+").toMutableList()
+        for (i in 0 until calc1tab.size) {
+            print("${calc1tab[i]} ")
```
arkon, line 295: Remove debug (?) logging.

src/fr/japscan/src/eu/kanade/tachiyomi/extension/fr/japscan/Japscan.kt
```diff
@@ -278,32 +275,53 @@ class Japscan : ConfigurableSource, ParsedHttpSource() {
...
+        var calc2tab = calc2.split("+").toMutableList()
+        for (i in 0 until calc2tab.size) {
+            print("${calc2tab[i]} ")
+            if (calc2tab[i].contains("0x")) {
+                var decoupeHexa = calc2tab[i].split("('")[1]
+                decoupeHexa = decoupeHexa.split("')")[0]
+                var indexkey = Integer.decode(decoupeHexa)
+                calc2tab[i] = listKey[indexkey - offsettab - 1]
+            }
         }
```
arkon, line 314: Since it's repeated logic, consider making a function?

Krylanc3lo:
thanks for the advices and the plugin. I am trying to decrypt the desktop version as well to practice but I think the logic to get the indexes is different. Do you have any idea?

zormy111:
@arkon Thanks for the optimization patch notes!

@Krylanc3lo Yes I know how to do it and I will have to do it... I was lucky during my dev, the script ran and there is the same problem as on pc. The table shifts until it reaches a specific value:

```
while (!![]) {
        try {
            const _0x5c5dfa = parseInt(_0x4db946(a0_0x594ee2._0x3cae83)) / 0x1 + -parseInt(_0x4db946('0x12c')) / 0x2 + parseInt(_0x4db946('0x121')) / 0x3 * (parseInt(_0x4db946('0x122')) / 0x4) + parseInt(_0x4db946('0xfb')) / 0x5 + parseInt(_0x4db946('0x13d')) / 0x6 + -parseInt(_0x4db946(a0_0x594ee2._0x8c6c5c)) / 0x7 * (parseInt(_0x4db946('0x136')) / 0x8) + -parseInt(_0x4db946(a0_0x594ee2._0xfb96e6)) / 0x9;
            if (_0x5c5dfa === _0x1dd53b)
                break;
            else
                _0x17cd93['push'](_0x17cd93['shift']());
        } catch (_0x4d43ab) {
            _0x17cd93['push'](_0x17cd93['shift']());
        }
    }
```

it is necessary that to add in the script this function and to carry out the same rotation until arriving at the good positioning of the table before recovering the keys.

(I think I should have waited for push, I assumed that on the mobile version there was no rotation like on pc. I was just "lucky")

zormy111:
I added the reverse of the missing function to have the array correctly sorted. I know that my modification is good because the script which is totally different for user agent PC works with my modification! I hope it can stay functional for a long time

src/fr/japscan/src/eu/kanade/tachiyomi/extension/fr/japscan/Japscan.kt
```diff
@@ -260,7 +260,28 @@ class Japscan : ConfigurableSource, ParsedHttpSource() {
...
+    }
+
+    private fun listJSToKey(jsList: MutableList<String>, offsettab: Int, listKey: List<String>): MutableList<String> {
+        for (i in 0 until jsList.size) {
+            if (jsList[i].contains("0x")) {
+                var decoupeHexa = jsList[i].split("('")[1]
+                decoupeHexa = decoupeHexa.split("')")[0]
+                var indexkey = Integer.decode(decoupeHexa) - offsettab - 1
+                if (indexkey < 0) {
+                    indexkey = listKey.size - 1
```
zormy111, line 276: It is possible that the value is -1, it actually means that in the JS it is the last value of the array, I have to enter this instruction manually, it is possible that other errors are possible, for example if it is above 1 compared to the size it might correspond to the first value of the array. If that happens I will do a debug and add the modification

src/fr/japscan/src/eu/kanade/tachiyomi/extension/fr/japscan/Japscan.kt
```diff
@@ -260,7 +260,28 @@ class Japscan : ConfigurableSource, ParsedHttpSource() {
             0L
         }
     }
-    
+
+    private fun extractQuotedContent(input: String): List<String> {
```
zormy111, line 264: This function is related to the part of the js which puts the table in order, it is this function which will recover the offsets of the table which are used to see if there are ints in the string

src/fr/japscan/src/eu/kanade/tachiyomi/extension/fr/japscan/Japscan.kt
```diff
@@ -275,47 +296,62 @@ class Japscan : ConfigurableSource, ParsedHttpSource() {
...
-            print("${calc1tab[i]} ")
-            if (calc1tab[i].contains("0x")) {
-                var decoupeHexa = calc1tab[i].split("('")[1]
-                decoupeHexa = decoupeHexa.split("')")[0]
-                var indexkey = Integer.decode(decoupeHexa)
-                calc1tab[i] = listKey[indexkey - offsettab - 1]
-            }
+        val listKeyOrder = extractQuotedContent(decoupeFuncOrder).toMutableList()
+
+        if (listKeyOrder.size < 3) {
```
zormy111, line 316: it is not normal to have less than 3 elements which dynamically check the values ​​containing int in the array, there are some which are directly hard in the calculation but there are necessarily more than 2

src/fr/japscan/src/eu/kanade/tachiyomi/extension/fr/japscan/Japscan.kt
```diff
@@ -275,47 +296,62 @@ class Japscan : ConfigurableSource, ParsedHttpSource() {
...
-        var calc2tab = calc2.split("+").toMutableList()
-        for (i in 0 until calc2tab.size) {
-            print("${calc2tab[i]} ")
-            if (calc2tab[i].contains("0x")) {
-                var decoupeHexa = calc2tab[i].split("('")[1]
-                decoupeHexa = decoupeHexa.split("')")[0]
-                var indexkey = Integer.decode(decoupeHexa)
-                calc2tab[i] = listKey[indexkey - offsettab - 1]
+        var goodorder = false
+        for (i in 0 until listKey.size) {
```
zormy111, line 321: This loop sorts the array until it reaches the condition:

If all previously retrieved offsets contain numbers, then the sort is good

src/fr/japscan/src/eu/kanade/tachiyomi/extension/fr/japscan/Japscan.kt
```diff
@@ -275,47 +296,62 @@ class Japscan : ConfigurableSource, ParsedHttpSource() {
...
+                    goodorder = false
+                    break
+                }
             }
+
+            if (goodorder) {
+                break
+            }
+
+            val firstElement = listKey.removeAt(0)
```
zormy111, line 335: in the JavaScript algorithm, as long as we are not in the correct sort, all the elements switch backwards, this is what I reproduce with this instruction

Krylanc3lo:
oh nice, it is very smart, I will have a look and play with it then :)
Thanks again

src/fr/japscan/src/eu/kanade/tachiyomi/extension/fr/japscan/Japscan.kt
Krylanc3lo: what should be the ouput of this section ? when I try to extract the string after =parseInt( and until ";", I do not find any string within quotes.

zormy111: there is indeed a problem on the pc script (so possibly also on the mobile version when it will reload the script), I have to add an option in the event that this "-" exists, there will surely be other conditions like this you have to do your best to add them all, thanks for the help!

zormy111: https://github.com/tachiyomiorg/tachiyomi-extensions/commit/52d9c53316e92d796ad7bfa4a566d4c53eb26464

I added it, it's simple but it's enough not to make it work, thanks again for your help

src/fr/japscan/src/eu/kanade/tachiyomi/extension/fr/japscan/Japscan.kt
arkon: You should leave all of these various comments as actual comments in the code.

zormy111:
@arkon I added comments, the script is currently stable and working, it can be merged whenever you want

oktapodia:
@zormy111 hey I don't really think this is related but since updating it, I cannot bypass cloudflare even in webview T_T

Krylanc3lo:
I think it is unrelated, JS did activate CF on the chapter pages

hutigh:
> @zormy111 hey I don't really think this is related but since updating it, I cannot bypass cloudflare even in webview T_T

Same issue here it won't bypass cf since the update I never had this issue in previous updates :(

Issue #17170, opened 2023-07-18
Manhuaren: failed to load listing and chapter pages


Erbzur:
### Source information

Manhuaren 1.4.12

### Source language

Chinese

### Steps to reproduce

1. open any chapter of a manga, such as chainsaw man

### Expected behavior

Load manga pictures successfully

### Actual behavior

Indicated unable to load the picture

### Tachiyomi version

0.14.6

### Android version

Android 12

### Acknowledgements

- [X] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open or closed issue.
- [X] I have written a short but informative title.
- [X] I have updated the app to version **[0.14.6](https://github.com/tachiyomiorg/tachiyomi/releases/latest)**.
- [X] I have updated all installed extensions.
- [X] I have tried the [troubleshooting guide](https://tachiyomi.org/help/guides/troubleshooting/).
- [X] If this is an issue with the app itself, I should be opening an issue in the [app repository](https://github.com/tachiyomiorg/tachiyomi/issues/new/choose).
- [X] I will fill out all of the requested information in this form.

stevenyomi:
@felixfon Time to remove extension I guess?

Erbzur:
> I can't reproduce it on my device. @Erbzur Can you provide more info?

{
  "errorResponse": {
    "code": 40011,
    "subCode": "",
    "message": "缺少token",
    "rgtk": null
  }
}
these are what I found when fetching the api

lok77chan:
> > I can't reproduce it on my device. @Erbzur Can you provide more info?
> 
> { "errorResponse": { "code": 40011, "subCode": "", "message": "缺少token", "rgtk": null } } these are what I found when fetching the api

I am having same issue. This has been the case for me for all the recent patch. The message in this json mean "missing token" just in case.
![Screenshot_20230719-110835](https://github.com/tachiyomiorg/tachiyomi-extensions/assets/15722025/04bf88d4-e156-48c2-b2cc-908ff57ee70b)

>

LAGMAN-TW:
I was able to make it work with VPN. Although I can see the images, the WebView showed the same error message for me

john-stones5:
I've never had this problem but the old "no value for response" thing appears again just now.

stevenyomi:
Closing due the fix in #17202. Open a new issue if it happens again.

Erbzur:
[...]

PR #17202, opened 2023-07-21
Fix Manhuaren unable to get token

felixfon:
I have corrected some incorrect values from #17134, but I'm unsure if #17170 is a result of this.
It seems that Manhuaren has blocked the Pixel 6 device model.
Should we consider adding fields on the configuration page to allow users to modify the Model, User-Agent, and other parameters?

Checklist:


stevenyomi:
> It seems that Manhuaren has blocked the Pixel 6 device model.

I doubt if it's because of this extension. If it is, we should remove it (and maybe you can maintain a closed-source version in your repo).

> Should we consider adding fields on the configuration page to allow users to modify the Model, User-Agent, and other parameters?

You can, but it's harder to get a set of consistent parameters than just a single user agent string.

PR #17256, opened 2023-07-25
Pixiv: new filters, group-by-series support

### LINKED ISSUE #16035: Pixiv: manga chapters are not grouped into series.
### Source information

Pixiv 1.4.3

### Source language

English

### Steps to reproduce

1. Search for any manga with multiple chapters.

### Expected behavior

One entry for the manga series should be shown, with chapters listed inside.

### Actual behavior

Each chapter shows as its own entry  with a single 'oneshot' chapter inside.

### Tachiyomi version

0.14.5

### Android version

13

### Other details

_No response_

### Acknowledgements

- [X] I have updated the app to version **[0.14.5](https://github.com/tachiyomiorg/tachiyomi/releases/latest)**.

Solitai7e:
Checklist:

Closes #16035


alessandrojean:
There's a lot of Flow/suspend/coroutines stuff being handled that seems like overkill, an extension doesn't need that at all. Should be able to use the overridable methods and do any extra requests needed on the parse methods.

src/all/pixiv/src/eu/kanade/tachiyomi/extension/all/pixiv/PixivFactory.kt
```diff
@@ -1,9 +0,0 @@
-package eu.kanade.tachiyomi.extension.all.pixiv
```
alessandrojean, line 1: Keep it in a separate file, please.

alessandrojean (changes requested):

src/all/pixiv/src/eu/kanade/tachiyomi/extension/all/pixiv/Pixiv.kt
```diff
@@ -1,216 +1,413 @@
...
-
-        headers = headersBuilder().add("Accept", "application/json").build(),
-        method = method,
-    )
-
-    private inline fun <reified T> apiResponseParse(response: Response): T {
-        if (!response.isSuccessful) {
-            throw Exception(response.message)
+
+    private val pixivLang = if (lang == "all") "ja" else lang
```
AntsyLich, line 42: Get rid of all lang then

AntsyLich (changes requested):

src/all/pixiv/src/eu/kanade/tachiyomi/extension/all/pixiv/Pixiv.kt
```diff
@@ -1,216 +1,413 @@
...
-    private fun illustUrlToId(url: String): String =
-        url.substringAfterLast("/")
+    private fun Flow<PixivIllust>.toSManga() = flow {
+        val seriesIdsSeen = mutableSetOf<String>()
+
+        collect { illust ->
+            val series = illust.series
+
+            if (series == null) {
+                val manga = SManga.create()
```
vetleledaal, line 77:
```kt
SManga.create().apply {
    setUrlWithoutDomain(...)
    title = "..."
    thumbnail_url = "..."
}
```

src/all/pixiv/src/eu/kanade/tachiyomi/extension/all/pixiv/Pixiv.kt
```diff
@@ -1,216 +1,413 @@
...
+                        }
+                    }
+
+                    predicates = buildList {
+                        filters.tags.toPredicate()?.let(::add)
+                        filters.rating.toPredicate()?.let(::add)
+                    }
+                } else {
+                    illustsFlow = makeIllustSearchFlow(
+                        word = filters.tags.state.ifBlank { "漫画" },
```
vetleledaal, line 196: if this is a multi-lang extension, consider using English or different text for each language

Solitai7e: That's actually just a fallback query string (the API doesn't seem to allow empty queries) for posts with the "manga" tag and I'm fairly certain the different language forms are all aliased to the same tag.

src/all/pixiv/src/eu/kanade/tachiyomi/extension/all/pixiv/Pixiv.kt
```diff
@@ -1,216 +1,403 @@
...
-    )
-
-    private inline fun <reified T> apiResponseParse(response: Response): T {
-        if (!response.isSuccessful) {
-            throw Exception(response.message)
+
+    override fun headersBuilder(): Headers.Builder =
+        super.headersBuilder().add("Referer", "$baseUrl/")
+
+    private open inner class HttpCall(href: String?) {
```
alessandrojean, line 31: I don't think there's a need for this class and the `ApiCall` one. If you need to always have the Accept header in the API calls, you can have an inner val like `apiHeaders` and always use it with the `GET`/`POST` methods provided by the ext-lib instead.

```kotlin
private fun apiHeadersBuilder(): Headers.Builder = super.headersBuilder()
    .add("Accept", "application/json")

private val apiHeaders by lazy { apiHeadersBuilder().build() }
```

src/all/pixiv/src/eu/kanade/tachiyomi/extension/all/pixiv/Pixiv.kt
```diff
@@ -1,216 +1,403 @@
...
 
-    private fun urlEncode(string: String): String =
-        URLEncoder.encode(string, "UTF-8").replace("+", "%20")
+                for (p in countUp(start = 1)) {
+                    call.url.setEncodedQueryParameter("page", p.toString())
 
-    private fun parseTimestamp(string: String) =
-        runCatching { dateFormat.parse(string)?.time!! }.getOrDefault(0)
+                    val entries = call.executeApi<PixivRankings>().ranking!!
+                    if (entries.isEmpty()) break
```
alessandrojean, line 67: Why instead of having to deal with a custom sequence, just do a normal while...?

Also, would be easier to actually use the parse (like `popularMangaParse`) methods instead of overriding the fetch and having to deal with Observables (which will be removed in the future).

A better way of dealing with this is making the first request with the normal popularMangaRequest(), and inside `popularMangaParse()` do the parsing of the first page by consuming the response and making any additional requests inside the parse method. If the source doesn't have pagination, can add a variable inside the class to act like a cache and then return sublists in each page.

Use sources like MangaDex as an code example, they do consume an API and are better structured to do so.

src/all/pixiv/src/eu/kanade/tachiyomi/extension/all/pixiv/Pixiv.kt
```diff
@@ -1,216 +1,403 @@
...
-                    .map { illustParse(it) }
-                    .doOnNext { illustCache.put(url, it) }
-            },
-        )
+    private val searchLock = object {}
+    private var searchNextPage = 1
+    private var searchHash: Int? = null
+    private lateinit var searchFlow: Iterator<SManga>
+
+    override fun fetchSearchManga(page: Int, query: String, filters: FilterList): Observable<MangasPage> {
```
alessandrojean, line 92: Same here in this method, the whole synchronized/coroutines/sequences things is making the code a little unreadable and very hard to understand. It's better to use the methods provided by the super class instead.

src/all/pixiv/src/eu/kanade/tachiyomi/extension/all/pixiv/Pixiv.kt
```diff
@@ -1,216 +1,403 @@
...
+                yield(manga)
             }
         }
+    }
+
+    private val latestMangaLock = object {}
+    private var latestMangaNextPage = 1
+    private lateinit var latestMangaIterator: Iterator<SManga>
+
+    override fun fetchLatestUpdates(page: Int): Observable<MangasPage> = synchronized(latestMangaLock) {
```
alessandrojean, line 245: Same here.

src/all/pixiv/src/eu/kanade/tachiyomi/extension/all/pixiv/Pixiv.kt
```diff
@@ -1,216 +1,403 @@
...
-                is FilterType -> parameters["type"] = filter.value
-                is FilterRating -> parameters["mode"] = filter.value
-                is FilterSearchMode -> parameters["s_mode"] = filter.value
-                is FilterOrder -> parameters["order"] = filter.value
-                is FilterDateBefore -> filter.value?.let { parameters["ecd"] = it }
-                is FilterDateAfter -> filter.value?.let { parameters["scd"] = it }
-                else -> {}
+    override fun getFilterList() = FilterList(PixivFilters())
+
+    private fun Sequence<PixivIllust>.toSManga() = sequence {
```
alessandrojean, line 219: I don't get it why there's a need of using sequences.

src/all/pixiv/src/eu/kanade/tachiyomi/extension/all/pixiv/Pixiv.kt
```diff
@@ -1,216 +1,403 @@
...
-        listOf(
-            FilterType(),
-            FilterRating(),
-            FilterSearchMode(),
-            FilterOrder(),
-            FilterDateBefore(),
-            FilterDateAfter(),
-        ),
-    )
+    override fun popularMangaRequest(page: Int): Request =
```
alessandrojean, line 397: Consider using them as it will make the code more easier to read and understand.

src/all/pixiv/src/eu/kanade/tachiyomi/extension/all/pixiv/Pixiv.kt
```diff
@@ -1,216 +1,403 @@
...
+        throw UnsupportedOperationException("Not used.")
+
+    override fun searchMangaParse(response: Response): MangasPage =
```
alessandrojean, line 400: Same here.

alessandrojean (changes requested):
A general comment as well: using custom classes for a lot of operations doesn't seem to be helping much, and is instead making the code bloated. Also a lot of advanced stuff like threads, mutexes and locks aren't needed.

Solitai7e:
Thing is there's now this client-side filtering step for the search/popular/latest functions to deal with stuff like removing duplicate series (see toSManga) and making (most of) the filters work when search results are being fetched from an API that doesn't support one or more filters. For example right now if you do a search with an empty query and the users filter instead of a regular search it actually does a user search followed by a query for the posts of each of the users, which requires me to keep track of the progress of both searches on top of a list of series IDs for removing duplicates and I don't really know of a good way to do all that that doesn't involve flows/sequences, closures, or god forbid even more class-scope variables (though I now realize it might be a better idea to bundle them together with the iterators)

The synchronization blocks are there because I thought maybe Tachi assumes the class is thread-safe and calls from multiple threads, if that's not the case then I'll get rid of them.

alessandrojean:
> The synchronization blocks are there because I thought maybe Tachi assumes the class is thread-safe and calls from multiple threads

You can assume the methods will be called already in a separate and unique thread for that, we didn't have any issues with these things before in other extensions, so should be safe.

> that doesn't support one or more filters

Better to addopt the strategy other extensions does of warning the user through a Title filter that only one filter can be used at once, and if the query is not empty the filters will be ignored. That will make the implementation easier and the users are already familiar with this approach.

> removing duplicates

You can keep a private variable inside the class to keep track of the IDs of the series already displayed, just a simple string/int list will do the job. If the app call with the page = 1, you can clear it and start again.

Solitai7e:
I can't access the Discord since only people who are SMS-verified are allowed to post. (Discord doesn't like my number for some reason)

If you're OK with doing this in DMs please send a friend request to `solitai7e`.

Solitai7e:
(I wish github had private messages)

Re:
> image

It still wouldn't let me click on submit on the required steps screen.

Solitai7e:
I fail to comprehend how this is more readable or understandable than simply using a sequence.

```kotlin
private val userSearchCall = HttpCall("/search_user.php?s_mode=s_usr")
private val userIllustsCall = ApiCall("/touch/ajax/user/illusts")

private var searchUserIllustsNextPage = 1
private var userSearchNextPage = 1
private var userIllustsNextPage = 1

private lateinit var userSearchUnprocessedResults: Iterator<String>

private fun searchUserIllustsNextUser(): Boolean {
    userSearchCall.url.addEncodedQueryParameter("p", userSearchNextPage++.toString())

    userSearchUnprocessedResults = Jsoup.parse(userSearchCall.execute().body.string())
        .select(".user-recommendation-item > a").eachAttr("href")
        .map { it.substringAfterLast('/') }
        .iterator()

    if (!userSearchUnprocessedResults.hasNext()) return false
    
    userIllustsNextPage = 1
    userIllustsCall.url.setEncodedQueryParameter("id", userSearchUnprocessedResults.next())
    
    return true
}

private fun searchUserIllusts(page: Int, nick: String, type: String?): List<PixivIllust>? {
    if (page == 1) {
        searchUserIllustsNextPage = 1
        userSearchNextPage = 1
        searchUserIllustsNextUser()
    } else {
        require(page == searchUserIllustsNextPage)
    }

    searchUserIllustsNextPage++
    userIllustsCall.url.setEncodedQueryParameter("id", userSearchUnprocessedResults.next())

    val illusts = userIllustsCall.executeApi<PixivResults>().illusts!!
    if (illusts.isEmpty() && !searchUserIllustsNextUser()) return null

    return illusts
}
```

alessandrojean:
I guess it can be kept like it is for now, and it will be open to a future rewrite instead when needed. I still think it's overcomplicated code, but if it works, it works.

alessandrojean:
@Solitai7e Can you make a new PR to change the LruCache instance to be created `by lazy`? It direct usage is making the extension inspector to fail and not release the new versions of the extensions.

alessandrojean:
> Still can't access the Discord by the way, is there something special I'm supposed to do for the bypass to work?

Doesn't think so, it should be already working.

PR #17269, opened 2023-07-26
Add a Status search filter and remove 1 deleted sort in Comick

iD-666:
Checklist:


src/all/comickfun/src/eu/kanade/tachiyomi/extension/all/comickfun/ComickFunFilters.kt
```diff
@@ -5,13 +5,14 @@ import eu.kanade.tachiyomi.source.model.FilterList
 
 fun getFilters(): FilterList {
     return FilterList(
-        Filter.Header(name = "NOTE: Everything below is ignored if using text search"),
```
iD-666, line 8: There is no text wrapping and the app only showed:
> NOTE: Everything below is ignored if using

PR #17279, opened 2023-07-27
Pixiv: LruCache fix

Solitai7e:
Checklist:

Addresses _https://github.com/tachiyomiorg/tachiyomi-extensions/pull/17256#issuecomment-1653687771_

Noticed in testing that LruCache::get seemed to be getting called concurrently and was causing redundant requests to be made which the caching is meant to eliminate in the first place. This is now fixed with synchronization but I now wonder if it's really fine to not synchronize the other stuff.


src/all/pixiv/build.gradle
```diff
@@ -6,7 +6,7 @@ ext {
     extName = 'Pixiv'
     pkgNameSuffix = 'all.pixiv'
     extClass = '.PixivFactory'
-    extVersionCode = 4
+    extVersionCode = 5
```
vetleledaal, line 9: probably not needed since CI is failing

PR #17318, opened 2023-07-30
YugenMangas: Fix not showing new chapters

bapeey:
~Users will have to migrate their series~

The extension was making slugs permanent all this time, it worked fine but now the permanent url doesn't show the new chapters; now only makes them permanent if fetchAllTitles is enabled.

Also, Yugen now uses timestamps in urls, so I added an option to enable fetchAllTitles. Users can activate the option or just migrate.

Checklist:


multisrc/overrides/heancms/yugenmangas/src/YugenMangas.kt
```diff
@@ -29,6 +43,29 @@ class YugenMangas : HeanCms(
         timeZone = TimeZone.getTimeZone("UTC")
     }
 
+    private fun getFetchAllSeriesPref(): Boolean {
+        return preferences.getBoolean(FETCH_ALL_SERIES_PREF, FETCH_ALL_SERIES_DEFAULT_VALUE)
+    }
+
+    override fun setupPreferenceScreen(screen: PreferenceScreen) {
+        val fetchAllSeriesPreference = androidx.preference.SwitchPreferenceCompat(screen.context).apply {
```
alessandrojean, line 51: Import it?

```suggestion
        val fetchAllSeriesPreference = SwitchPreferenceCompat(screen.context).apply {
```

multisrc/overrides/heancms/yugenmangas/src/YugenMangas.kt
```diff
@@ -29,6 +43,29 @@ class YugenMangas : HeanCms(
...
+            summary = FETCH_ALL_SERIES_SUMMARY
+            setDefaultValue(FETCH_ALL_SERIES_DEFAULT_VALUE)
+
+            setOnPreferenceChangeListener { _, newValue ->
+                val checkValue = newValue as Boolean
+
+                preferences.edit()
+                    .putBoolean(FETCH_ALL_SERIES_PREF, checkValue)
+                    .commit()
+            }
```
alessandrojean, line 63: Can remove this, the app already handles the changes and commit it.

```suggestion
```

multisrc/overrides/heancms/yugenmangas/src/YugenMangas.kt
```diff
@@ -79,4 +116,14 @@ class YugenMangas : HeanCms(
...
         Genre("Yuri", 44),
     )
+
+    companion object {
+        private const val FETCH_ALL_SERIES_PREF = "fetchAllSeriesPref"
+        private const val FETCH_ALL_SERIES_TITLE = "Buscar todas las series"
+        private const val FETCH_ALL_SERIES_SUMMARY = "Busca las URLs actuales de las series. " +
+            "Habilitar esta opción evita la necesidad de migrar, " +
+            "pero puede llevar tiempo dependiendo del número total de series. \n" +
+            "Tendrá que migrar cada vez que cambie esta opción."
```
alessandrojean, line 126: Good to mention the app needs to be restarted after the change.

alessandrojean (changes requested):

AwkwardPeak7:
Saving temporary urls in database will end up messing with library badge and users like to complain about that a lot...

Also, I was wondering if we can change the temporary slug logic, calling the front-end with the permanent url redirects to temporary url, so could technically use this behaviour instead of fetching all titles.

alessandrojean:
> calling the front-end with the permanent url redirects to temporary url

Yeah, but that will mean +1 additional request in the mangaDetails and chapterList methods, would be bad during library global update. Fetching one time only would be better (as it's currently done).

alessandrojean (approved):

bapeey:
> Also, I was wondering if we can change the temporary slug logic, calling the front-end with the permanent url redirects to temporary url, so could technically use this behaviour instead of fetching all titles.

I tried it a while ago, but some series redirect to a different one. Also, right now the permanent url no longer redirects, instead it shows a version without the updated chapters.

AwkwardPeak7:
> +1 additional request

Yeah, but for sources like yugen, fetching all titles already cause too many requests (iirc).

> some series redirect to a different one. Also, right now the permanent url no longer redirects, instead it shows a version without the updated chapters

Ah wasn't aware of this.

alessandrojean:
> fetching all titles already cause too many requests

Unfortunately, but at least it's done only at the app start one time only if the setting is enabled. Also by what bapeey said, we can't rely on redirects no more.

PR #17336, opened 2023-07-31
Add a new `lib-i18n` to make message translation easier

beerpiss:
Continuation of #16942, except the translation files are stored under `assets/i18n` instead of `res/raw`, as stuff under `assets/` are [addressable by name](https://developer.android.com/guide/topics/resources/providing-resources#OriginalFiles), and should work even on release configuration.

Below here lies the original PR description.

----

Adds a new `lib-i18n` to load message `.properties` files from the `res/raw` folder and provide an easier support for string localization in the sources. It uses Java's `PropertyResourceBundle` under the hoods.

Using Properties files allows users with low programming knowledge to be able to translate the files as well, and it also has a good support in Android Studio with the official JetBrain's [Resource Bundle Editor plugin](https://plugins.jetbrains.com/plugin/17035-resource-bundle-editor), which allow translating in a nice UI with all languages at once, besides having autocomplete when using `intl["key"]`.

For now it can only be used in individual sources as multisrc ones doesn't have a good support for additional libraries through the `additional.gradle` file.

MangaDex and M+ are included initially as example.

Checklist:


alessandrojean:
Hoping it doesn't break the apps again :)

PR #17556, opened 2023-08-15
Kemono: Add option to load low resolution image in settings

HisuianZoroark69:
Checklist:


multisrc/src/main/java/eu/kanade/tachiyomi/multisrc/kemono/Kemono.kt
```diff
@@ -211,6 +212,13 @@ open class Kemono(
             entryValues = mirrorUrls
             setDefaultValue(defaultUrl)
         }.let(screen::addPreference)
+
+        SwitchPreferenceCompat(screen.context).apply {
+            key = USE_LOW_RES_IMG
+            title = "Enable low resolution image"
+            summary = "Reduce load time significantly\nNeeds to clear chapter cache to take full effect."
```
alessandrojean, line 219:
```suggestion
            summary = "Reduce load time significantly. You need to clear the chapter cache to take effect."
```

stevenyomi (changes requested):
You can override `imageRequest()` and insert the `/thumbnail` part to the network request to make it take effect immediately.

HisuianZoroark69:
> You can override `imageRequest()` and insert the `/thumbnail` part to the network request to make it take effect immediately.

I think the cached image is linked to the page itself, not the url. So adding the ```/thumbnail``` in neither the ```pageListParse()``` nor the ```imageRequest()``` will effect immediately without clearing the chapter cache.
The best solution right now would be clearing the chapter cache automatically in code or prompt the user to change it like in your ```MangaReader``` extension: https://github.com/tachiyomiorg/tachiyomi-extensions/blob/ef58b57c6547516045c4c7ca1d6f7a1718d78f35/multisrc/overrides/mangareader/mangareaderto/src/Preferences.kt#L9-L18

But apparently I have yet to find a way to clear the chapter cache by code 😹.

stevenyomi:
What I mean is: if you open a chapter but images fail to load, then you turn on showing thumbnails, you don't need to clear the chapter cache (cached URLs) and the extension will request thumbnails upon opening the chapter again.

With your current solution, you'd always need to clear the chapter cache for opened chapters to reload.

Indeed, if you have already loaded thumbnails and then turn off this setting to view original images, you have to clear chapter cache. This is required in any solution.

I suggest you implement my suggestion and reword the summary to "When turning off, clear chapter cache to remove cached low resolution images."

Regarding MangaReader, it's because they don't have a predictable URL pattern to modify in `imageRequest()` (or I ddin't figure out), so I had to use the provided image quality preference when fetching image URLs.

HisuianZoroark69:
I have implemented your suggestion by adding the ```/thumbnail``` at ```imageRequest()``` instead of ```pageListParse()``` and indeed it worked like what you've said.😸

I'll also reword the summary to "Reduce load time significantly. When turning off, clear chapter cache to remove cached low resolution images." as you've suggested.

Thank you for your help.

multisrc/src/main/java/eu/kanade/tachiyomi/multisrc/kemono/Kemono.kt
```diff
@@ -182,9 +183,11 @@ open class Kemono(
 
     override fun pageListParse(response: Response): List<Page> {
         val post: List<KemonoPostDto> = response.parseAs()
-        return post[0].images.mapIndexed { i, path -> Page(i, imageUrl = baseUrl + path) }
+        return post[0].images.mapIndexed { i, path -> Page(i, imageUrl = path) }
     }
 
+    override fun imageRequest(page: Page) = GET(baseUrl + (if (preferences.getBoolean(USE_LOW_RES_IMG, false)) "/thumbnail" else "") + page.imageUrl!!, headers)
```
stevenyomi, line 189: Keep the original image URL to avoid breaking exisiting cached chapters. Instead, split the URL and insert the `/thumbnail` part in `imageRequest()`.

stevenyomi (changes requested):

HisuianZoroark69:
Is this ok?

multisrc/src/main/java/eu/kanade/tachiyomi/multisrc/kemono/Kemono.kt
```diff
@@ -185,6 +186,14 @@ open class Kemono(
         return post[0].images.mapIndexed { i, path -> Page(i, imageUrl = baseUrl + path) }
     }
 
+    override fun imageRequest(page: Page): Request {
+        var url: StringBuilder = StringBuilder(page.imageUrl)
+        if (preferences.getBoolean(USE_LOW_RES_IMG, false)) {
+            url.insert(baseUrl.length, "/thumbnail")
+        }
+        return GET(url.toString(), headers)
```
stevenyomi, line 194:
```suggestion
        val imageUrl = page.imageUrl!!
        if (!preferences.getBoolean(USE_LOW_RES_IMG, false)) return GET(imageUrl, headers)
        val index = imageUrl.indexOf('/', startIndex = 8) // https://
        val url = buildString { 
            append(imageUrl, 0, index)
            append("/thumbnail")
            append(imageUrl, index, imageUrl.length)
        }
        return GET(url, headers)
```

Use `imageUrl.indexOf('/', startIndex = 8)` since `baseUrl` might change (by preference, etc).

You can use the insert method; early return and more `append` lines are just optimizations.

stevenyomi (changes requested):

multisrc/src/main/java/eu/kanade/tachiyomi/multisrc/kemono/Kemono.kt
```diff
@@ -225,5 +245,6 @@ open class Kemono(
         private fun List<SManga>.filterUnsupported() = filterNot { it.author == "Discord" }
 
         private const val BASE_URL_PREF = "BASE_URL"
+        private var USE_LOW_RES_IMG = "USE_LOW_RES_IMG"
```
stevenyomi, line 248:
```suggestion
        private const val USE_LOW_RES_IMG = "USE_LOW_RES_IMG"
```

stevenyomi (changes requested):

alessandrojean (approved):

Issue #17478, opened 2023-08-11
Can't read anything Japscan


Mousssous:
### Source information

Japscan 1.4.43

### Source language

"French"

### Steps to reproduce

1. Click on Japscan source or make a global search
2. Select a manga/webtoon/manhwa (no matter which one)
3. Click on a chapter of the selected manga

### Expected behavior

The manga's chapter images should appear

### Actual behavior

Error : "Attendait 2 chaines de recherche dans ZJS, en a trouvé 88"
And brought me back to chapter selection

### Tachiyomi version

"0.14.6"

### Android version

"Android 10"

### Other details

This type of error already happened several times and has already been fixed, apparently it's a lookup tables problem. The number of "zjs" found may change depending on the manga selected. 

### Acknowledgements

- [X] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open or closed issue.
- [X] I have written a short but informative title.
- [X] I have updated the app to version **[0.14.6](https://github.com/tachiyomiorg/tachiyomi/releases/latest)**.
- [X] I have updated all installed extensions.
- [X] I have tried the [troubleshooting guide](https://tachiyomi.org/help/guides/troubleshooting/).
- [X] If this is an issue with the app itself, I should be opening an issue in the [app repository](https://github.com/tachiyomiorg/tachiyomi/issues/new/choose).
- [X] I will fill out all of the requested information in this form.

KirinRaikage:
Okay, now we're starting to hit a rock where Japscan, despite all our efforts, is fixed, broken, fixed, broken again.
We're playing a really annoying game where Japscan keeps breaking the extension on purpose. 
I would personally remove the extension.
WDYT?

aboudiom59:
> Okay, now we're starting to hit a rock where Japscan, despite all our efforts, is fixed, broken, fixed, broken again. We're playing a really annoying game where Japscan keeps breaking the extension on purpose. I would personally remove the extension. WDYT?

Well Japscan is the french source with the most content, like by far, there is a lot of content there that isn't anywhere else, sushi-scan is a better site but it's not even touching the sheer content there. But honestly if they're just gonna purposefully break the extension for traffic they're not gonna get anyway maybe it's just better to give up and leave it alone. It'd be a shame but unless the site actually comes to some sort of agreement maybe it's not worth the effort.

PR #17543, opened 2023-08-14
Japscan

Rajh:
Close #17478

Added a preference to set custom keys (advanced users can extract keys and use them directly without having to wait for a new version)

The zjs now contains several matching keys. This version try every combinaisons.

Saved working keys in class attributes as a "cache".

Checklist:

- [X] Updated `extVersionCode` value in `build.gradle` for individual extensions
- [X] Updated `overrideVersionCode` or `baseVersionCode` as needed for all multisrc extensions
- [X] Referenced all related issues in the PR body (e.g. "Closes #xyz")
- [X] Added the `isNsfw = true` flag in `build.gradle` when appropriate
- [X] Have not changed source names
- [X] Have explicitly kept the `id` if a source's name or language were changed
- [X] Have tested the modifications by compiling and running the extension through Android Studio

PR #17585, opened 2023-08-18
Update Japscan

zormy111:
Japscan has again changed its obfuscation, but it is the most complete source of French scans, so don't give up. It is now impossible to recover the two encryption keys only with the reading of the JS file. But if we execute the javascript we can quite easily recover the two decryption keys thanks to the debugger, so with a single instance of a browser it is possible to recover the keys and share them with everyone!

This is what I managed to set up: https://gist.github.com/zormy111/f6abdc7f9385e95203e2b6a64af15ea3/revisions

The JS script changes its key and the name of its variables at each time change, it's a fixed time, thanks to a nodejs script I get the two keys and I modify my gist to share them, they are global keys they can therefore be used by anyone.

What it changes: There is a server dependency, however I think it is better to choose this system because it will be much more stable than a script that tries to read the key in the obfuscated JS.

**Do I have to share my node.js script?**

I could make it public, however, I think I understood that the japscan site was closely monitoring the current extension and as soon as we find a solution they quickly patch the code in a simple way, if I keep it on the server side ( and I remind you that it concerns ONLY the decryption keys and nothing else) they will have a lot more trouble patching or even not being able to patch it if they keep the same system.

For me it works very well, the script recovers the key without worries and adapts as it should to each change of script, I undertake to maintain it as best as possible (I am a very regular user of the application). If it happens that I no longer maintain for some reason I will share my script.

I also specify that only mobile user agents work because the script changes depending on whether it is a pc user agent and a mobile user agent. (My script retrieves the one from the mobile)

Checklist:


zormy111:
Closes https://github.com/tachiyomiorg/tachiyomi-extensions/issues/17478

curche:
While I appreciate your commitment to maintaining the script/gist so that the extension can figure out the keys in real-time (for as long as you maintain it and they use the same setup), ...
...I believe it'd be better to merge both efforts (#17585 and #17543) so that the extension has the option to either use your approach OR let the user enter the keys themselves (if you werent fast enough to update keys or are busy irl). This lets users not wait and figure out the keys themselves if they can. Maybe you can have a public gist for that node.js script sooner right now than later although I'm not sure how japscan will retaliate to that.

Now all this hinges on the fact that japscan continues to use the same algorithm and just keeps rotating keys but I assume that you are confident that they won't change it anytime soon.

tbh, given the history between tachiyomi and japscan, I'd rather just remove the extension completely and end this chapter for good but if you're trying to continue efforts on a cat-and-mouse-game, I provide my 2 cents. There are always other readers/downloaders available.

Additionally, perhaps it might be a good idea to increase the rate limit to make it even more strict
Also, the `Closes ##` has to be made in the first comment otherwise it doesn't work

zormy111:
Thank you for your answer, actually it's not just those readers but the French scans are generally very badly referenced except on this site, hence the importance for a large part of us of maintaining it. It is a very good idea to use the possibility of custom keys in parallel with the key search. This will allow anyone who knows how to decipher the scans for a short time. I made all these modifications and shared my nodejs code: https://github.com/zormy111/japscangetkey

Rajh:
I'm not an expert and I don't have that much time to dig into it, but if anyone want to improve the preferences feel free to do it :)

I just did it as a workaround when the extension break, it could use 2 settings instead of splitting it though...

alessandrojean:
I think at this point it would just be better to remove the extension, doesn't seem a good thing on depending at a external Gist. Can be removed as a cat and mouse game again, it's causing a lot of trouble and they will sure find a way to break your approach as soon it gets merged.

zormy111:
As you want it's up to you, on my side in an unnoficial way I would continue to maintain anyway, and many people will go through an unnoficial version if it is deleted, but if to make your repo clean of plugins you decide to remove I would understand. As explained previously, there is no other real reliable and complete source for French scans, even mangadex only provides very few complete lists (many missing chapters). If this does not bother you on the side of your organization, it is currently preferable in my opinion to leave it as long as there is no better source. Whatever your decision thank you for your answers and your attention for these requests 👍

PR #17590, opened 2023-08-18
HeanCms: Re-add fetchAllTitles

bapeey:
Checklist:


alessandrojean:
Can you throw an Exception in `mangaDetailsRequest`, `chapterListRequest` and `pageListRequest` if the `fetchAllTitles` is true and the `manga.url` or `chapter.url` contains a timestamp in it?

In this way, users that happened to add entries or migrated in their library these days will be forced to migrate to the permanent URL in order to keep receiving updates. Just a simple if throwing a regular Exception like MangaDex does it.

PR #17691, opened 2023-08-26
MangaFR: Fix chapterNames and mangaName while in manga details

KirinRaikage:
Checklist:

### DIFF
--- a/multisrc/overrides/mmrcms/mangafr/src/MangaFR.kt
+++ b/multisrc/overrides/mmrcms/mangafr/src/MangaFR.kt
@@ -40,15 +40,16 @@ class MangaFR : MMRCMS("Manga-FR", "https://manga-fr.me", "fr") {
         // Parse date
         val dateText = element.getElementsByClass("date-chapter-title-rtl").text().trim()
 
-        companion object {
-            val dateFormat by lazy {
-                SimpleDateFormat("d MMM. yyyy", Locale.US)
-            }
-        }
         chapter.date_upload = runCatching {
             dateFormat.parse(dateText)?.time
         }.getOrNull() ?: 0L
 
         return chapter
+    }
+
+    companion object {
+        val dateFormat by lazy {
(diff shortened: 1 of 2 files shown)


multisrc/overrides/mmrcms/mangafr/src/MangaFR.kt
```diff
@@ -0,0 +1,47 @@
...
+        val chapterTitle = titleWrapper.getElementsByTag("em")!!.text()
+        if (chapterTitle.toIntOrNull() != null) {
+            chapter.name = "Chapitre $chapterNumber"
+        } else {
+            chapter.name = "Chapitre $chapterNumber : $chapterTitle"
+        }
+
+        // Parse date
+        val dateText = element.getElementsByClass("date-chapter-title-rtl").text().trim()
+        val dateFormat = SimpleDateFormat("d MMM. yyyy", Locale.US)
```
alessandrojean, line 42: Make it a constant in a `companion object`, otherwise it will be recreated for every single chapter.

multisrc/overrides/mmrcms/mangafr/src/MangaFR.kt
```diff
@@ -0,0 +1,47 @@
...
+        chapter.date_upload = dateFormat.parse(dateText)?.time ?: 0L
```
alessandrojean, line 43: `SimpleDateFormat`'s `parse` method can throw some Exceptions, wrap it into a `runCatching` or a simple `try-catch` like suggested in the CONTRIBUTING.md guide.

alessandrojean (changes requested):

multisrc/overrides/mmrcms/mangafr/src/MangaFR.kt
KirinRaikage: runCatching looks better visually speaking, I'll stick to that.

Issue #17592, opened 2023-08-19
Mangasee can't sync chapters with tracker when chapter numbers are in fraction


12ohit:
### Source information

MangaSee 1.4.35

### Source language

English 

### Steps to reproduce

Open a Manga in MangaSee( try One-Punch Man) add tracker ( I use MAL) and then read any chapter which has fraction in it's chapter number.

That chapter won't be tracked in tracker because chapter number is wrong

### Expected behavior

The fraction chapter should also be synchronised with tracker

### Actual behavior

Chapter doesn't get synchronised with tracker 

### Tachiyomi version

0.14.6

### Android version

Android 11

### Other details

I would say add "ch. ${noOfChapter}" at the end of every chapter name for this source so even when chapter name contains fraction it will be synchronised correctly.

How the chapter number are right now:
![Screenshot_2023-08-19-13-38-02-183](https://github.com/tachiyomiorg/tachiyomi-extensions/assets/53574579/2fb5ff8b-71f2-4708-b11f-b2285429e8ae)

How they can be updated for good chapter sync with tracker(using Webtoons extension for this screenshot):
![Screenshot_2023-08-19-13-38-25-302](https://github.com/tachiyomiorg/tachiyomi-extensions/assets/53574579/13efda0c-5070-4f5b-9423-5f7f68c6d12f)

### Acknowledgements

- [X] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open or closed issue.
- [X] I have written a short but informative title.
- [X] I have updated the app to version **[0.14.6](https://github.com/tachiyomiorg/tachiyomi/releases/latest)**.
- [X] I have updated all installed extensions.
- [X] I have tried the [troubleshooting guide](https://tachiyomi.org/help/guides/troubleshooting/).
- [X] If this is an issue with the app itself, I should be opening an issue in the [app repository](https://github.com/tachiyomiorg/tachiyomi/issues/new/choose).
- [X] I will fill out all of the requested information in this form.

Seishirou101:
thats probably not something that will be done by the extension since its showing you exactly how the website shows it... it would be something that the app has to deal with, and you can make a request in the app repo, though i think it may have already been requested

PR #17708, opened 2023-08-28
fix Managasee chapters not syncing with tracker

12ohit:
Explanation: #17592 

I am not that familiar to Kotlin so code can be a little ugly, please let me know if any suggestions or corrections in code

Checklist:


12ohit:
> If a series has chapter one divided in 5 part (1.1, 1.2, ...) then chapter 2 will be marked as chapter 6

But that's how MAL want it the fraction chapters are also synchronised as a chapter

AntsyLich:
What MAL wants is the total chapter count not the latest chapter number.

12ohit:
Well if the chapter number is 8.5 then it won't get synced with MAL so I did this but it's ok if you think the way I am putting chapter numbers is wrong🥲

PR #18420, opened 2023-10-07
Mangaplus: Support midChapterList

edo9300:
Checklist:

The reponse json from the api other than firstChapterList and lastChapterList also returns midChapterList ([example with chainsaw man](https://jumpg-webapi.tokyo-cdn.com/api/title_detailV3?title_id=100037&format=json)), that apparently in this case even got a 7th chapter available.
Also update the error message displaying as the ``language`` field in the returned json is not populated for the english error message ([example](https://jumpg-webapi.tokyo-cdn.com/api/title_detailV3?title_id=1007&format=json)). (I just hcked up something for this, don't really know how to do things with kotlin.)


alessandrojean:
The midChapterList is hidden in the app by design to replicate how the site does it, as we don't want users to have access to things they can't already see in their site. Appreciate the PR, but will close because of this.

edo9300:
Yeah, I see why you'd not want to add this (in fact I have other modifications I won't publish for the same reasons), I just assumed that since that info is now being returned by the public api it shouldn't have been much of an issue.

Issue #6804, opened 2021-05-03
[Source Request] project suki


daminokun:
## Source website
- Name: Project Suki
- Link: https://projectsuki.com/
- Language(s): English

keinezeit8:
duplicate of #6275

Issue #6275, opened 2021-03-24
[Source Request] <ProjectSuki>


hjdrm:
## Source website
- Name: Project Suki
- Link: projectsuki.com
- Language(s): English

## Other details
Similar app to Mangadex that lets scanlator groups to upload their stuff.

PR #18774, opened 2023-10-28
New Source: Project Suki

npgx:
### Checklist:

### Not needed:

### Related Issues:
#6275 - mentions Project Suki
#6804 - mentions Project Suki

### Comment:
I've tested the extension for a reasonable amount of mangas, everything seems to be working.
Latest isn't supported because Project Suki doesn't exactly differentiate between trending, latest etc... in the main page.
Basic idea is to just grab all the mangas by url and figure out the thumbnail and title.

I'm unsure about the language filtering of chapters, so would appreciate the help/review of anyone more knowledgeable.


src/all/projectsuki/src/eu/kanade/tachiyomi/extension/all/projectsuki/NormalizedURL.kt
```diff
@@ -0,0 +1,132 @@
+package eu.kanade.tachiyomi.extension.all.projectsuki
+
+import org.jsoup.nodes.Element
+import java.util.Locale
+import java.util.TreeMap
+
+/** @param domains last is TLD, 2nd last is domain */
+data class NormalizedURL(
```
alessandrojean, line 8: Why having all this work instead of just using OkHttp's `HttpUrl` builder?

src/all/projectsuki/src/eu/kanade/tachiyomi/extension/all/projectsuki/ProjectSuki.kt
```diff
@@ -0,0 +1,494 @@
...
+import rx.Observable
+import uy.kohesive.injekt.Injekt
+import uy.kohesive.injekt.api.get
+import java.util.Locale
+
+@Suppress("unused")
+class ProjectSuki : HttpSource(), ConfigurableSource {
+    override val name: String = PS.name
+    override val baseUrl: String = "${PS.protocol}://${PS.domain}"
+    override val lang: String = PS.lang
```
alessandrojean, line 47: Doesn't need to have these constants, just put it directly like all sources do.

src/all/projectsuki/src/eu/kanade/tachiyomi/extension/all/projectsuki/ProjectSuki.kt
```diff
@@ -0,0 +1,494 @@
...
+    private val preferences: SharedPreferences by lazy {
+        Injekt.get<Application>().getSharedPreferences("source_$id", 0x0000)
+    }
+
+    private fun String.processLangPref(): List<String> = split(",").map { it.trim().lowercase(Locale.US) }
+
+    @Volatile
+    private var whitelistedLanguages: List<String> = preferences
+        .getString(PS.PREFERENCE_WHITELIST_LANGUAGES, "")!!
+        .processLangPref()
```
alessandrojean, line 58: Using it like this will require users restarting the app to the settings to apply. Use this instead:

```suggestion
    private val SharedPreferences.whitelistedLanguages: List<String>
        get() = getString(PS.PREFERENCE_WHITELIST_LANGUAGES, "")!!
            .processLangPref()
```

Then use it as `preferences.whitelistedLanguages`.

src/all/projectsuki/src/eu/kanade/tachiyomi/extension/all/projectsuki/ProjectSuki.kt
```diff
@@ -0,0 +1,494 @@
...
+
+    @Volatile
+    private var blacklistedLanguages: List<String> = preferences
+        .getString(PS.PREFERENCE_BLACKLIST_LANGUAGES, "")!!
+        .processLangPref()
```
alessandrojean, line 63: Same here.

src/all/projectsuki/src/eu/kanade/tachiyomi/extension/all/projectsuki/ProjectSuki.kt
```diff
@@ -0,0 +1,494 @@
...
+                summary = PS.PREFERENCE_WHITELIST_LANGUAGES_SUMMARY
+                setOnPreferenceChangeListener { _, newValue ->
+                    when (val value = ((newValue as? String) ?: newValue?.toString())) {
+                        null -> false
+                        else -> {
+                            whitelistedLanguages = value.processLangPref()
+                            true
+                        }
+                    }
+                }
```
alessandrojean, line 81: Can remove as the suggested extension function will already have the most updated value.

```suggestion
```

src/all/projectsuki/src/eu/kanade/tachiyomi/extension/all/projectsuki/ProjectSuki.kt
```diff
@@ -0,0 +1,494 @@
...
+                summary = PS.PREFERENCE_BLACKLIST_LANGUAGES_SUMMARY
+                setOnPreferenceChangeListener { _, newValue ->
+                    when (val value = ((newValue as? String) ?: newValue?.toString())) {
+                        null -> false
+                        else -> {
+                            blacklistedLanguages = value.processLangPref()
+                            true
+                        }
+                    }
+                }
```
alessandrojean, line 98: Same here.

```suggestion
```

src/all/projectsuki/src/eu/kanade/tachiyomi/extension/all/projectsuki/ProjectSuki.kt
```diff
@@ -0,0 +1,494 @@
...
+            },
+        )
+    }
+
+    override val client: OkHttpClient by lazy {
+        debug { "INIT -> OkHttpClient" }
```
alessandrojean, line 104: Remove these log statements. Also the client doesn't need to be build by lazy.

src/all/projectsuki/src/eu/kanade/tachiyomi/extension/all/projectsuki/ProjectSuki.kt
```diff
@@ -0,0 +1,494 @@
...
+            .add("Content-Type", "application/json;charset=UTF-8")
+            .build()
+
+        val callpageBody = Json.encodeToString(
+            @Suppress("RemoveExplicitTypeArguments")
+            buildMap<String, String> {
+                this["bookid"] = bookid
+                this["chapterid"] = chapterid
+                this["first"] = true.toString()
+            },
```
alessandrojean, line 451: Can use `mapOf`.

```suggestion
            mapOf(
                "bookid" to bookid,
                "chapterid" to chapterid,
                "first" to "true",
            ),
```

alessandrojean (changes requested):

src/all/projectsuki/src/eu/kanade/tachiyomi/extension/all/projectsuki/NormalizedURL.kt
npgx: honestly, never used OkHttp, I could change it.
I assume you're saying it because OkHttp can normalize Urls?

alessandrojean: It is supposed to do, yes.

npgx: I tested it on a couple of mangas and it seems to be working

npgx:
One thing to note:
when searching for mangas, it will always report that there are "no results found" when scrolling to the end of the results.
On projectsuki there is always a "next" page button when there's at least one manga in the current window, even if the next page will have 0 results, as such hasNextPage is always true when creating the MangasPage for the search results.

Any clues on how to fix this?
(there always seem to be 30 manga in a full results page, but that could be changes easily and there could still be 0 results in the following page even if the current one has 30 mangas)

alessandrojean:
No problem with it, you can make it return false if the length of the page <= 30 and it's fine as well. The button always appearing on the site is something they need to fix.

npgx:
A single request to a very generic search like 'a' would give you the updated max number of results in the page, but that wouldn't fix the edge case where the current page is full and the next one is empty, so 30 is good-enough.

alessandrojean (commented):
Seems okay now, just need to remove all the `Log` and `debug` statements, otherwise they will end up appearing in the crash logs and that's not desired.

npgx:
Tested and everything seems to _still_ be working

npgx:
Found some edge cases where title shows up as "show more", thought I fixed them, will investigate.

PR #18946, opened 2023-11-13
Restore a new source to Madara

matheus-caldeira:
Checklist:


arkon:
Are you willing to continue maintaining this when the source continually makes changes to break it? If not, then this won't be accepted.

matheus-caldeira:
Yes, I am committed to keeping this updated, as it is the main scan I use for reading. If the source continually makes changes that could potentially break it, I will adapt and update accordingly to ensure its continued functionality.

alessandrojean:
Appreciate the PR, but gonna deny as this source in particular caused a lot of headache to our support team back then. They're known for changing their titles URLs almost daily and even broke the WordPress direct link by ID on purpose to avoid apps like Tachiyomi. Nowadays aggregators already reupload their content anyway, so really not worth it.

PR #19060, opened 2023-11-21
Update Fansubs.cat API calls, add new NSFW site, and implement search filters

fansubscat:
Fansubs.cat no longer hosts NSFW content. A separate NSFW site has been launched, named "Fansubs.cat - Hentai". Changes have been made to the API calls (which now differ between sites) and it is now possible to filter searches by additional parameters.

This PR:

- Implements a new extension for the NSFW site ("Fansubs.cat - Hentai")
- Updates the API calls to their current implementation
- Implements new search filters
- Sets `isNsfw = false` for the main site ("Fansubs.cat")
- Updates the site logo

Checklist:


vetleledaal:
There is a lot of duplicate code, you should consider using a `SourceFactory` to avoid that.

fansubscat:
> There is a lot of duplicate code, you should consider using a `SourceFactory` to avoid that. Or is that not possible when `isNsfw` is different?

Yes, that would be better, but unfortunately, if we use a `SourceFactory`, we would have to declare both sites as `isNsfw = true` since this attribute is set at the extension level.

alessandrojean:
Can use a `ThemeGenerator` alternatively, would fit better since it's two separate extensions. Using a `ThemeGenerator` would allow you to set the `isNsfw` only in one source.

multisrc/overrides/fansubscat/fansubscathentai/src/FansubsCatHentai.kt
```diff
@@ -0,0 +1,10 @@
+package eu.kanade.tachiyomi.extension.ca.fansubscathentai
+
+import eu.kanade.tachiyomi.multisrc.fansubscat.FansubsCat
+
+class FansubsCatHentai : FansubsCat(
+    "Fansubs.cat - Hentai",
+    "https://hentai.fansubs.cat/manga",
+    "ca",
+    true,
```
vetleledaal, line 9: I think it would be beneficial to explicitly name uncommon arguments, so `isHentaiSite = true` here and `isHentaiSite = false` for the other site.

fansubscat:
Thanks for your help. This is now a multisrc that generates two extensions: "Fansubs.cat" (SFW) and "Fansubs.cat - Hentai" (NSFW). We have explicitly kept the old package name for the main (SFW) extension. Please let us know if there are further changes needed.

multisrc/src/main/java/eu/kanade/tachiyomi/multisrc/fansubscat/FansubsCat.kt
```diff
@@ -0,0 +1,417 @@
...
+    override fun mangaDetailsRequest(manga: SManga): Request {
+        return GET("$baseUrl/${manga.url}", headers)
+    }
+
+    private fun apiMangaDetailsRequest(manga: SManga): Request {
+        return GET(
+            "$apiBaseUrl/manga/details/${manga.url.substringAfterLast('/')}?hentai=$isHentaiSite",
+            headers,
+        )
+    }
```
AwkwardPeak7, line 149:
```suggestion
    override fun mangaDetailsRequest(manga: SManga): Request {
        return GET(
            "$apiBaseUrl/manga/details/${manga.url.substringAfterLast('/')}?hentai=$isHentaiSite",
            headers,
        )
    }
    
    override fun getMangaUrl(manga: SManga): String {
        return "$baseUrl/${manga.url}"
    }
```
This kind of workaround is no longer needed as you can override `getMangaUrl`/`getChapterUrl` for correct webview links.

multisrc/src/main/java/eu/kanade/tachiyomi/multisrc/fansubscat/FansubsCat.kt
```diff
@@ -0,0 +1,417 @@
...
+        return GET(
+            "$apiBaseUrl/manga/pages/${chapter.url.substringAfterLast('/')}?hentai=$isHentaiSite",
+            headers,
+        )
+    }
+
+    // Return the real URL for "Open in browser"
+    override fun pageListRequest(chapter: SChapter): Request {
+        return GET("$baseUrl/${chapter.url.replace("/", "?f=")}", headers)
+    }
```
AwkwardPeak7, line 202:
```suggestion
    override fun pageListRequest(chapter: SChapter): Request {
        return GET(
            "$apiBaseUrl/manga/pages/${chapter.url.substringAfterLast('/')}?hentai=$isHentaiSite",
            headers,
        )
    }

    override fun getChapterUrl(chapter: SChapter): String {
        return "$baseUrl/${chapter.url.replace("/", "?f=")}"
    }
```
same here

fansubscat:
Updated as suggested, thanks!

Issue #18885, opened 2023-11-07
999Hentai: Tag filter


Opi-Txm:
### Source name

999Hentai 1.4.3

### Source language

Multi

### Describe your suggested feature

A tag filter should be implemented, similar to the one in NHentai (Image below).

Also, when clicking on a tag in a manga it should filter and only show the mangas that also have the tag.
Currently, clicking on a tag e.g. "Animated" would search for mangas containing "Animated" in the title regardless of the tag.

![NHentai filtering](https://github.com/tachiyomiorg/tachiyomi-extensions/assets/106656898/f39861b3-bf34-4b1d-8158-2b44331cfc58)

### Acknowledgements

- [X] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open or closed issue.
- [X] I have written a short but informative title.
- [X] If this is an issue with the app itself, I should be opening an issue in the [app repository](https://github.com/tachiyomiorg/tachiyomi/issues/new/choose).
- [X] I have updated the app to version **[0.14.7](https://github.com/tachiyomiorg/tachiyomi/releases/latest)**.
- [X] I will fill out all of the requested information in this form.

PR #19105, opened 2023-11-26
999hentai tag search

Opi-Txm:
Closes: [#18885](https://github.com/tachiyomiorg/tachiyomi-extensions/issues/18885)

I changed the tag search to a ```Filter.Select``` type for the following reasons:

1. When a you search a genre by clicking on a tag, the function below eventually gets called in the main tachiyomi repository. The type ```Filter.Text``` won't get catched in this method and ends up just searching for the tag in the search bar. Also after skimming all sources, the only other source that has a genre tag filter of type ```Filter.Text``` is **nhentai**, which has a built-in tag searching algorithm in their website and therefore is appropriate.
```kotlin
//When a Tag gets clicked, performGenreSearch() in app/src/main/java/eu/kanade/tachiyomi/ui/manga/MangaScreen.kt 
//gets called, which calls searchGenre() in BrowseSourceScreen.kt in the same folder, which calls the method below

app/src/main/java/eu/kanade/tachiyomi/ui/browse/source/browse/BrowseSourceScreenModel.kt 

fun searchGenre(genreName: String) {
        if (source !is CatalogueSource) return

        val defaultFilters = source.getFilterList()
        var genreExists = false

        filter@ for (sourceFilter in defaultFilters) {
            if (sourceFilter is SourceModelFilter.Group<*>) {
                for (filter in sourceFilter.state) {
                    if (filter is SourceModelFilter<*> && filter.name.equals(genreName, true)) {
                        when (filter) {
                            is SourceModelFilter.TriState -> filter.state = 1
                            is SourceModelFilter.CheckBox -> filter.state = true
                            else -> {}
                        }
                        genreExists = true
                        break@filter
                    }
                }
            } else if (sourceFilter is SourceModelFilter.Select<*>) {
                val index = sourceFilter.values.filterIsInstance<String>()
                    .indexOfFirst { it.equals(genreName, true) }

                if (index != -1) {
                    sourceFilter.state = index
                    genreExists = true
                    break
                }
            }
        }

        mutableState.update {
            val listing = if (genreExists) {
                Listing.Search(query = null, filters = defaultFilters)
            } else {
                Listing.Search(query = genreName, filters = defaultFilters)
            }
            it.copy(
                filters = defaultFilters,
                listing = listing,
                toolbarQuery = listing.query,
            )
        }
    }


```
2. The structure of the website **999hentai** allows multiple exclusion tags, but only one including tag. If you search multiple tags in the tachiyomi application, e.g. "uncensored, animated", the result is a union of single tag searches, which I don't think is an expected behavior for many users. (For this example, you will have a list of mangas containing either the tag "uncensored" or "animated") Therefore, instead of ```Filter.Group```, I implemented it with ```Filter.Select```.

Also fixed a bug that ignores the filter of min/max pages when only the max amount of pages is set.

Checklist:

### DIFF
--- a/src/all/ninenineninehentai/src/eu/kanade/tachiyomi/extension/all/ninenineninehentai/NineNineNineHentai.kt
+++ b/src/all/ninenineninehentai/src/eu/kanade/tachiyomi/extension/all/ninenineninehentai/NineNineNineHentai.kt
@@ -115,7 +115,7 @@ open class NineNineNineHentai(
                     format = filters.firstInstanceOrNull<FormatFilter>()?.selected,
                     tags = filters.firstInstanceOrNull<IncludedTagFilter>()?.selected,
                     excludeTags = filters.firstInstanceOrNull<ExcludedTagFilter>()?.tags,
-                    pagesRangeStart = filters.firstInstanceOrNull<MinPageFilter>()?.value,
+                    pagesRangeStart = if (filters.firstInstanceOrNull<MinPageFilter>()?.value == null && filters.firstInstanceOrNull<MaxPageFilter>()?.value != null) 1 else filters.firstInstanceOrNull<MinPageFilter>()?.value,
                     pagesRangeEnd = filters.firstInstanceOrNull<MaxPageFilter>()?.value,
                 ),
             ),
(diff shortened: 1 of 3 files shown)


AwkwardPeak7 (commented):
The site have 100s of tags so having it implemented as a text filter allows the user to search for any of them, while having a small subset of them hardcoded in the extension won't allow that. 

also
> Updated extVersionCode value in build.gradle for individual extensions

Opi-Txm:
> The site have 100s of tags so having it implemented as a text filter allows the user to search for any of them, while having a small subset of them hardcoded in the extension won't allow that.

What let me decide to change it in this way is that some sources I skimmed through, that has a lot of tags, also used hard coded tags, e.g. EHentai (not as much as 999hentai but still a lot), and that as mentioned, Filter.Text won't get catched in the searchGenre() method in main.

A temporary solution is to keep both Tagfilter .Text and .Select, which works but is not so clean and nice.
I'll try to find a better way though.

>Updated extVersionCode value in build.gradle for individual extensions

Updated it. Thanks for letting me know.

CarlosEsco:
Marking this draft until an idea is decided on

Issue #14581, opened 2022-12-17
ReadComicOnline: Publisher Filter


lvlahraam:
### Source name and language

ReadComicOnline

### Describe your suggested feature

a publisher filter that can show the comics that the publishers have published
this feature already exists in the website

### Other details

it could be a entry box (a box where the user can write the name of the publisher) and you can get the publisher by that cause there could be a alot of publishers and the filter section might look bad and also there is a possiblity where there is a new publisher and someone has to update the publishers list in the filter section

so having a text/entry box for the user to write/type in it is the best way in my opinion

### Acknowledgements

- [X] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open or closed issue.
- [X] I have written a short but informative title.
- [X] If this is an issue with the app itself, I should be opening an issue in the [app repository](https://github.com/tachiyomiorg/tachiyomi/issues/new/choose).
- [X] I have updated the app to version **[0.14.2](https://github.com/tachiyomiorg/tachiyomi/releases/latest)**.
- [X] I will fill out all of the requested information in this form.

PR #19262, opened 2023-12-10
Readcomiconline: Added new filters

Opi-Txm:
Closes [#14581](https://github.com/tachiyomiorg/tachiyomi-extensions/issues/14581)
Additionally to the Publisher filter the issue wanted to have, I implemented Writer and Artist filter as well and the 4 sorting orders when one of the three is set.

I couldn't implement it sothat the new filters works with the tag and title search that was originally there, because when you want to search Publishers/Writers/Artists, you will have to go to this kind of an url: 
```https://readcomiconline.li/Publisher/DC-Comics/```,
which is varies from the original link: 
```https://readcomiconline.li/AdvanceSearch?comicName=&ig=&eg=41%2C&status=```

So the new filters is only activated the query and the genre tag filter is not set.
Also, when you set multiple of the new filters (e.g. Publisher = DC Comics, Artist = Adam DeKraker), the priority of the three filters looks like the following: **Publisher > Writer > Artist.** 
So in the example, Only Publisher = DC Comics is going to get applied.

Checklist:

### DIFF
--- a/src/en/readcomiconline/build.gradle
+++ b/src/en/readcomiconline/build.gradle
@@ -5,7 +5,7 @@ ext {
     extName = 'ReadComicOnline'
     pkgNameSuffix = 'en.readcomiconline'
     extClass = '.Readcomiconline'
-    extVersionCode = 16
+    extVersionCode = 17
 }
 
 dependencies {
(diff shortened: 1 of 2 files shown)


Opi-Txm:
Thanks for letting me know.
I'll try to fix it and create a new PR or just make a new issue.

Opi-Txm:
This [PR](https://github.com/tachiyomiorg/tachiyomi-extensions/pull/19315) should fix it.

PR #19341, opened 2023-12-17
MangaPark v5

### LINKED ISSUE #15310: Mangapark v5
### Source name and language

eu.kanade.tachiyomi.extension.all.mangapark

This extension should have the support to mangapark v5 new website. Some titles are only in v5 and not in V3.
Also, some chapters are doubled in the site, can you add the support to view both chapters (same you did for mangadex)?

### Other details

_No response_

### Acknowledgements

AwkwardPeak7:
closes #15310 

Changed source Name from "MangaPark v3" -> "MangaPark"
Users will have to migrate existing library entries

Checklist:


src/all/mangapark/src/eu/kanade/tachiyomi/extension/all/mangapark/MangaParkDto.kt
```diff
@@ -0,0 +1,132 @@
...
+        title = name
+        thumbnail_url = cover
+        author = authors?.joinToString()
+        artist = artists?.joinToString()
+        description = buildString {
+            summary?.let { append(Jsoup.parse(it).text(), "\n\n") }
+            altNames?.takeUnless { it.isEmpty() }
+                ?.joinToString("\n") { "• ${it.trim()}" }
+                ?.let { append("Alternative Names:\n", it) }
+        }
```
vetleledaal, line 55: Consider trimming or reworking the logic. If `summary` is present, but not `altNames` you end up with 2 extra newlines. `summary` might also contain leading whitespace.

src/all/mangapark/src/eu/kanade/tachiyomi/extension/all/mangapark/MangaParkDto.kt
```diff
@@ -0,0 +1,132 @@
...
+            "cancelled" -> SManga.CANCELLED
+            else -> SManga.UNKNOWN
+        }
+        initialized = true
+    }
+
+    companion object {
+        private fun String.toCamelCase(): String {
+            if (isEmpty()) {
+                return this
```
vetleledaal, line 76: Wouldn't it alternatively execute:
```kt
val result = StringBuilder(0)
var capitalize = true
return result.toString()
```

Would remove the check unless you're doing it like this for performance reasons.

src/all/mangapark/src/eu/kanade/tachiyomi/extension/all/mangapark/MangaParkFilters.kt
```diff
@@ -2,300 +2,132 @@ package eu.kanade.tachiyomi.extension.all.mangapark
...
+            Pair("99~90", "90-99"),
+            Pair("89~80", "80-89"),
+            Pair("79~70", "70-79"),
+            Pair("69~60", "60-69"),
+            Pair("59~50", "50-59"),
+            Pair("49~40", "40-49"),
+            Pair("39~30", "30-39"),
+            Pair("29~20", "20-29"),
+            Pair("19~10", "10-19"),
+            Pair("9~1", "1-9"),
```
vetleledaal, line 130: Why reverse the order? Assuming first is the display name, i.e. why 9\~1 and not 1\~9?
On second thought, is it because they are listed in descending order?

AwkwardPeak7: that's just scrapped from the site 😅

src/all/mangapark/src/eu/kanade/tachiyomi/extension/all/mangapark/MangaParkDto.kt
AwkwardPeak7: `Jsoup.parse` generally removes any newlines in the string, so I don't think trimming is necessary

vetleledaal: The reason why I mentioned it is because I recently ran into an issue with whitespace, didn't look into the root case though https://github.com/tachiyomiorg/tachiyomi-extensions/pull/19239/files#diff-339573f5479832d53c22d410b7cda7d6915ca0d19233242c38a5d2bf3b5f4be1R156

The net effect was that you had to scroll 2 screens down for the chapters to show. Don't have the IDE in front of me, so if the same situation can't happen here it should be fine to leave as-is

Issue #19394, opened 2023-12-22
Picacomic: Crashes the app due to Gson depencency on JWT library


YangHgRi:
### Source information

Picacomic 1.4.2

### Source language

Simp Chinese

### Steps to reproduce

Use all the features associated with it
e.g. open source, view the manga in library but from that source, migrate manga from or into this extension

### Expected behavior

normal

### Actual behavior

make app crashed

### Tachiyomi version

0.14.7

### Android version

Android 13

### Other details

![Screenshot_2023-12-22-16-52-30-899_eu kanade tachiyomi-edit](https://github.com/tachiyomiorg/tachiyomi-extensions/assets/57426791/63970037-409c-4314-9094-6f7e2c50ffbb)

[tachiyomi_crash_logs.txt](https://github.com/tachiyomiorg/tachiyomi-extensions/files/13755330/tachiyomi_crash_logs.txt)

### Acknowledgements

- [X] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open or closed issue.
- [X] I have written a short but informative title.
- [X] I have updated the app to version **[0.14.7](https://github.com/tachiyomiorg/tachiyomi/releases/latest)**.
- [X] I have updated all installed extensions.
- [X] I have tried the [troubleshooting guide](https://tachiyomi.org/help/guides/troubleshooting/).
- [X] If this is an issue with the app itself, I should be opening an issue in the [app repository](https://github.com/tachiyomiorg/tachiyomi/issues/new/choose).
- [X] I will fill out all of the requested information in this form.

PR #19400, opened 2023-12-23
Fix the problem that the GSON dependency brought by the JWT library causes the APP to crash in picacomic extension

YangHgRi:
Checklist:

Closes https://github.com/tachiyomiorg/tachiyomi-extensions/issues/19394


TheKingTermux (commented):
You forget to bump extVersionCode

~~extVersionCode = 2~~
extVersionCode = 3

Then add this to your Checklist
> Closes #19394

Then check this
> [x] Referenced all related issues in the PR body (e.g. "Closes #xyz")

You also forget to check this because you forget to bump extVersionCode
> [x] Updated `extVersionCode` value in `build.gradle` for individual extensions

You can uncheck this because the Name, Theme or Language didn't change
> [ ] Have explicitly kept the 'id' if a source's name or language were changed

You can uncheck this too because it aleardy on that extension 
> [ ] Added the `isNsfw = true` flag in `build.gradle` when appropriate

YangHgRi:
> You forget to bump extVersionCode
> 
> ~extVersionCode = 2~ extVersionCode = 3
> 
> Then add this to your Checklist
> 
> > Closes #19394
> 
> Then check this
> 
> > * [x]  Referenced all related issues in the PR body (e.g. "Closes #xyz")
> 
> You also forget to check this because you forget to bump extVersionCode
> 
> > * [x]  Updated `extVersionCode` value in `build.gradle` for individual extensions
> 
> You can uncheck this because the Name, Theme or Language didn't change
> 
> > * [ ]  Have explicitly kept the `id` if a source's name or language were changed
> 
> You can uncheck this too because it aleardy on that extension
> 
> > * [ ]  Added the `isNsfw = true` flag in `build.gradle` when appropriate

got it

src/zh/picacomic/src/eu/kanade/tachiyomi/extension/zh/picacomic/Picacomic.kt
```diff
@@ -80,6 +86,43 @@ class Picacomic : HttpSource(), ConfigurableSource {
         t
     }
 
+    @RequiresApi(Build.VERSION_CODES.O)
```
arkon, line 89: Can't do this, this will crash on older versions. If you need `Base64`, use https://developer.android.com/reference/android/util/Base64

arkon (changes requested):

src/zh/picacomic/src/eu/kanade/tachiyomi/extension/zh/picacomic/Picacomic.kt
YangHgRi: > Can't do this, this will crash on older versions. If you need `Base64`, use https://developer.android.com/reference/android/util/Base64

got fixed in latest commit, thank u, as u see, i'm a beginner for android development

PR #19449, opened 2023-12-28
Update Solar and Sundry API

GRA0007:
Updates the [Solar and Sundry](https://sas.ewanb.me/) extension to use the new version of the API, and also includes a new image and description.

Checklist:


alessandrojean:
While you're into it, if you can also do the migration of org.json to kotlinx.serialization, it would be appreciated.

src/en/solarandsundry/src/eu/kanade/tachiyomi/extension/en/solarandsundry/SolarAndSundry.kt
```diff
@@ -123,9 +127,9 @@ class SolarAndSundry : HttpSource() {
     // Pages
 
     override fun pageListParse(response: Response): List<Page> {
-        val page = JSONObject(response.body.string())
+        val page = Json.decodeFromString<SasPage>(response.body.string())
```
arkon, line 130: You can use the singleton instance the app manages -- `private val json: Json by injectLazy()`

PR #19494, opened 2023-12-30
Add Repaer Scans Unoriginal

AnshumanSarkar2001:
Closes https://github.com/tachiyomiorg/tachiyomi-extensions/issues/19185

Checklist:


multisrc/src/main/java/eu/kanade/tachiyomi/multisrc/mangathemesia/MangaThemesiaGenerator.kt
```diff
@@ -129,6 +129,7 @@ class MangaThemesiaGenerator : ThemeSourceGenerator {
         SingleLang("Rawkuma", "https://rawkuma.com/", "ja"),
         SingleLang("ReadGojo", "https://readgojo.com", "en"),
         SingleLang("Readkomik", "https://readkomik.com", "en", className = "ReadKomik", overrideVersionCode = 1),
+        SingleLang("Reaper-Scans (unoriginal)", "https://reaper-scans.com", "en"),
```
alessandrojean, line 132: This won't work as the class name and package name generated will be invalid.

```suggestion
        SingleLang("Reaper Scans (Unoriginal)", "https://reaper-scans.com", "en", className = "ReaperScansUnoriginal"),
```

Have you really tested and compiled through Android Studio?

alessandrojean (changes requested):

multisrc/src/main/java/eu/kanade/tachiyomi/multisrc/mangathemesia/MangaThemesiaGenerator.kt
AnshumanSarkar2001: Yes I did. So after attaching the class it will work right?

alessandrojean: It's weird as it won't have compiled as the build confirms by failing.

The className parameter is supposed to fix, yes. But really compile locally and try installing and check if it works on the app.

PR #19502, opened 2023-12-31
Fix order of chapters/volumes for Manhuagui.

ThomasMengMXC:
Groups chapters/volumes and special chapters together into blocks for Manhuagui chapters.

Manhuagui provides sources for volumes, chapters, and special chapters however the current implementation simply sorts the chapters in Tachiyomi by the numerical value in the chapter title (currently line 322 of Manhuagui.kt). This would lead to chapters and volumes being sorted together similar to the example below.

- Chapter 03
- Volume 03
- Chapter 02
- Volume 02
- Special chapter 01
- Chapter 01
- Volume 01
- etc

Additionally, special chapters which have numbers in the title will be sorted out of order.

The updated code now takes the original order shown on Manhuagui. which should be mostly the correct order.

If there is anything else I need to change/do, please let me know 😄.

Checklist:


ThomasMengMXC:
Accidentally used the wrong account to commit the changes. Going to resubmit

PR #19323, opened 2023-12-15
Project Suki: Enhancements and fixes

npgx:
### Checklist:

- [x] Updated `extVersionCode` value in `build.gradle` for individual extensions
- [x] Updated `overrideVersionCode` or `baseVersionCode` as needed for all multisrc extensions
- [x] Referenced all related issues in the PR body (e.g. "Closes #xyz")
- [x] Added the `isNsfw = true` flag in `build.gradle` when appropriate
- [x] Have not changed source names
- [x] Have explicitly kept the `id` if a source's name or language were changed
- [x] Have tested the modifications by compiling and running the extension through Android Studio

### Comment:

This is a draft as I still need to do some work and test it thoroughly, but it's at a good enough point.
Will try to write up on what changed near the end when this is going to be ready for review and merge, but most info should be in the commit messages.
Feel free to leave suggestions.

(For anyone curious I'm experimenting with these: https://www.conventionalcommits.org/en/v1.0.0/)

PR #19513, opened 2023-12-31
Project Suki: refactors and fixes

npgx:
### Checklist:

### Comment:

Found a rare bug that slipped through during daily use.
Also refactored some code, added activities etc.
see CHANGELOG for full difference.

Multiple activities have been used because of [this](https://github.com/tachiyomiorg/tachiyomi-extensions/pull/19323#issuecomment-1858932113).
Unfortunately the options are multiple activities or Java.

Will be draft until fully tested again.
Feel free to leave suggestions.


npgx:
### Testing done

- [ ] Checked "Popular" tab
- [ ] Checked "Latest" tab
- [ ] Checked all Popular books for errors (open, wait to load, close)
- [ ] Checked bunch of site-search results for errors
- [ ] Checked activities
- [ ] Checked open link in browser
- [ ] Checked URL as search query
- [ ] Checked truncated URL as search query (without "projectsuki.com")
- [ ] Checked some entries' chapters for errors
- [ ] Checked language whitelist and blacklist
- [ ] Checked site-search filters

src/all/projectsuki/src/eu/kanade/tachiyomi/extension/all/projectsuki/SmartBookSearchHandler.kt
```diff
@@ -0,0 +1,223 @@
...
+private typealias IsWord = Boolean
+
+/**
+ * Handles the "Smart" book search mode, requires [API][RequiresApi] [Build.VERSION_CODES.N]
+ * because it uses [android.icu.text] classes.
+ *
+ * Tries to avoid wasting resource by normalizing text only once and using [ThreadLocal] instances.
+ */
+@Suppress("MemberVisibilityCanBePrivate")
+@RequiresApi(Build.VERSION_CODES.N)
```
alessandrojean, line 33: This won't be good as the app supports devices older than Nougat. Need to do in a way that doesn't depend on platform specific code, unless you're already handling this and I didn't find where in the code.

src/all/projectsuki/src/eu/kanade/tachiyomi/extension/all/projectsuki/DataExtractor.kt
```diff
@@ -254,37 +278,122 @@ class DataExtractor(val extractionElement: Element) {
...
-        OFFICIAL("Official:", """official:?""".toRegex(RegexOption.IGNORE_CASE)),
-        PURCHASE("Purchase:", """purchase:?""".toRegex(RegexOption.IGNORE_CASE)),
-        GENRE("Genres:", """genre(?:\(s\))?:?""".toRegex(RegexOption.IGNORE_CASE)),
-        ;
+            }
+        }
+
+        object Views : BookDetail() {
+            override val regex: Regex = """views?:?""".toRegex(RegexOption.IGNORE_CASE)
+            override fun label(element: Element?) = "Views:"
```
alessandrojean, line 368: Why instead of dealing with regexes you just use Jsoup's `:contains` selector? The `:contains` allows you to select elements that contains given text inside it, such as `td:contains(Views)`.

alessandrojean (changes requested):

src/all/projectsuki/src/eu/kanade/tachiyomi/extension/all/projectsuki/SmartBookSearchHandler.kt
npgx: It should be handled in ProjectSuki, line 422: fetchSearchManga does this
 - first check if query comes from activity/intent
 - is it an url?
 - if it's neither, is search mode set to Smart?
   - does the user have Build.VERSION.SDK_INT < Build.VERSION_CODES.N? tell user to switch to simple mode (can be set as default)
   - if not go ahead and do stuff with SmartBookSearchHandler
 - is search mode set to Simple? use regex fallback (should work on older devices? I didn't find an error)
 - if all else fails do a /search on the website

alessandrojean: Ah, it was already in the current code, so that's why I didn't find it on the PR diff. It's fine as it is, can ignore this one now.

src/all/projectsuki/src/eu/kanade/tachiyomi/extension/all/projectsuki/DataExtractor.kt
npgx: well I wanted to match all of
 - view
 - views
 - view:
 - views:
both upper and lower case, the ? are to make both the 's' and ':' optional

I did go see some documentation and `The search is case-insensitive;` but I don't know if it can do optional characters etc

alessandrojean: Using `:contains(View)` should do the job, doesn't it?

npgx: Well, `:contains` seems to match any element that has the text in it, like `:contains(View)` would also match a paragraph containing the text "The view from the hill was great".
The regex is used for an exact match of the text (DataExtractor.kt:396), according to the Selector documentation there's a `:matches(regex)` Pseudo selector, but that takes a regex as input so it wouldn't make much of a difference.
The only place that uses that regex indirectly is DataExtractor.kt:448 that tries to extract the detail from the details table.

Also right now I'm doing this:
 - find the details table in a convoluted way (DataExtractor.kt:430)
   - basically first find what's findable via href links and find the nearest common parent
 - iterate over the children
 - get the first and second "columns" (or skip in either case if fail)
 - try to determine the detail type from the first column (regex here)
 - get the detail value from the second column

I could try to use `:contains` but I would have to filter for false positives and rewrite that part

npgx: By the way I'm manually putting line and file numbers, is there a way to reference a file + line in the current pull request?

alessandrojean: You can use `:contains` together with the element tag name, so it would already filter and be more specific (assuming you know the type of the table, such as using `tr`, `td` or so on). You can also use the contains in a parent element, such as `tr:contains(View) > td:last-child`.

npgx: unfortunately it's a table per say, it's just divs with bootstrap css styles, pretty much nothing apart from 2 ids are used in this extension

alessandrojean: > By the way I'm manually putting line and file numbers, is there a way to reference a file + line in the current pull request?

You can reference it by selecting the line in the code, clicking the three dots and copying the permalink. Pasting it here should show the code in theory.

> it's just divs with bootstrap css styles

Aren't the CSS classes reliable for selection? A lot of extensions uses just classes to select everything. If it's not the case than can keep as it is, assuming the Jsoup selectors would turn out to be more complex than the current method.

npgx: They are when automated tools are used to create the website (such as wordpress), the projectsuki website looks very hand made
html
using [Bootstrap](https://getbootstrap.com/) I think
So I try not to depend on classes as they could be changed on the fly to change the style of the website without impacting it much.

alessandrojean: Ah, they don't seem to be using the existing table element from Bootstrap, I thought it was the case when you mentioned the framework.

Yeah, such classes aren't good to rely on. No problem, can keep as it is then.

PR #19515, opened 2023-12-31
Picacomic: Fix json decoding error of non-ASCII chars (#19434)

### LINKED ISSUE #19434: "bad base-64"is displayed in PicaComic
### Source information

PicaComic 1.4.3

### Source language

zh

### Steps to reproduce

1.更新picacomic
2.打开picacomic即复现

### Expected behavior

没更新之前可以正常加载

### Tachiyomi version

0.14.7

### Android version

android 10

### Other details

_No response_

### Acknowledgements

- [X] I have updated the app to version **[0.14.7](https://github.com/tachiyomiorg/tachiyomi/releases/latest)**.

fumiama:
Closes #19434
Caused by #19400

Checklist:

I have no condition to test because the repo is too heavy, but it can work theoretically for its just a small change.

### DIFF
--- a/src/zh/picacomic/src/eu/kanade/tachiyomi/extension/zh/picacomic/Picacomic.kt
+++ b/src/zh/picacomic/src/eu/kanade/tachiyomi/extension/zh/picacomic/Picacomic.kt
@@ -30,6 +30,7 @@ import org.json.JSONObject
 import uy.kohesive.injekt.Injekt
 import uy.kohesive.injekt.api.get
 import java.net.URLEncoder
+import java.nio.charset.Charset
 import java.text.SimpleDateFormat
 import java.util.Date
 import java.util.Locale
@@ -100,7 +101,7 @@ class Picacomic : HttpSource(), ConfigurableSource {
             )
         }
 
-        val payload = parts[1]?.let { JSONObject(Base64.decode(it, Base64.DEFAULT).decodeToString()) }
+        val payload = parts[1]?.let { JSONObject(Base64.decode(it, Base64.DEFAULT).toString(Charsets.UTF_8)) }
 
         val exp = payload?.getLong("exp")?.let {
             Date(it * 1000)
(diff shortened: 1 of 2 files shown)


src/zh/picacomic/src/eu/kanade/tachiyomi/extension/zh/picacomic/Picacomic.kt
```diff
@@ -100,7 +100,7 @@ class Picacomic : HttpSource(), ConfigurableSource {
             )
         }
 
-        val payload = parts[1]?.let { JSONObject(String(Base64.decode(it, Base64.DEFAULT))) }
+        val payload = parts[1]?.let { JSONObject(Base64.decode(it, Base64.DEFAULT).decodeToString()) }
```
alessandrojean, line 103: While you're into it, if you can remove the `org.json` usages in favor of `kotlinx.serialization` (as the extension already uses in other places and even has a `Json` instance), it would be better. It's not a good practice to mix two parsers.

fumiama: That's all right.

fumiama (commented):
Modified.

drelover:
I have updated the new version，but the  same problem still exist.

fumiama:
> I have updated the new version，but the  same problem still exist.

Yes, I have also updated and the same error occurs again. Maybe we should refer to some other parts. The newly introduced json lib may be a possible reason.
I am quite sure that this problem is caused by this line, because there is a symptom that when you reinstall it completely, the problem fixed for the first time and then occurred again.

alessandrojean:
>  The newly introduced json lib may be a possible reason

No as the line modified in particular doesn't even use kotlinx.serialization, it is still using the old `org.json.JSONObject`, which should be already replaced at this point.

fumiama:
> Please do follow the PR checklist next time and compile locally and test before submitting new untested PRs.

OK. Maybe it will cost me a few days or weeks to install all the environments so I will left this problem to someone else who would like to fix.
May my thoughts mentioned above help you.

PR #19537, opened 2024-01-02
[skip ci] Update the contributing guide to add alerts

alessandrojean:
Also added information about the Android Studio Network Inspector.

Checklist:


CONTRIBUTING.md
```diff
@@ -279,12 +279,14 @@ dependencies {
 If you find yourself needing additional functionality, you can add more dependencies to your `build.gradle` file.
 Many of [the dependencies](https://github.com/tachiyomiorg/tachiyomi/blob/master/app/build.gradle.kts) from the main Tachiyomi app are exposed to extensions by default.
 
-> Note that several dependencies are already exposed to all extensions via Gradle version catalog.
-> To view which are available view `libs.versions.toml` under the `gradle` folder
+> [!NOTE]
+> Several dependencies are already exposed to all extensions via Gradle version catalog.
```
vetleledaal, line 283:
```suggestion
> Several dependencies are already exposed to all extensions via Gradle's version catalog.
```
Sounds like it should be "a version catalog", not "the version catalog"

CONTRIBUTING.md
```diff
@@ -279,12 +279,14 @@ dependencies {
...
+> To view which are available check the `libs.versions.toml` file under the `gradle` folder.
```
vetleledaal, line 284:
```suggestion
> To view which are available check the `gradle/libs.versions.toml` file.
```

CONTRIBUTING.md
```diff
@@ -431,6 +433,14 @@ To test if the URL intent filter is working as expected, you can try opening the
 $ adb shell am start -d "<your-link>" -a android.intent.action.VIEW
 ```
 
+> [!CAUTION]
+> The activity has no support to any Kotlin Intrinsics specific methods or calls,
+> and using it will causes crashes in the activity. Consider using Java's equivalent
+> methods instead, such as using `String`'s `equals()` instead of using `==`.
```
vetleledaal, line 439:
```suggestion
> The activity does not support any Kotlin Intrinsics specific methods or calls,
> and using them will causes crashes in the activity. Consider using Java's equivalent
> methods instead, such as `String`'s `equals()` instead of `==`.
```

CONTRIBUTING.md
```diff
@@ -452,10 +462,18 @@ override val id: Long = <the-id>
...
+> The package name **needs** to be the same (even if it has the old name), otherwise users will not receive the extension update when it gets published in the repository. If you're changing the name of a multisrc source, you can manually set it in the generator class of the theme by using `pkgName = "oldpackagename"`.
 
 The `id` also needs to be explicity set to the old value if you're changing the `lang` attribute.
 
+> [!NOTE]
+> If the source has also changed their theme and hence making the old saved URLs
+> in the library incompatible, you can skip this part and just rename by changing
+> the `name` in the source class and in the Gradle file, but keeping the old package.
+> By doing so, a new `id` will be generated so users will be forced to migrate to
+> update their entries' URLs.
```
vetleledaal, line 475:
```suggestion
> If the source has also changed their theme you can instead just change
> the `name` field in the source class and in the Gradle file. By doing so
> a new `id` will be generated and users will be forced to migrate.
```
*might be throwing away too much information, but the details aren't really that important here, IMHO

CONTRIBUTING.md
```diff
@@ -603,7 +620,8 @@ And for a release build of Tachiyomi:
 -W -S -n eu.kanade.tachiyomi/eu.kanade.tachiyomi.ui.main.MainActivity -a eu.kanade.tachiyomi.SHOW_CATALOGUES
 ```
 
-If you're deploying to Android 11 or higher, enable the "Always install with package manager" option in the run configurations.
+> [!IMPORTANT]
+> If you're deploying to Android 11 or higher, enable the "Always install with package manager" option in the run configurations, otherwise you may face issues such as building the extension without your new modifications.
```
vetleledaal, line 624:
```suggestion
> If you're deploying to Android 11 or higher, enable the "Always install with package manager" option in the run configurations. Without this option enabled, changes will not apply until you restart Tachiyomi.
```

CONTRIBUTING.md
```diff
@@ -624,11 +642,17 @@ You can also elect to simply rely on logs printed from your extension, which
 show up in the [`Logcat`](https://developer.android.com/studio/debug/am-logcat) panel of Android Studio.
 
 ### Inspecting network calls
+
 One of the easiest way to inspect network issues (such as HTTP errors 404, 429, no chapter found etc.) is to use the [`Logcat`](https://developer.android.com/studio/debug/am-logcat) panel of Android Studio and filtering by the `OkHttpClient` tag.
 
 To be able to check the calls done by OkHttp, you need to enable verbose logging in the app, that is not enabled by default and is only included in the Preview versions of Tachiyomi. To enable it, go to More -> Settings -> Advanced -> Verbose logging. After enabling it, don't forget to restart the app.
 
-Inspecting the Logcat allows you to get a good look at the call flow and it's more than enough in most cases where issues occurs. However, alternatively, you can also use an external tool like `mitm-proxy`. For that, refer to the next section.
+Inspecting the Logcat allows you to get a good look at the call flow and it's more than enough in most cases where issues occurs. However, alternatively, you can also use an external tool like `mitm-proxy`. For that, refer to the next sections.
```
vetleledaal, line 650:
```suggestion
Inspecting the Logcat allows you to get a good look at the call flow and it's more than enough in most cases where issues occurs. However, alternatively, you can also use an external tool like `mitm-proxy`. For that, refer to the subsequent sections.
```

CONTRIBUTING.md
```diff
@@ -624,11 +642,17 @@ You can also elect to simply rely on logs printed from your extension, which
...
+
+On newer Android Studio versions, you can use its built-in Network Inspector inside the
+App Inspection tool window, which provides a nice GUI to inspect the requests done in the app.
```
vetleledaal, line 653:
```suggestion
On newer Android Studio versions, you can use its built-in Network Inspector found within the
App Inspection tool window. This feature provides a nice GUI to inspect the requests made in the app.
```

CONTRIBUTING.md
```diff
@@ -624,11 +642,17 @@ You can also elect to simply rely on logs printed from your extension, which
...
+
+To use it, follow the [official documentation](https://developer.android.com/studio/debug/network-profiler) and select Tachiyomi package name on the process list.
```
vetleledaal, line 655:
```suggestion
To use it, follow the [official documentation](https://developer.android.com/studio/debug/network-profiler) and select Tachiyomi package name in the process list.
```

vetleledaal (commented):
Not 100% on everything, so thread with caution 🐍

CONTRIBUTING.md
alessandrojean: That's not really accurate. If the option isn't marked, what could happen is Android Studio compiling it but when you run it doesn't have the changes in the extension which you might have done.

vetleledaal: Not sure how you would word it then.

As an anecdote, I have observed the described issue, and solved it by restarting Tachiyomi. I did not have to recompile it for it to start working, kinda like the app was still holding on to a reference to the old version.

